PR #138, opened 2016-12-16
The patch about stagePreRun(maybe a simple begin of stream shuffle)

xuanyuanking:

Hi chenghao & daoyuan, as today I mentioned, there's a simple realization in Baidu of limit optimization by preRun the limit stage. Also some refetching logic, please have a look when you have time. Thanks :)
We mainly do these things below:
    1. Add new func `preRunGetMapSizesByExecutorId` in `MapOutputTracker` for the scenario of preRun=open, atomize `getStatus` and `outputMissing` operations
    2. Set `isOutputMissing` private
    3. Add new Class PreRunShuffleBlockFetcherIterator
    4. Add config `spark.sql.baidu.stagePreRun`
          1) reader check the stagePreRun config to make a decision of using which shuffleBlockIterator
          2) mapOutputTracker support the preRun config and default value
    5. Add the comments of new added config

## How was this patch tested?

It currently running in baidu internal and it can accelerate some limit scenario.
And this patch has a bug here when the upstream is down, it will still refetch, I will fix it by adding a max retry time.

### DIFF
--- a/core/src/main/scala/org/apache/spark/MapOutputTracker.scala
+++ b/core/src/main/scala/org/apache/spark/MapOutputTracker.scala
@@ -184,7 +184,9 @@ private[spark] abstract class MapOutputTracker(conf: SparkConf) extends Logging 
     if (statuses == null) {
       return true
     }
-    size != statuses.length
+    val length = statuses.length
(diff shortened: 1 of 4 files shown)


adrian-wang:
I'll close this since OAP is now public.

Issue #227, opened 2017-05-25
ArrayIndexOutOfBoundsException in IndexedScan


adrian-wang:
```
create table hive_test (key int, value string);
LOAD DATA LOCAL INPATH '/sql/hive/src/test/resources/data/files/small_kv.txt' OVERWRITE INTO TABLE hive_test;
create table hive_test_spn (key int, value string) using spn;
insert into hive_test_spn select * from hive_test;
create sindex hive_test_spn_idx on hive_test_spn (key, value);
```
I see
```
java.lang.ArrayIndexOutOfBoundsException: 0
	at org.apache.spark.sql.catalyst.expressions.GenericInternalRow.genericGet(rows.scala:227)
	at org.apache.spark.sql.catalyst.expressions.BaseGenericInternalRow$class.getAs(rows.scala:35)
	at org.apache.spark.sql.catalyst.expressions.BaseGenericInternalRow$class.get(rows.scala:37)
	at org.apache.spark.sql.catalyst.expressions.GenericInternalRow.get(rows.scala:221)
	at org.apache.spark.sql.execution.datasources.spinach.io.DataFile$$anonfun$encodeInterval$2.apply(DataFile.scala:156)
	at org.apache.spark.sql.execution.datasources.spinach.io.DataFile$$anonfun$encodeInterval$2.apply(DataFile.scala:155)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
	at scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
	at org.apache.spark.sql.execution.datasources.spinach.io.DataFile$.encodeInterval(DataFile.scala:155)
	at org.apache.spark.sql.execution.datasources.spinach.index.IndexScanner.encodeIntervalAndSchema(IndexScanner.scala:87)
	at org.apache.spark.sql.execution.datasources.spinach.index.BPlusTreeScanner.initialize(BPlusTreeScanner.scala:42)
[...]

PR #241, opened 2017-06-07
Revert "Modify Index To Use Dictionary Encoding (#209)"

linhong-intel:

This reverts commit a141d3a3c8363fcca1d929b50556e1004d6bd577.

Why I revert this commit:
1. To utilize dictionary encoding in Index, current Index implementation need
   to be modified in both writer and scanner. But there is no plan yet. So, this
   commit can't bring any benefit while adding some extra latency.

2. In current code structure, IndexScanner has to read Data File first to decide
   if dictionary encoding is needed. This act against the purpose of index.

3. Data Writer used a 'fallback' strategy to decide if a column can use dictionary
   encoding. This will cause one column has two or more encoding types. It's hard
   for Index to handle this senario.

So,  revert this commit until it can bring some benefit.

## How was this patch tested?

Unit test

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,11 +36,6 @@ spinach.rowgroup.szie
 * Default: 1024
 * Usage: `System.setProperty("spinach.rowgroup.size", "1024")`
 
-spinach.encoding.dictionaryEnabled
-* To enable/disable dictionary encoding
-* Default: False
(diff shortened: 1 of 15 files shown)


linhong-intel:
Closes #227

linhong-intel:
Test on demo data set and revert this commit shows improvement in index scan

PR #248, opened 2017-06-13
fix parquet index building

### LINKED ISSUE #243: No Lease Exception on Parquet File Format in Index Building
Parquet data file is splittable while reading. So multiple tasks can access one index file.

adrian-wang:

Not allow parquet to split one file into multiple tasks to eliminate no lease error when building index, if we are using OAP for parquet data.

## How was this patch tested?

add a test case.
Fixes #243


adrian-wang:
I'm merging this.

xuanyuanking:
We will support Parquet file splitable later? Is this patch just the work-around in currently?

adrian-wang:
For parquet, if we let them split files, multiple tasks may read the same index file, which lead to more IO operations. For now we just disable splitable, later we may figure out if we really need split file, or maybe only disable splitable when index is useful.

Issue #249, opened 2017-06-13
SampleBasedStatisticsSuite is flaky


adrian-wang:
sometimes, we will get
```
- test analyze function *** FAILED ***
  0.03333333333333333 did not equal 0 (SampleBasedStatisticsSuite.scala:144)
```
when running SampleBasedStatisticsSuite

yueawang:
Just found this, I also made a PR for the same issue.

linhong-intel:
Still be seen in test.

PR #250, opened 2017-06-13
Fix SampleBasedStatistics bug.

gfl94:

Fix `SampleBaseStatisticsSuite` bug #249 .

This happens with low possibility, for the sample based property.

`StatisticsManager` does not has this problem for its `analyze` function.

## How was this patch tested?

unit test.


adrian-wang:
Thanks!

PR #254, opened 2017-06-14
Fix a random test error.

yueawang:
It happens when random number overlaps with the query interval. At
that time, analysis function report coverage instead of INDEX type.

To fix a random test error.

## How was this patch tested?

unit test.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/spinach/statistics/SampleBasedStatistics.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/spinach/statistics/SampleBasedStatistics.scala
@@ -86,10 +86,13 @@ private[spinach] class SampleBasedStatistics extends Statistics {
         if (Statistics.rowInIntervalArray(row, intervalArray, ordering)) hitCnt += 1
       }
 
-      if (hitCnt < sampleArray.length)
+      if (hitCnt < sampleArray.length) {
         StaticsAnalysisResult.USE_INDEX
+      }
       else
+      {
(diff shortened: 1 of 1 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/spinach/statistics/SampleBasedStatistics.scala
```diff
@@ -85,7 +85,14 @@ private[spinach] class SampleBasedStatistics extends Statistics {
...
       }
-      hitCnt * 1.0 / sampleArray.length
+
+      if (hitCnt < sampleArray.length) {
+        StaticsAnalysisResult.USE_INDEX
+      }
+      else
+      {
+        StaticsAnalysisResult.FULL_SCAN
+      }
```
gfl94: I think for these `Statistics` with value result, we can return a double type result, instead of the `StaticsAnalysisResult`, we can design more algorithm on these `statistics` in `StatisticsManager.analyze`.

yueawang: Yes, I agree, here I actually don't know why it returns a flag while it claims to return a double value. maybe a (FLAG, DOUBLE) tuple looks better if necessary.

linhong-intel:
This PR acts against Geifei's statistic design. I don't think it's a good idea. 

> At that time, analysis function report coverage instead of INDEX type.

This is expected.

I think the #249 is caused by wrong test case design, right?

adrian-wang:
I'm closing this.

PR #262, opened 2017-06-15
Fix ArrayIndexOutOfBoundsException in #252

### LINKED ISSUE #253: a lot of "ArrayIndexOutOfBoundsException: 0" found by PR #252
All of them should be caused by calling `IndexScanner.DUMMY_KEY_START.get(i)`

gfl94:

ArrayIndexOutOfBoundsException while get DUMMY_START & DUMMY_END 

closes #252 
closes #253 

## How was this patch tested?

integration tests

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/spinach/statistics/PartByValueStatistics.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/spinach/statistics/PartByValueStatistics.scala
@@ -114,14 +114,10 @@ private[spinach] class PartByValueStatistics extends Statistics {
   // value array:(-inf, r0) [r0,r1) [r1,r2) [r2,r3) [r3,r4) [r4,r5]  (r5, +inf)
   protected def getIntervalIdx(row: Key, include: Boolean): Int = {
     var i = 0
-    if (row == IndexScanner.DUMMY_KEY_START) 0 // -inf
-    else if (row == IndexScanner.DUMMY_KEY_END) metas.length + 1 // +inf
-    else {
-      while (i < metas.length && (include && ordering.gteq(row, metas(i).row)
-        || !include && ordering.gt(row, metas(i).row))) i += 1
-      if (include && ordering.compare(metas.last.row, row) == 0) metas.last.idx // for row == r5
-      else i
-    }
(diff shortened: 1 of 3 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/spinach/statistics/PartByValueStatistics.scala
```diff
@@ -114,8 +114,8 @@ private[spinach] class PartByValueStatistics extends Statistics {
   // value array:(-inf, r0) [r0,r1) [r1,r2) [r2,r3) [r3,r4) [r4,r5]  (r5, +inf)
   protected def getIntervalIdx(row: Key, include: Boolean): Int = {
     var i = 0
-    if (row eq IndexScanner.DUMMY_KEY_START) 0 // -inf
-    else if (row eq IndexScanner.DUMMY_KEY_END) metas.length + 1 // +inf
+    if (row == IndexScanner.DUMMY_KEY_START) 0 // -inf
+    else if (row == IndexScanner.DUMMY_KEY_END) metas.length + 1 // +inf
```
gfl94: Can we distinguish whether a DUMMY row is a `DUMMY_KEY_START` or a `DUMMY_KEY_END` here?

adrian-wang: I think they are the same...

linhong-intel:
Thanks Gefei.
But seems your code is not considering multiple index key. like below:
```
test("multi index") {
    val data: Seq[(Int, String)] = (1 to 300).map { i => (i, s"this is test $i") }
    data.toDF("key", "value").createOrReplaceTempView("t")
    sql("insert overwrite table spinach_test select * from t")
    sql("create sindex index1 on spinach_test (a, b)")

    checkAnswer(sql("SELECT * FROM spinach_test WHERE a == 1 and b < 'this is test 2'"),
      Row(1, "this is test 1") :: Nil)

    sql("drop sindex index1 on spinach_test")
  }
```

gfl94:
I think this can be regards as a new feature for `Statistics`, I haven't
taken this into consideration while designing the `analyze` function. So
sorry for that:(

linhong-intel:
Merge it anyway. Will fix multiple index problem later.

PR #263, opened 2017-06-16
Use InternalRow instead of UnsafeRow to save Memory

linhong-intel:

For index key, the field size usually is small, so InternalRow is better.

## How was this patch tested?

Unit test


linhong-intel:
will merge it by today.

adrian-wang:
LGTM

PR #265, opened 2017-06-20
Read spinach config from Table Property

### LINKED ISSUE #235: System.setProperty(key, value) not working in cluster mode
in cluster mode, driver and workers are not in a same process. so `System.setProperty` is not working.

linhong-intel:

Read spinach config from Table Property
fix some code style problem

TODO: Should we have our own config util instead of SqlConf?
TODO: How to confirm the config takes effect?

Fixes #235 

## How was this patch tested?

Unit test

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,16 +31,29 @@ mvn test
 * Fine-grained cache
 * Parquet Data Adaptor
 ## Configuration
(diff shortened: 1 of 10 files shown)


linhong-intel:
I'll merge it by today

PR #267, opened 2017-06-20
add create time to index name

adrian-wang:

Due to current index name is only decided by data file name and index name, drop an index and create another index using the same name will cause some bizarre because of caching key is file name. In this patch, we are using the hex string of system time when creating to distinguish different indices.

To be specific, index `idx1` created at time `15cc47fb3d8` will use the index name of `....15cc47fb3d8.idx1.index`.

## How was this patch tested?

Existed tests with some modification.


linhong-intel:
LGTM

PR #269, opened 2017-06-20
[BUILD] use java 7 and turn off verbose mode of scalastylecheck

adrian-wang:

use java 7 and turn off verbose mode of scalastylecheck.

## How was this patch tested?

Current tests.


adrian-wang:
Setting the version to java 7 allow us to build the project with java 7, but we can still run it on java 8. The old setting would not allow user to build it on java 7.

PR #270, opened 2017-06-21
Save MinMax into Data File Handle

linhong-intel:

Main Changes:
- Create ColumnStatistics to store Min Max
- Gather encoding, dictionary, statistic data into ColumnMeta
- Change read/write sequence due to some data is not fixed length
- Store MinMax during writing rows

## How was this patch tested?

Unit Test


linhong-intel:
I'll merge it by today

PR #280, opened 2017-06-28
Shade all parquet dependencies to avoid lib conflict in Spark.

yueawang:

Shade all parquet dependencies to avoid lib conflict in Spark.

## How was this patch tested?
mvn test pass


adrian-wang:
LGTM

Issue #279, opened 2017-06-28
tryToReadStatistics should be put before IndexScanner initialize


linhong-intel:
```
filterScanner match {
      case Some(fs) if fs.existRelatedIndexFile(path, conf) =>
        fs.initialize(path, conf)
        val indexPath = IndexUtils.indexFileFromDataFile(path, fs.meta.name, fs.meta.time)
        val statsAnalyseResult = tryToReadStatistics(indexPath, conf)
```
`fs.initialize()` will load index file from disk. We should determine whether to skip index before this line.

Issue #277, opened 2017-06-28
Index Statistics is not configurable


linhong-intel:
After refactor, seems some Index Statistics parameters are hard-coded or not used.
For example:
SPINACH_FULL_SCAN_THRESHOLD = hard coded to 0.8
SPINACH_STATISTICS_TYPES = not used
SPINACH_STATISTICS_SAMPLE_RATE = hard coded to 0.1
SPINACH_BLOOMFILTER_MAXBITS = hard coded to 1 << 20
SPINACH_BLOOMFILTER_NUMHASHFUNC = hard coded to 3

@gfl94 Can you confirm with this?

gfl94:
Statistics parameters can be configured through `StatisticsManager` object method, please refer to [code here](https://github.com/Intel-bigdata/OAP/blob/master/src/main/scala/org/apache/spark/sql/execution/datasources/spinach/statistics/StatisticsManager.scala#L180).

linhong-intel:
>Statistics parameters can be configured through `StatisticsManager` object method, please refer to [code here](https://github.com/Intel-bigdata/OAP/blob/master/src/main/scala/org/apache/spark/sql/execution/datasources/spinach/statistics/StatisticsManager.scala#L180).

Thanks, but I thinks it's better to config it through Configuration. I made the PR #281 enabled this, you can review it.

PR #281, opened 2017-06-29
Make Statistics Parameter configurable

linhong-intel:

1. Use sparkSession.sessionState.newHaddopConf in CreateIndex. Because
   sparkSession.sparkContext.hadoopConfiguration doesn't have full configs
2. Setup Statistics parameters during initialization
3. Put indexScanner.initialization after tryToReadStatistics

## How was this patch tested?

Unit Tests

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexWriter.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexWriter.scala
@@ -120,7 +120,7 @@ private[oap] class BTreeIndexWriter(
 
     def writeTask(): Seq[IndexBuildResult] = {
       val statisticsManager = new StatisticsManager
-      statisticsManager.initialize(BTreeIndexType, keySchema)
+      statisticsManager.initialize(BTreeIndexType, keySchema, configuration)
       // key -> RowIDs list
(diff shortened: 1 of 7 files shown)


linhong-intel:
Fixes #277
Fixes #279

src/main/scala/org/apache/spark/sql/execution/datasources/spinach/statistics/StatisticsManager.scala
```diff
@@ -170,17 +202,19 @@ object StatisticsManager {
...
         BloomFilterStatisticsType, PartByValueStatisticsType))
 
-  var sampleRate: Double = 0.1
-  var partNumber: Int = 5
-  var bloomFilterMaxBits: Int = 1 << 20
-  var bloomFilterHashFuncs: Int = 3
+  var sampleRate: Double = _
+  var partNumber: Int = _
+  var bloomFilterMaxBits: Int = _
+  var bloomFilterHashFuncs: Int = _
```
gfl94: Will these parameters be required while loading statistics from disk?

linhong-intel: OK, I see. Seems parameters are not initialized during read statistics. I will check it.

PR #283, opened 2017-06-29
remove string related to spinach

adrian-wang:

rename spinach to OAP

## How was this patch tested?

existed tests


linhong-intel:
Double checked there is no "spinach" in project. Except one line:

> OAP - Optimized Analytics Package (Spinach as code name)

adrian-wang:
I think it is fine to keep that.

linhong-intel:
Suggest we merge this ASAP to avoid conflict.

Issue #276, opened 2017-06-27
Index Statistics does not support multi-column index


linhong-intel:
PR #275 Skipped Statistics for multi-column index to avoid bugs.
But we still need to fix bugs and enable it.

linhong-intel:
Fixed in #284

Issue #264, opened 2017-06-20
"ArrayIndexOutOfBoundsException: 0" During multiple column index build.


linhong-intel:
```
test("multi index") {
    val data: Seq[(Int, String)] = (1 to 300).map { i => (i, s"this is test $i") }
    data.toDF("key", "value").createOrReplaceTempView("t")
    sql("insert overwrite table spinach_test select * from t")
    sql("create sindex index1 on spinach_test (a, b)")

    checkAnswer(sql("SELECT * FROM spinach_test WHERE a == 1 and b < 'this is test 2'"),
      Row(1, "this is test 1") :: Nil)

    sql("drop sindex index1 on spinach_test")
  }
```

linhong-intel:
Fixed in #284

PR #284, opened 2017-06-30
Fix bugs in Index Statistics for Multi-Column case

linhong-intel:

In Multi-Column case, an interval has more than one columns and the last one
maybe DUMMY_KEY. For example interval.start = (1, 100, "a", DUMMY_KEY_START)
But, interval.start != DUMMY_KEY_START, and compare it may still encounter
ArrayIndexOutOfBoundException. This PR fixes this problem

## How was this patch tested?

Unit tests and added case


linhong-intel:
Fixes #276 and #264

PR #286, opened 2017-07-03
Fix PartByValueStatistics wrong index issue.

yueawang:
Sort keys before we count statistics, otherwise
index is incorrect. Meanwhile, All statistics are
disabled for no practical application so far.

## How was this patch tested?
mvn test pass.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/PartByValueStatistics.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/PartByValueStatistics.scala
@@ -50,7 +50,9 @@ private[oap] class PartByValueStatistics extends Statistics {
                                          curMaxId: Int, accumulatorCnt: Int)
   protected lazy val metas: ArrayBuffer[PartedByValueMeta] = new ArrayBuffer[PartedByValueMeta]()
 
-  override def write(writer: IndexOutputWriter, sortedKeys: ArrayBuffer[Key]): Long = {
+  override def write(writer: IndexOutputWriter, keys: ArrayBuffer[Key]): Long = {
+    val sortedKeys = keys.sortWith(ordering.compare(_, _) < 0) // sort before do statistics.
+
     var offset = super.write(writer, sortedKeys)
     val hashMap = new java.util.HashMap[Key, Int]()
     val uniqueKeys: ArrayBuffer[Key] = new ArrayBuffer[Key]()
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/PartByValueStatistics.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/PartByValueStatistics.scala
@@ -50,9 +50,7 @@ private[oap] class PartByValueStatistics extends Statistics {
                                          curMaxId: Int, accumulatorCnt: Int)
   protected lazy val metas: ArrayBuffer[PartedByValueMeta] = new ArrayBuffer[PartedByValueMeta]()
 
-  override def write(writer: IndexOutputWriter, keys: ArrayBuffer[Key]): Long = {
-    val sortedKeys = keys.sortWith(ordering.compare(_, _) < 0) // sort before do statistics.
-
+  override def write(writer: IndexOutputWriter, sortedKeys: ArrayBuffer[Key]): Long = {
     var offset = super.write(writer, sortedKeys)
     val hashMap = new java.util.HashMap[Key, Int]()
     val uniqueKeys: ArrayBuffer[Key] = new ArrayBuffer[Key]()
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
@@ -167,7 +167,6 @@ object StatisticsManager {
   val STATISTICSMASK: Long = 0x20170524abcdefabL // a random mask for statistics begin
 
   val statisticsTypeMap: scala.collection.mutable.Map[AnyIndexType, Array[StatisticsType]] =
-    // temporarily disable all statistics.
     scala.collection.mutable.Map(
       BTreeIndexType -> Array(MinMaxStatisticsType, SampleBasedStatisticsType,
         BloomFilterStatisticsType, PartByValueStatisticsType),
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
@@ -101,10 +101,10 @@ class StatisticsManager {
   private def sortKeys(): ArrayBuffer[Key] = {
     if (!sortFlag) {
       sortFlag = true
-      content.sortWith((l, r) => ordering.compare(l, r) < 0)
-    } else {
-      content
+      content = content.sortWith((l, r) => ordering.compare(l, r) < 0)
     }
+
+    content
   }
 
   def read(bytes: Array[Byte], s: StructType): Unit = {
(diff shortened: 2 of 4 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
```diff
@@ -164,11 +164,10 @@ object StatisticsManager {
   val STATISTICSMASK: Long = 0x20170524abcdefabL // a random mask for statistics begin
 
   val statisticsTypeMap: scala.collection.mutable.Map[AnyIndexType, Array[StatisticsType]] =
+    // temporarily disable all statistics.
```
linhong-intel: Why do this?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/PartByValueStatistics.scala
```diff
@@ -50,7 +50,9 @@ private[oap] class PartByValueStatistics extends Statistics {
                                          curMaxId: Int, accumulatorCnt: Int)
   protected lazy val metas: ArrayBuffer[PartedByValueMeta] = new ArrayBuffer[PartedByValueMeta]()
 
-  override def write(writer: IndexOutputWriter, sortedKeys: ArrayBuffer[Key]): Long = {
+  override def write(writer: IndexOutputWriter, keys: ArrayBuffer[Key]): Long = {
+    val sortedKeys = keys.sortWith(ordering.compare(_, _) < 0) // sort before do statistics.
```
linhong-intel: Add a test case for this fix. I mean a test case will fail with this fix.

gfl94: For integration test, keys are sorted inside [`StatisticsManager`](https://github.com/Intel-bigdata/OAP/blob/master/src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala#L81), I am just wondering whether an additional sort needed here.

linhong-intel: In `StatisticsManager` I saw this, `if (!sortFlag) sortKeys()` and `sortFlag` is never set to true. We can't assume User will set it.
But I agree to move sortKeys() to `StaticsManager`

gfl94: Thanks,`sortFlag` is a private variable, user will not have access to this, I forget to set this flag after sort function.

adrian-wang:
```
class IndexedQuerySuite  extends QueryTest with SharedSQLContext with BeforeAndAfterEach {
  import testImplicits._

  private var path: File = _
  override def beforeAll(): Unit = {
    super.beforeAll()
    path = Utils.createTempDir()
    path.delete()

    val data: Seq[(Int, String)] =
      (2, "this is test 1")::
        (6, "this is test 2")::
        (5, "this is test 3")::
        (7, "this is test 4")::
        (8, "this is test 5")::
        (3, "this is test 6")::
        (8, "this is test 7")::
        (9, "this is test 8")::
        (6, "this is test 9")::
        (9, "this is test 10")::
        (1, "this is test 11")::
        (2, "this is test 12")::
        (0, "this is test 13")::
        (8, "this is test 14")::
        (1, "this is test 15")::
        (1, "this is test 16")::
        (3, "this is test 17")::
        (9, "this is test 18")::
        (1, "this is test 19")::
        (7, "this is test 20"):: Nil

    data.toDF("a", "b").write.format("spn").mode(SaveMode.Overwrite).save(path.getAbsolutePath)
  }

  test("index test 1") {
    val df = sqlContext.read.format("spn").load(path.getAbsolutePath)

    df.createOrReplaceTempView("spinach_test_1")
    sql("create sindex index1 on spinach_test_1 (a) using bitmap") // Neither btree nor bitmap works

    val dfwithIdx = sql("SELECT * FROM spinach_test_1 WHERE a = 8")
    dfwithIdx.show
    val count1 = dfwithIdx.count
    sql("drop sindex index1 on spinach_test_1")
    val count2 = sql("SELECT * FROM spinach_test_1 WHERE a = 8").count

    if (count1 != count2) {
      sql("SELECT * FROM spinach_test_1").show
    }
    assert(count1 == count2)
  }

}
```

yueawang: @linhong-intel I saw you have already added a test case for that.  I will fix the sortFlag in this PR.
And I remember that you mentioned so far statistics has no usecase, so we can disable it, do I misunderstand you??

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
```diff
@@ -78,7 +78,6 @@ class StatisticsManager {
   }
 
   def write(out: IndexOutputWriter): Long = {
-    if (!sortFlag) sortKeys()
```
linhong-intel: Why we can't just put a `content.sortWith((l, r) => ordering.compare(l, r) < 0)` here and remove `sortFlag` and `sortKeys`. I can't find a real difference between the two implementations.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapDDLSuite.scala
```diff
@@ -123,5 +123,20 @@ class OapDDLSuite extends QueryTest with SharedSQLContext with BeforeAndAfterEac
       new Configuration()).globStatus(new Path(path,
       "oap_partition_table/b=2/c=c2/*.index")).length != 0)
   }
+
+  test("index integrity") {
+      val rand = new java.util.Random(System.currentTimeMillis())
+      val data: Seq[(Int, String)] = (1 to 300).map {
+        i => (rand.nextInt(300), s"this is test $i") }
```
linhong-intel: Seems you can use `scala.util.Random.shuffle(1 to 300)` here. I prefer a scala class in scala code.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapDDLSuite.scala
```diff
@@ -123,5 +123,20 @@ class OapDDLSuite extends QueryTest with SharedSQLContext with BeforeAndAfterEac
...
+
+      data.toDF("key", "value").createOrReplaceTempView("t")
+      sql("insert overwrite table oap_test_1 select * from t")
+      sql("create sindex index1 on oap_test_1 (a) using bitmap")
+
+      val dfwithIdx = sql("SELECT * FROM oap_test_1 WHERE a > 8 and a <= 200")
+      sql("drop sindex index1 on oap_test_1")
+      val dfWithoutIdx = sql("SELECT * FROM oap_test_1 WHERE a > 8 and a <= 200")
```
linhong-intel: Use `SELECT * FROM t WHERE a > 8 and a <= 200` as a comparison in case we have a bug even if drop the index.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
yueawang: content.sortWith() does sort but returns a sorted array and leave content itself unchange, I am not sure if there is a mutable sort can be applied to change itself, if so we can use that in future.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapDDLSuite.scala
yueawang: OK, I will try to use scala native class.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapDDLSuite.scala
adrian-wang: I think we'd better have a separate suite for all kinds of  bug fix validations. How about `IndexedQuerySuite`?

linhong-intel:
LGTM

adrian-wang:
LGTM

adrian-wang:
pending test

gfl94:
LGTM

PR #289, opened 2017-07-05
[WORKAROUND] Skip using index if index file is too large

linhong-intel:

Skip using index if index file is too large and fix some style problem

## How was this patch tested?

Unit test

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -85,7 +85,7 @@ private[oap] trait AbstractFiberCacheManger extends Logging {
         // also be *FAR* less than Int.MaxValue. If totalWeight > Int.MaxValue, overflow, then
         // totalWeight will treated as a very small value, and never greater than maximumWeight.
         logWarning(s"${SQLConf.OAP_FIBERCACHE_SIZE.key}): $weightConfig is too large." +
-          s"Reducing to 8TB")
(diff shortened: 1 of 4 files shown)


src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapIndexQuerySuite.scala
```diff
@@ -25,7 +25,7 @@ import org.apache.spark.sql.test.SharedSQLContext
 import org.apache.spark.util.Utils
 
 
-class OapDDLIndexQuerySuite extends QueryTest with SharedSQLContext with BeforeAndAfterEach {
+class OapIndexQuerySuite extends QueryTest with SharedSQLContext with BeforeAndAfterEach {
```
adrian-wang: good catch!

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -75,12 +85,14 @@ private[oap] trait AbstractFiberCacheManger extends Logging {
         // also be *FAR* less than Int.MaxValue. If totalWeight > Int.MaxValue, overflow, then
         // totalWeight will treated as a very small value, and never greater than maximumWeight.
         logWarning(s"${SQLConf.OAP_FIBERCACHE_SIZE.key}): $weightConfig is too large." +
-          s"Reducing to 4TB")
+          s"Reducing to 8TB")
```
adrian-wang: Reducing to 8TB or less

PR #290, opened 2017-07-05
add version info in index file

adrian-wang:

Add index version info in index file, to support backward compatibility while further optimization

## How was this patch tested?

Exists tests.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexWriter.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexWriter.scala
@@ -164,6 +164,7 @@ private[oap] class BTreeIndexWriter(
       var i = 0
       var fileOffset = 0L
       val offsetMap = new java.util.HashMap[InternalRow, Long]()
+      fileOffset += writeHead(writer, 1)
       // write data segment.
       while (i < partitionUniqueSize) {
         offsetMap.put(uniqueKeys(i), fileOffset)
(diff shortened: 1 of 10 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexWriter.scala
```diff
@@ -164,6 +164,7 @@ private[oap] class BTreeIndexWriter(
       var i = 0
       var fileOffset = 0L
       val offsetMap = new java.util.HashMap[InternalRow, Long]()
+      fileOffset += writeHead(writer, 1)
```
yueawang: why don't use 0.2.0 here? 1 looks like a magic number.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/IndexFile.scala
```diff
@@ -57,4 +58,18 @@ private[oap] case class IndexFile(file: Path) {
...
   }
+
+  def version(conf: Configuration): Int = {
+    val fs = file.getFileSystem(conf)
+    val fin = fs.open(file)
+    val bytes = new Array[Byte](8)
+    fin.read(bytes, 0, 8)
+    fin.close()
+    bytes(6) << 8 + bytes(7)
+    1
```
linhong-intel: what is this? line 69

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexWriter.scala
adrian-wang: it just starts from 1, since we are not going to change index schema every release.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/IndexFile.scala
adrian-wang: oh, it is a line for debug... I'll delete it.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexWriter.scala
adrian-wang: in later versions, we will have a map from version to different reader and writer(maybe only keep one writer in one release, but multiple reader for compatibility).

adrian-wang:
I'll merge this today.

PR #292, opened 2017-07-10
[WORKAROUND] Using full scan if index file is larger than data file.

yueawang:
## How was this patch tested?
Unit test


linhong-intel:
lgtm

linhong-intel:
Checked the unit test log. With this PR, index is never used in test suite. We need address this in next PR.

adrian-wang:
In my opinion, we can add a invisible configuration value to enable or disable this, and we only disable this for tests.

PR #296, opened 2017-07-11
optimize oap config

linhong-intel:

Set default row group size to 1048576
Set default compression codec to GZIP
Set default fiber cache size to 0.3 * EXECUTOR_MEMORY

## How was this patch tested?

Unit test


adrian-wang:
Can you also update the default values in readme,md?

PR #297, opened 2017-07-12
use a internal config to indicate test is ongoing

linhong-intel:

use a internal config to indicate test is ongoing

## How was this patch tested?

Unit test


adrian-wang:
LGTM

PR #298, opened 2017-07-13
Fix nit of index check logic in DataSourceMeta

xuanyuanking:

Find a nit in `isSupportedByIndex` logic, the AttributeReference may appear in both side of BinaryComparison. Example in `TypeCoercion.scala`

## How was this patch tested?
Add tests in origin DataSourceMetaSuite.


adrian-wang:
LGTM

PR #304, opened 2017-08-01
Avoid NPE if bestIndexer is null when debug is on.

lee-lei:

add if statement to check that the bestIndexer is not null.
Use logInfo instead of logDebug to avoid scalastyle exception.

## How was this patch tested?
Use mvn test and all the tests passed.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
@@ -137,7 +137,9 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
       lastIdx = availableIndexes.head._1
       bestIndexer = availableIndexes.head._2
     }
-    logDebug("\t" + bestIndexer.toString + "; lastIdx: " + lastIdx)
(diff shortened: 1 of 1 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -137,7 +137,9 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
       lastIdx = availableIndexes.head._1
       bestIndexer = availableIndexes.head._2
     }
-    logDebug("\t" + bestIndexer.toString + "; lastIdx: " + lastIdx)
+    if (bestIndexer != null) {
+      logInfo("\t" + bestIndexer.toString + "; lastIdx: " + lastIdx)
```
linhong-intel: What's the scalastyle problem here if use debug level? Also add an else statement to indicate there is no index available.

adrian-wang: +1 keep it as logDebug

lee-lei:
keep logDebug and added else statement according to the comments from the reivewers. Thanks.

PR #305, opened 2017-08-02
Enable orderby limit feature

yueawang:
Enable order-by/limit feature on a specific column on OAP if related index is presented.

## How was this patch tested?
mvn test passed.


src/main/scala/org/apache/spark/sql/execution/OAPStrategies.scala
```diff
@@ -0,0 +1,283 @@
...
+        logInfo(s"Pushed Filters: ${pushedDownFilters.mkString(",")}")
+
+        val readFile = files.fileFormat.asInstanceOf[OapFileFormat].buildReaderWithPartitionValues(
+          sparkSession = files.sparkSession,
+          dataSchema = files.dataSchema,
+          partitionSchema = files.partitionSchema,
+          requiredSchema = prunedDataSchema,
+          filters = pushedDownFilters,
+          order.isAscending,
+          limit,
```
adrian-wang: Is it possible that `limit` is negative here?

src/main/scala/org/apache/spark/sql/execution/OAPStrategies.scala
```diff
@@ -0,0 +1,283 @@
...
+    private def getBlockLocations(file: FileStatus): Array[BlockLocation] = file match {
+      case f: LocatedFileStatus => f.getBlockLocations
+      case f => Array.empty[BlockLocation]
+    }
+
+    // Given locations of all blocks of a single file, `blockLocations`, and an `(offset, length)`
+    // pair that represents a segment of the same file, find out the block that contains the largest
+    // fraction the segment, and returns location hosts of that block.
+    // If no such block can be found, returns an empty array.
+    private def getBlockHosts(
```
adrian-wang: put this in some utils file?

src/main/scala/org/apache/spark/sql/execution/OAPStrategies.scala
```diff
@@ -0,0 +1,283 @@
...
+        val withProjections = if (projects == withFilter.output) {
+          withFilter
+        } else {
+          execution.ProjectExec(projects, withFilter)
+        }
+
+        withProjections
+    }
+
+    private def getBlockLocations(file: FileStatus): Array[BlockLocation] = file match {
```
adrian-wang: In some utils file?

src/main/scala/org/apache/spark/sql/execution/OAPStrategies.scala
```diff
@@ -0,0 +1,283 @@
...
+          partitionSchema = files.partitionSchema,
+          requiredSchema = prunedDataSchema,
+          filters = pushedDownFilters,
+          order.isAscending,
+          limit,
+          options = files.options,
+          hadoopConf = files.sparkSession.sessionState.newHadoopConfWithOptions(files.options))
+
+        val plannedPartitions = {
+            val defaultMaxSplitBytes = files.sparkSession.sessionState.conf.filesMaxPartitionBytes
```
adrian-wang: Nit: indention

src/main/scala/org/apache/spark/sql/execution/OAPStrategies.scala
```diff
@@ -0,0 +1,283 @@
...
+          }
+          else PlanLater(child)
+        case _ => PlanLater(child)
+      }
+    }
+
+    def pushDownSortToOAPFileScanExec(projects: Seq[NamedExpression], filters: Seq[Expression],
+                                      l: LogicalPlan, files : HadoopFsRelation,
+                                      table: Option[TableIdentifier],
+                                      limit : Int, order : SortOrder) : SparkPlan = {
```
adrian-wang: It seems most of this function is copied from `FileSourceStrategy`. Maybe we should put some of these code to a class, so more oap strategies can use them easily, and code here would look clearer.

src/main/scala/org/apache/spark/sql/execution/OAPStrategies.scala
yueawang: Analyzer checks the value before, a negative value will not be accepted.

src/main/scala/org/apache/spark/sql/execution/OAPStrategies.scala
yueawang: Yes, basically I removed a few code which is useless for OAP and copied the rest here. I think we can extract common logic in future as this time there is no other strategy exist.

src/main/scala/org/apache/spark/sql/execution/OAPStrategies.scala
yueawang: Can be, I will try to find a util file to place them.

src/main/scala/org/apache/spark/sql/execution/OAPStrategies.scala
```diff
@@ -0,0 +1,283 @@
...
+      }
+      case _ => Nil
+    }
+
+    def calcChildPlan(child : LogicalPlan, limit : Int, order : Seq[SortOrder]) : SparkPlan = {
+      child match {
+        case PhysicalOperation(projectList, filters,
+                                relation@LogicalRelation(file: HadoopFsRelation, _, table)) =>
+          val filterAttributes = AttributeSet(ExpressionSet(filters))
+          val orderAttributes = AttributeSet(ExpressionSet(order.map {_.child}))
```
yueawang: what's the diff between these 2? I always mix up them in scala.

src/main/scala/org/apache/spark/sql/execution/OAPStrategies.scala
```diff
@@ -0,0 +1,283 @@
...
+import org.apache.spark.sql.execution.datasources._
+import org.apache.spark.sql.execution.datasources.oap.OapFileFormat
+import org.apache.spark.sql.Strategy
+
+trait OAPStrategies {
+
+  object SortPushDownStrategy extends Strategy with Logging{
+    def apply(plan: LogicalPlan): Seq[SparkPlan] = plan match {
+      case logical.ReturnAnswer(rootPlan) => rootPlan match {
+        case logical.Limit(IntegerLiteral(limit), logical.Sort(order, true, child)) =>
```
yueawang: Looks case class does not obey this rule when which is used for pattern matching. 
I add global = true, but the compiler gives error out.

adrian-wang: you're right

src/main/scala/org/apache/spark/sql/execution/OAPStrategies.scala
adrian-wang: just style, `map()` looks better.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFileFormat.scala
```diff
@@ -188,7 +202,7 @@ private[sql] class OapFileFormat extends FileFormat
             ScannerBuilder.build(supportFilters, ic)
           }
         }
-//        val filterScanner = ic.getScannerBuilder.map(_.build)
+        //        val filterScanner = ic.getScannerBuilder.map(_.build)
```
linhong-intel: It's safe to remove this line.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFileFormat.scala
```diff
@@ -218,26 +232,41 @@ private[sql] class OapFileFormat extends FileFormat
     }
   }
 
+  private def indexHashSetList = {
+    assert(meta.isDefined)
+    val hashSetList = new mutable.ListBuffer[mutable.HashSet[String]]()
+    val bTreeIndexAttrSet = new mutable.HashSet[String]()
+    val bitmapIndexAttrSet = new mutable.HashSet[String]()
+    var idx = 0
+    val m = meta.get
```
linhong-intel: Use `indexHashSetList(m: DataSourceMeta)` as signature, so you can avoid using `val = meta.get`

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/OapDataReaderWriter.scala
```diff
@@ -156,7 +156,9 @@ private[oap] class OapDataReader(
   filterScanner: Option[IndexScanner],
   requiredIds: Array[Int]) extends Logging {
 
-  def initialize(conf: Configuration): Iterator[InternalRow] = {
+  def initialize(conf: Configuration,
+                 ascending: Boolean = true,
+                 limit: Int = 0): Iterator[InternalRow] = {
```
linhong-intel: Just curious about what happen if use sql like `**** LIMIT 0`

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
```diff
@@ -0,0 +1,101 @@
...
+      sql("SELECT b FROM oap_sort_opt_table WHERE a >= 1 AND a <= 10 ORDER BY a LIMIT 4"),
+      Row("this is test 1") ::
+        Row("this is test 2") ::
+        Row("this is test 3") ::
+        Row("this is test 4") :: Nil)
+
+    sql("drop oindex index1 on oap_sort_opt_table")
+    sql("drop oindex index2 on oap_sort_opt_table")
+  }
+
```
linhong-intel: extra newline

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
```diff
@@ -0,0 +1,101 @@
...
+    val dataRDD = spark.sparkContext.parallelize(data, 10)
+
+    dataRDD.toDF("key", "value").createOrReplaceTempView("t")
+    sql("insert overwrite table oap_sort_opt_table select * from t")
+    sql("create oindex index1 on oap_sort_opt_table (a)")
+    sql("create oindex index2 on oap_sort_opt_table (b)")
+
+    checkAnswer(
+      sql("SELECT b FROM oap_sort_opt_table WHERE a >= 1 AND a <= 10 ORDER BY a LIMIT 4"),
+      Row("this is test 1") ::
```
linhong-intel: indention

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
```diff
@@ -0,0 +1,101 @@
...
+
+import org.apache.spark.sql.{QueryTest, Row}
+import org.apache.spark.sql.execution.OAPStrategies
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.test.SharedSQLContext
+import org.apache.spark.util.Utils
+
+case class Source(name: String, age: Int, addr: String, phone: String, height: Int)
+
+class OapPlannerSuite
```
linhong-intel: Please add tests for "ORDER BY only" and "LIMIT only". Since the strategy does not support them and we haven't tested these cases before.

src/main/scala/org/apache/spark/sql/execution/OAPStrategies.scala
```diff
@@ -0,0 +1,302 @@
...
+                                        limit, order.head)
+        }
+        else PlanLater(child)
+      case _ => PlanLater(child)
+    }
+
+
+    /**
+     * Pretty much like FileSourceStrategy.apply() as the only difference is the sort
+     * and limit config were pushed down to FileScanRDD's reader function.
```
linhong-intel: Also add a TODO here to check whether we copied unnecessary code from FileSourceStrategy.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/OapDataReaderWriter.scala
yueawang: limit 0 returns nothing, I remember there are some code checked that condition.

src/main/scala/org/apache/spark/sql/hive/OapSessionState.scala
```diff
@@ -32,6 +33,26 @@ class OapSessionState(sparkSession: OapSession) extends HiveSessionState(sparkSe
...
+  override def planner: SparkPlanner = {
+    new SparkPlanner(sparkSession.sparkContext, conf, experimentalMethods.extraStrategies)
+    with OAPStrategies
+    {
+      override def strategies: Seq[Strategy] = {
+            experimentalMethods.extraStrategies ++ (
+              FileSourceStrategy ::
+              DataSourceStrategy ::
+              DDLStrategy ::
+              SortPushDownStrategy ::
```
yueawang: no special concern, just like the way HiveSessionState does. I need to find another place to initialize extraStrategies if I put this in it.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
yueawang: I think "ORDER BY only" or "LIMIT only" test do not belong to OAP suite, and our strategy so far has nothing to do with them individually. I can add in future if our strategy starts supporting them individually.

linhong-intel: Just want to make sure current strategy has no side effect to "ORDER BY only" and "LIMIT only"

adrian-wang:
hi, can you update this to trigger the test?

adrian-wang:
You need to fix the conflict.

yueawang:
Just updated :)

adrian-wang:
LGTM now.

PR #308, opened 2017-08-03
copy BlockId to OAP

linhong-intel:

We need to modify BlockId use Spark Memory Management.
This is because BlockId is attributed "sealed". So new kind of BlockId must
be in the same file.

This commit doesn't change anything in BlockId.scala. Just a Init copy.

## How was this patch tested?

mvn test


linhong-intel:
I'll merge it by today, since it's just a simple copy.

adrian-wang:
So the modification to this file will be in your follow up pr?

linhong-intel:
> So the modification to this file will be in your follow up pr?

@adrian-wang Yes.
I want to separate spark code and my change into two different PRs. For clearness。

adrian-wang:
sure, lgtm

PR #310, opened 2017-08-04
Add the compression type for the oap data file name

lee-lei:
If the oap is uncompressed, keep the original file name.
Verified that the data file name is changed correctly as the
compression type is changed dynamically.

## How was this patch tested?
1. mvn test
2. changed the compression type at runtime, saved data and checked that the data file name is added with the corresponding compression type.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFileFormat.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFileFormat.scala
@@ -30,7 +30,7 @@ import org.apache.parquet.hadoop.util.{ContextUtil, SerializationUtil}
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.{Row, SparkSession}
 import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.{Expression, JoinedRow}
+import org.apache.spark.sql.catalyst.expressions.{AttributeSet, Expression, JoinedRow, SortOrder}
 import org.apache.spark.sql.catalyst.expressions.codegen.GenerateUnsafeProjection
 import org.apache.spark.sql.execution.datasources._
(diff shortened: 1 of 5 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFileFormat.scala
```diff
@@ -344,11 +344,19 @@ private[oap] class OapOutputWriter(
...
     val isCompressed: Boolean = FileOutputFormat.getCompressOutput(context)
-    val file: Path = new Path(path, getFileName(OapFileFormat.OAP_DATA_EXTENSION))
-    val fs: FileSystem = file.getFileSystem(context.getConfiguration)
+    val conf: Configuration = context.getConfiguration()
+    var compressionFormat: String = ""
+    compressionFormat =
+      conf.get(OapFileFormat.COMPRESSION, OapFileFormat.DEFAULT_COMPRESSION).toString()
+    if (!compressionFormat.isEmpty() && !compressionFormat.matches("UNCOMPRESSED")) {
+      compressionFormat = "." + compressionFormat.toLowerCase()
+    }
```
adrian-wang: 1. I think here we can rewrite the code to avoid having a `var`.
2. if in line 349 you get "UNCOMPRESSED", old file name "a.data" will be "a.UNCOMPRESSEDdata"; and if it is "SNAPPY", the new file name would be "a.snappy.data". Is this expected?

lee-lei: The if statement in line 351 will ensure that the oap data file name will add the compression type only if it will really be  compressed. In other words, it excludes "uncompressed" option. In your case, I think it's expected as long as the oap data will be saved with overwrite mode after the compression type is changed.

linhong-intel:
My suggestion: You can read parquet code as a reference for how to add name extension.

PR #311, opened 2017-08-04
Use OapSession in Test Suite

linhong-intel:

This is similar to SharedSQLContext
How to use:
`class OapIndexQuerySuite extends QueryTest with SharedSQLContext`
change to
`class OapIndexQuerySuite extends QueryTest with OapSharedSQLContext`

Current problem (Solved):
~~~To use OapSession, hive support must be enable. Then table info is stored in disk. `sql("Create Table tablename")` can fail if the table exists~~~


## How was this patch tested?
mvn test


linhong-intel:
This PR is OK to review now.

linhong-intel:
I uploaded another version of TestContext in file "TestOap". Please check if it's the one you want.

adrian-wang:
There are two object TestOAPContext in your code ...

linhong-intel:
close it until I find out a better solution

PR #312, opened 2017-08-06
Release CodecFactory When It's Not Used

linhong-intel:

Modifications:
1. Some Maps in CodecFactory to cache decompressor and compressor
2. Release function in CodecFactory to return decompressor and compressor
3. Close function in OapDataFile to release resources after reading is done
4. Relase CodecFactory after reading/writing is done

## How was this patch tested?

mvn test

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/CodecFactory.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/CodecFactory.scala
@@ -24,16 +24,15 @@ import scala.collection.mutable
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.io.compress.{CodecPool, CompressionCodec}
 import org.apache.hadoop.util.ReflectionUtils
+import org.apache.parquet.format.{CompressionCodec => ParquetCodec}
 import org.apache.parquet.hadoop.metadata.CompressionCodecName
 
 // This is a simple version of parquet's CodeFactory.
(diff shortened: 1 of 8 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/CodecFactory.scala
```diff
@@ -66,11 +89,18 @@ private[oap] class BytesCompressor(compressionCodec: Option[CompressionCodec]) {
       case None => bytes
     }
   }
+
+  def release(): Unit = {
+    if (compressor != null) CodecPool.returnCompressor(compressor)
```
adrian-wang: looks like `returnCompressor` already checked if `compressor` is null. And since `compressor` is a lazy val, I think it can never be `null`, right?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/CodecFactory.scala
```diff
@@ -19,38 +19,61 @@ package org.apache.spark.sql.execution.datasources.oap.io
 
 import java.io.{ByteArrayInputStream, ByteArrayOutputStream, DataInputStream}
 
+import scala.collection.mutable
```
linhong-intel: Because code style checker warning me to use "mutable.HashMap" rather than "HashMap"

adrian-wang:
LGTM now

PR #313, opened 2017-08-09
bypass data file based on filter and column statistics

adrian-wang:

Bypass file reading by looking at column statistics and pushed down filter.

## How was this patch tested?

added a suite

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -249,6 +249,20 @@
       <!--<version>${spark.version}</version>-->
       <version>${spark.internal.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.spark</groupId>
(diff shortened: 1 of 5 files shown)


linhong-intel:
Just FYI.
You can use below code if it helps.
```
minMax = getStatsFromSchema(schema).foreach(_.setMinMaxFromBytes(min, max))
minMax(i).getMin()
```

adrian-wang:
Thanks Linhong, I'll update this soon.

adrian-wang:
@linhong-intel , by using the method you mentioned, StringType data will also be casted to `Binary`

linhong-intel:
I see. then your current is better.

PR #315, opened 2017-08-17
Use logInfo to output more index information for OAP users

lee-lei:
Three categories of information are provided:
1. which filters are supported by OAP index
2. available indexes and which best index will be used
3. how long index, stats and iterator are loaded or constructed

## How was this patch tested?


lee-lei:
So far I haven't found the appropriate place to get the updated value of accumulator variable, since the accumulator value will be finally updated after actions are finished on the executors. The accumulator value may not be deterministic in the transformation body.

lee-lei:
Any suggestions or comments are really appreciated. Thanks.

lee-lei:
The information required is which data will be using index or not. Some data (or tables) may use index. Some may not. Right now it can be reported easily on the executor side. 

Currently there are two approaches to get the above information on the driver side:
1. It requires a call back or some place on driver side to be informed that the actions (queries) are finished on the executor side. In the call back, we can use the accumulator's final value to decide which data was used index or not on the executor side. 
2. On the executor side, during building file scanner in the initialization of OapDataReader, it will be known if index will be used or not. I am not sure if there is any approach to immediately return this kind of information back to the driver.

lee-lei:
Close this PR. Please refer to the PR#318. The link is https://github.com/Intel-bigdata/OAP/pull/318.

PR #317, opened 2017-08-22
Add one configuration for developers to easily ignore index

lee-lei:
This option is for OAP developers to easily analyze the OAP
performance w/ and w/o index even if the index is created already.

Usage:
sqlContext.conf.setConfString(SQLConf.OAP_USE_INDEX.key, value)

The value is true by default. This means that the index will be
used by default. Otherwise, the index won't be used.

## How was this patch tested?
mvn test

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFileFormat.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFileFormat.scala
@@ -275,8 +275,8 @@ private[sql] class OapFileFormat extends FileFormat
 
         hadoopConf.setDouble(SQLConf.OAP_FULL_SCAN_THRESHOLD.key,
           sparkSession.conf.get(SQLConf.OAP_FULL_SCAN_THRESHOLD))
-        hadoopConf.setBoolean(SQLConf.OAP_USE_INDEX.key,
-          sparkSession.conf.get(SQLConf.OAP_USE_INDEX))
+        hadoopConf.setBoolean(SQLConf.OAP_USE_INDEX_FOR_DEVELOPERS.key,
+          sparkSession.conf.get(SQLConf.OAP_USE_INDEX_FOR_DEVELOPERS))
         val broadcastedHadoopConf =
           sparkSession.sparkContext.broadcast(new SerializableConfiguration(hadoopConf))
 
@@ -332,20 +332,24 @@ private[sql] class OapFileFormat extends FileFormat
   }
 
   def hasAvailableIndex(expressions: Seq[Expression]): Boolean = {
-    meta match {
-      case Some(m) =>
-        expressions.exists(m.isSupportedByIndex(_, indexHashSetList))
-      case None => false
-    }
+    if (sparkSession.conf.get(SQLConf.OAP_USE_INDEX_FOR_DEVELOPERS)) {
+      meta match {
+        case Some(m) =>
+          expressions.exists(m.isSupportedByIndex(_, indexHashSetList))
+        case None => false
+      }
+    } else false
   }
 
   def hasAvailableIndex(attributes: AttributeSet): Boolean = {
-    meta match {
-      case Some(m) =>
-        attributes.map{attr =>
-          indexHashSetList.map(_.contains(attr.name)).reduce(_ || _)}.reduce(_ && _)
-      case None => false
-    }
+    if (sparkSession.conf.get(SQLConf.OAP_USE_INDEX_FOR_DEVELOPERS)) {
+      meta match {
+        case Some(m) =>
+          attributes.map{attr =>
+            indexHashSetList.map(_.contains(attr.name)).reduce(_ || _)}.reduce(_ && _)
+        case None => false
+      }
+    } else false
   }
 }
 
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFileFormat.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFileFormat.scala
@@ -334,8 +334,7 @@ private[sql] class OapFileFormat extends FileFormat
   def hasAvailableIndex(expressions: Seq[Expression]): Boolean = {
     (meta, sparkSession.conf.get(SQLConf.OAP_USE_INDEX_FOR_DEVELOPERS)) match {
       case (Some(m), true) => expressions.exists(m.isSupportedByIndex(_, indexHashSetList))
-      case (None, true) => false
-      case (None, false) => false
+      case (None, _) => false
       case (_, false) => false
     }
   }
@@ -345,8 +344,7 @@ private[sql] class OapFileFormat extends FileFormat
       case (Some(m), true) =>
           attributes.map{attr =>
             indexHashSetList.map(_.contains(attr.name)).reduce(_ || _)}.reduce(_ && _)
-      case (None, true) => false
-      case (None, false) => false
+      case (None, _) => false
       case (_, false) => false
     }
   }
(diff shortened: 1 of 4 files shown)


adrian-wang:
LGTM

yueawang:
If OAP_USE_INDEX is disabled, can OAPDataReader report no available index??
Because sometimes upper level queries if index exists and then chooses path accordingly.

adrian-wang:
@lee-lei Can you update this PR and explain @yueawang 's comments?

lee-lei:
Thanks. I will discuss with him. I am not sure if the additional log can meet yue's requirement when index is not used.

lee-lei:
Hi Yue, except the requirements from hasAvailableIndex, do you have any other requirements? Thanks.

yueawang:
LGTM

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/OapDataReaderWriter.scala
```diff
@@ -176,8 +176,15 @@ private[oap] class OapDataReader(
         val dataFileSize = path.getFileSystem(conf).getContentSummary(path).getLength
         val isTesting = conf.getBoolean(SQLConf.OAP_IS_TESTING.key,
                               SQLConf.OAP_IS_TESTING.defaultValue.get)
+        val use_index_for_developers = conf.getBoolean(SQLConf.OAP_USE_INDEX_FOR_DEVELOPERS.key,
```
chenghao-intel: `use_index_for_developers`  variable name should be in camel style.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/OapDataReaderWriter.scala
```diff
@@ -176,8 +176,15 @@ private[oap] class OapDataReader(
...
+                              SQLConf.OAP_USE_INDEX_FOR_DEVELOPERS.defaultValue.get)
         val iter =
-          if (limit <= 0 && indexFileSize > dataFileSize * 0.7 && !isTesting) {
+          // Below is for OAP developers to easily analyze and compare performance without removing
+          // the index after it's created.
+          if (!use_index_for_developers) {
+            logWarning("Index won't be used")
```
chenghao-intel: Instead of saying `Index won't be used`, probably `Using index is disabled by key 'spark.sql.oap.use.index', set it to true to enable the index functionality` is better.

chenghao-intel:
Thank you @lee-lei can you please also add the unit test?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFileFormat.scala
```diff
@@ -330,20 +332,24 @@ private[sql] class OapFileFormat extends FileFormat
...
-        expressions.exists(m.isSupportedByIndex(_, indexHashSetList))
-      case None => false
-    }
+    if (sparkSession.conf.get(SQLConf.OAP_USE_INDEX_FOR_DEVELOPERS)) {
+      meta match {
+        case Some(m) =>
+          expressions.exists(m.isSupportedByIndex(_, indexHashSetList))
+        case None => false
+      }
+    } else false
```
chenghao-intel: Please use `{ }` for `else` block in different lines. Another option is to use the `pattern match` instead.

Probably something like:
```scala
(meta, sparkSession.conf.get(SQLConf.OAP_USE_INDEX_FOR_DEVELOPERS)) match {
  case (_, false) => false
  case (None, true) => false
  case (Some(m), true => expressions.exists(m.isSupportedByIndex(_, indexHashSetList))
}
```

lee-lei:
Will add unit test in the next commit. Thanks.

lee-lei:
There are two approaches to test this option. The first is to test the hadoop configuration which has the expected option value. But this depends that the option by SQLConf will be broadcasted to each executors. The second is to test the iterator, because different option values(true or false) will use the different iterators of OapDataFile. But this behavior is finished inside of initialization of OapDataReader. So far I haven't figured out effective unit test case for this option yet. Thansk.

yueawang:
Just think that query with or without index lead to different number of scan output rows (can be gotten from metrics, but I don't know if it can be easily gotten in spark test framework), maybe you can leverage this property to create a testcase. 
Like below:
// enable index flag
scan oap output row == 10
// disable index flag
scan oap output rot == 300
You can investigate that if necessary.

lee-lei:
Thanks!

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapSuite.scala
```diff
@@ -21,9 +21,15 @@ import java.io.File
 
 import org.scalatest.BeforeAndAfter
 
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.Path
```
adrian-wang: new line after this

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/OapDataReaderWriter.scala
```diff
@@ -175,7 +175,8 @@ private[oap] class OapDataReader(
         val dataFileSize = path.getFileSystem(conf).getContentSummary(path).getLength
         val isTesting = conf.getBoolean(SQLConf.OAP_IS_TESTING.key,
                               SQLConf.OAP_IS_TESTING.defaultValue.get)
-
+        val useIndexForDev = conf.getBoolean(SQLConf.OAP_USE_INDEX_FOR_DEVELOPERS.key,
+                               SQLConf.OAP_USE_INDEX_FOR_DEVELOPERS.defaultValue.get)
```
adrian-wang: indention

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFileFormat.scala
```diff
@@ -312,19 +314,20 @@ private[sql] class OapFileFormat extends FileFormat
   }
 
   def hasAvailableIndex(expressions: Seq[Expression]): Boolean = {
-    meta match {
-      case Some(m) =>
-        expressions.exists(m.isSupportedByIndex(_, indexHashSetList))
-      case None => false
+    (meta, sparkSession.conf.get(SQLConf.OAP_USE_INDEX_FOR_DEVELOPERS)) match {
+      case (Some(m), true) => expressions.exists(m.isSupportedByIndex(_, indexHashSetList))
```
adrian-wang: how about
```
meta match {
  case Some(m) if sparkSession.conf.get(SQLConf.OAP_USE_INDEX_FOR_DEVELOPERS) =>
  case _ =>
}
```

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapSuite.scala
```diff
@@ -100,6 +106,48 @@ class OapSuite extends QueryTest with SharedSQLContext with BeforeAndAfter {
...
+    val itNoIndex = readerNoIndex.initialize(conf)
+    assert (itNoIndex.size == 100)
+    val ic = new IndexContext(dataSourceMeta)
+    val filters: Array[Filter] = Array(
+      And(GreaterThan("a", 9), LessThan("a", 14)))
+    ScannerBuilder.build(filters, ic)
+    var filterScanner = ic.getScanner
+    val readerIndex = new OapDataReader(filePath, dataSourceMeta, filterScanner, requiredIds)
+    val itIndex = readerIndex.initialize(conf)
+    assert (itIndex.size == 4)
```
adrian-wang: no space need between assert and (

src/main/scala/org/apache/spark/sql/internal/SQLConf.scala
```diff
@@ -648,6 +648,12 @@ object SQLConf {
       .booleanConf
       .createWithDefault(false)
 
+  val OAP_USE_INDEX_FOR_DEVELOPERS =
+    SQLConfigBuilder("spark.sql.oap.use.index")
```
adrian-wang: better use `spark.sql.oap.oindex.enabled`
and the val name be `OAP_ENABLE_OINDEX`.
mark it as developer only in the doc.

adrian-wang:
LGTM now, pending tests.

Issue #320, opened 2017-08-31
Bug fix for refresh index


xuanyuanking:
In the scenario of test case 'filtering by string with duplicate refresh' in FilterSuite, the newer insert part of files has no index created by the refresh command. This bug has been fixed in spark-2.1 branch, give a backport for master.

PR #321, opened 2017-08-31
Bug fix for refresh index #320

xuanyuanking:

Fix refresh command bug in isAppend=true

## How was this patch tested?

Exist test 'filtering by string with duplicate refresh'

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexWriter.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexWriter.scala
@@ -59,15 +59,16 @@ private[oap] class BTreeIndexWriter(
     // configuration.set(DATASOURCE_OUTPUTPATH, outputPath)
     if (isAppend) {
       val fs = FileSystem.get(configuration)
-      var skip = true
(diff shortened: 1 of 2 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexWriter.scala
```diff
@@ -59,15 +59,16 @@ private[oap] class BTreeIndexWriter(
     // configuration.set(DATASOURCE_OUTPUTPATH, outputPath)
     if (isAppend) {
       val fs = FileSystem.get(configuration)
-      var skip = true
       var nextFile = InputFileNameHolder.getInputFileName().toString
-      iterator.next()
+      var skip = fs.exists(IndexUtils.indexFileFromDataFile(new Path(nextFile), indexName, time))
+      // iterator.next()
```
adrian-wang: Let's keep this line. if not call `next()` in advance, `nextFile` and `cacheFile` are both empty in the first loop, and in the second loop, `cacheFile` is still empty, but `nextFile` will be something, so the first file cannot be bypassed this way.

adrian-wang:
LGTM, I'm merging this

PR #329, opened 2017-09-18
Index btree interface

linhong-intel:
Draft patch


linhong-intel:
close as it's just for review.

PR #330, opened 2017-09-18
Add OAP aggregation optimization.

yueawang:

1. Support group by aggregation if OAP index exists, so far it does NOT support code gen because group reading is not supported by code gen framework.
2. Rename OAPStrategies -> OapStrategies for name convention.
3. Add IsNotNull() filter if there is no filter being passed down, therefore we can choose which index we want.
4. WIP for index/non-index switch, code gen support.

## How was this patch tested?
mvn test pass.


src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
```diff
@@ -180,4 +176,26 @@ class OapPlannerSuite
...
+
+    val sqlString =
+      "SELECT min(a), max(a) FROM oap_sort_opt_table"
+
+    checkKeywordsExist(sql("explain " ++: sqlString), "*OapAggregationFileScanExec")
+    checkAnswer(sql(sqlString), Seq(Row(1, 300)))
+
+    sql("drop oindex index1 on oap_sort_opt_table")
+  }
+
```
linhong-intel: unnecessary new line

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -148,10 +160,104 @@ trait OAPStrategies extends Logging {
...
+
+  /**
+   * Work as Aggregation in SparkStrategies with OAP optimization.
+   * So far we support Min/Max only because we need extra UDF to do
+   * things like count which is a aggregation result instead of a row
+   * from data source.
+   * Now the only workable case is:
+   * SELECT [min/max](column name) FROM table.
+   */
+  object OapAggregation extends Strategy {
```
linhong-intel: Maybe OapAggregationStrategy

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -85,8 +94,13 @@ trait OAPStrategies extends Logging {
           val oapOption = new CaseInsensitiveMap(file.options +
             (OapFileFormat.OAP_QUERY_LIMIT_OPTION_KEY -> limit.toString) +
             (OapFileFormat.OAP_QUERY_ORDER_OPTION_KEY -> order.head.isAscending.toString))
+          val indexHint = {
+            if (!filters.isEmpty) filters
```
linhong-intel: filters.nonEmpty

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -145,15 +161,110 @@ trait OAPStrategies extends Logging {
...
+    }
+
+    private def calcChildPlan(
+        groupExpressions: Seq[NamedExpression],
+        aggExpressions: Seq[AggregateExpression],
+        resultExpressions: Seq[NamedExpression],
+        child : LogicalPlan
+    ) : SparkPlan = child match {
+      case PhysicalOperation(
+      projectList, filters, relation@LogicalRelation(file: HadoopFsRelation, _, table)) =>
```
linhong-intel: Please consider below change and remove isInstanceOf:
```
case PhysicalOperation(
             projectList, filters, relation@LogicalRelation(file: HadoopFsRelation, _, table)) =>
```
To
```
case PhysicalOperation(projectList, filters,
  relation @ LogicalRelation(
    file @ HadoopFsRelation(_, _, _, _, _: OapFileFormat, _), _, table)) =>
```

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -17,23 +17,32 @@
 
 package org.apache.spark.sql.execution
 
+import scala.collection.mutable.ArrayBuffer
```
linhong-intel: unused

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -145,15 +161,110 @@ trait OAPStrategies extends Logging {
...
+   *   FROM table
+   *   [WHERE filter on columns]
+   *   GROUP BY one specific column
+   */
+  object OapGroupAggregateStrategy extends Strategy {
+    def apply(plan: LogicalPlan): Seq[SparkPlan] = plan match {
+      case PhysicalAggregation(
+      groupingExpressions, aggregateExpressions, resultExpressions, child) =>
+
+        val (functionsWithDistinct, functionsWithoutDistinct) =
```
linhong-intel: `functionsWithoutDistinct` is unused

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
linhong-intel: In this file, I suggest to add some comment to indicate code copied from spark.
For example:
```
// --- Start: Code Copied From Spark, File Name: ---
Some Code
// --- End: Code Copied From Spark ---
```

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
```diff
@@ -0,0 +1,214 @@
...
+
+package org.apache.spark.sql.execution.aggregate
+
+import org.apache.spark.TaskContext
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions._
+import org.apache.spark.sql.catalyst.expressions.aggregate._
+import org.apache.spark.sql.catalyst.expressions.codegen.GenerateUnsafeRowJoiner
+import org.apache.spark.sql.execution.{UnsafeFixedWidthAggregationMap, UnsafeKVExternalSorter}
```
linhong-intel: unused

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
```diff
@@ -0,0 +1,214 @@
...
+import org.apache.spark.sql.execution.metric.SQLMetric
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.unsafe.KVIterator
```
linhong-intel: unused

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
```diff
@@ -0,0 +1,214 @@
...
+    resultExpressions,
+    newMutableProjection) with Logging {
+
+  ///////////////////////////////////////////////////////////////////////////
+  // Part 1: Initializing aggregate functions.
+  ///////////////////////////////////////////////////////////////////////////
+
+  // Remember spill data size of this task before execute this operator so that we can
+  // figure out how many bytes we spilled for this operator.
+  private val spillSizeBefore = TaskContext.get().taskMetrics().memoryBytesSpilled
```
linhong-intel: unused. Is this variable copied from spark?

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
```diff
@@ -0,0 +1,214 @@
...
+      super.generateResultProjection()
+    }
+  }
+
+  // An aggregation buffer containing initial buffer values. It is used to
+  // initialize other aggregation buffers.
+  private[this] val initialAggregationBuffer: UnsafeRow = createNewAggregationBuffer()
+  private[this] val groupAggregationBuffer: UnsafeRow = initialAggregationBuffer.copy()
+
+  var currGroupHead : InternalRow = null
```
linhong-intel: convert `null` to `_`

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
```diff
@@ -175,7 +179,7 @@ private[oap] object ScannerBuilder extends Logging {
         else combineIntervalMaps(leftMap, rightMap, ic, needMerge = true)
     )
 
-    if (intervalMap != null) {
+    if (!intervalMap.isEmpty) {
```
linhong-intel: I'm not sure if `intervalMap != null` is necessary. If you aren't sure either, change to
```
if (intervalMap != null && intervalMap.nonEmpty)
```

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -85,8 +94,13 @@ trait OAPStrategies extends Logging {
           val oapOption = new CaseInsensitiveMap(file.options +
             (OapFileFormat.OAP_QUERY_LIMIT_OPTION_KEY -> limit.toString) +
             (OapFileFormat.OAP_QUERY_ORDER_OPTION_KEY -> order.head.isAscending.toString))
+          val indexHint = {
+            if (!filters.isEmpty) filters
+            else IsNotNull(orderAttributes.head) :: Nil
```
adrian-wang: This is to trigger queries without filters?

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -129,7 +143,9 @@ trait OAPStrategies extends Logging {
         if joinType == LeftSemi && canBroadcast(right) =>
         Seq(joins.BroadcastHashJoinExec(
           leftKeys, rightKeys, joinType, BuildRight, condition, planLater(left),
-          calcChildPlan(right, rightKeys.map{SortOrder(_, Ascending)})))
+          calcChildPlan(right, rightKeys.map {
```
adrian-wang: the map here is single line function, maybe use `rightKeys.map(SortOrder(_, Ascending))`

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
yueawang: good suggestion! Thanks!

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -145,15 +161,110 @@ trait OAPStrategies extends Logging {
...
+   * Now the workable case is:
+   *   SELECT [agg](columns)
+   *   FROM table
+   *   [WHERE filter on columns]
+   *   GROUP BY one specific column
+   */
+  object OapGroupAggregateStrategy extends Strategy {
+    def apply(plan: LogicalPlan): Seq[SparkPlan] = plan match {
+      case PhysicalAggregation(
+      groupingExpressions, aggregateExpressions, resultExpressions, child) =>
```
adrian-wang: break here should indent by 4

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
yueawang: intervalMap != null can not happen here as it comes from filters which is an in param of this function. If filters == null, function exits immediately. So here condition should be isEmpty rather than == null

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -145,15 +161,110 @@ trait OAPStrategies extends Logging {
...
+
+        val (functionsWithDistinct, functionsWithoutDistinct) =
+          aggregateExpressions.partition(_.isDistinct)
+        if (functionsWithDistinct.map(_.aggregateFunction.children).distinct.length > 1) {
```
adrian-wang: use `.nonEmpty`

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
yueawang: removed, yes it was, and last clean commit missed this one

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
yueawang: Before I thought agg with distinct would be added soon, so I kept it. let me remove it now and add it back if necessary in future.

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
yueawang: ok

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
yueawang: normally I added comment in the code (usually a function) beginning, because the copy part might be scattered.

yueawang:
Just wonder you review the code on web or local??

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
adrian-wang: maybe we should consider in index selection part, so that we don't have to create such a fake filter here. Or we create a specific filter for OAP use.

adrian-wang: maybe the name `indexHint` is a little bit weird...

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
linhong-intel: OK, then. intervalMap.nonEmpty will be better.

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
linhong-intel: I checkout `FileSourceStrategy`. It's indent by 2
```
object FileSourceStrategy extends Strategy with Logging {
  def apply(plan: LogicalPlan): Seq[SparkPlan] = plan match {
    case PhysicalOperation(projects, filters,
      l @ LogicalRelation(_fsRelation: HadoopFsRelation, _, table)) =>
```

linhong-intel:
local
________________________________

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
adrian-wang: even in spark code , indent 2 and 4 is random in this case. anyway, here indent by 0 is uncommon.

yueawang: It is indented automatically by IDE, seems there is no standard code style config for scala.

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
yueawang: Yes, this is a way to tell the scanner to use index which is binding on filter. Without change the interface, either fake filter (easier, as this commit implemented) or extra option (need to parse name to attribute in scanner) can be used for this purpose.
Finally this logic will be combined with index selection.

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
```diff
@@ -0,0 +1,208 @@
...
+  ///////////////////////////////////////////////////////////////////////////
+  // Part 7: Iterator's public methods.
+  ///////////////////////////////////////////////////////////////////////////
+
+  override final def hasNext: Boolean = {
+    currGroupHead != null
+  }
+
+  override final def next(): UnsafeRow = {
+    if (hasNext) {
```
chenghao-intel: Usually in `Iterator.next()`, we don't need to check the `hasNext`, as we assume the caller will check the `hasNext` before call the `next()`.

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
```diff
@@ -0,0 +1,208 @@
...
+  if (currGroupHead == null && inputIter.hasNext) {
+    currGroupHead = inputIter.next()
+  }
+
+  ///////////////////////////////////////////////////////////////////////////
+  // Part 7: Iterator's public methods.
+  ///////////////////////////////////////////////////////////////////////////
+
+  override final def hasNext: Boolean = {
+    currGroupHead != null
```
chenghao-intel: I am wondering if we can simplify the code by removing the `currGroupHead` in this class, and here (`hasNext`) only check the `inputIter.hasNext()`.

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  override final def next(): UnsafeRow = {
+    if (hasNext) {
+      val res = {
+        val aggBuffer : (UnsafeRow, UnsafeRow) = processOneGroupInputs()
+        // We did not fall back to sort-based aggregation.
+        val result = generateOutput(aggBuffer._1, aggBuffer._2)
+        result.copy()
```
chenghao-intel: I think probably it's the caller's responsiblity to decide if it's really need to `row.copy()`, otherwise, we may have duplicated `copy()` of the result row, can you double confirm this?

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
```diff
@@ -0,0 +1,208 @@
...
+      super.generateResultProjection()
+    }
+  }
+
+  // An aggregation buffer containing initial buffer values. It is used to
+  // initialize other aggregation buffers.
+  private[this] val initialAggregationBuffer: UnsafeRow = createNewAggregationBuffer()
+  private[this] val groupAggregationBuffer: UnsafeRow = initialAggregationBuffer.copy()
+
+  var currGroupHead : InternalRow = _
```
chenghao-intel: We probably can simplify the implementation by removing the member variable...can we?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
```diff
@@ -144,6 +144,10 @@ private[oap] object ScannerBuilder extends Logging {
       case LessThan(attribute, ic(key)) =>
         val ranger = new RangeInterval(IndexScanner.DUMMY_KEY_START, key, true, false)
         mutable.HashMap(attribute -> ArrayBuffer(ranger))
+      case IsNotNull(attribute) =>
```
chenghao-intel: Can you confirm is this related to this PR?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/DataSourceMeta.scala
```diff
@@ -42,12 +42,10 @@ import org.apache.spark.sql.types._
  *    .
  *    .
  * FileMeta N
- * IndexMeta 1      -- 768 bytes
```
chenghao-intel: can you please confirm if we really have to change it in this PR? or we can do it in an separated PR?

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -129,33 +141,129 @@ trait OAPStrategies extends Logging {
...
       case _ => Nil
     }
 
     def calcChildPlan(
         child: LogicalPlan,
         order: Seq[SortOrder]): SparkPlan = child match {
       case PhysicalOperation(
-      projectList, filters, relation@LogicalRelation(file: HadoopFsRelation, _, table)) =>
+      projectList, filters, relation@LogicalRelation(
+      file @ HadoopFsRelation(_, _, _, _, _ : OapFileFormat, _), _, table)) =>
```
chenghao-intel: indent

README.md
```diff
@@ -7,7 +7,7 @@ OAP - Optimized Analytics Package (previously known as Spinach) is designed to a
 mvn -DskipTests package
 ```
 ## Prerequisites
-You should have [Apache Spark](http://spark.apache.org/) of version 2.0.2 installed in your cluster. Refer to Apache Spark's [documents](http://spark.apache.org/docs/2.0.2/) for details.
```
chenghao-intel: Separated PR please.

yueawang: This is 2.0.2->2.1.0 change which is already merged, I had rebased this PR before and I thought it should have been gone, let me double check.

src/main/scala/org/apache/spark/storage/BlockId.scala
```diff
@@ -114,6 +114,7 @@ object BlockId {
   val TASKRESULT = "taskresult_([0-9]+)".r
   val STREAM = "input-([0-9]+)-([0-9]+)".r
   val TEST = "test_(.*)".r
+  val FIBER = "fiber_(.*)".r
```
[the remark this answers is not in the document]
yueawang: code already merged, I think it should have gone rebase.

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
yueawang: Yes, that's better! :)

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -129,33 +141,129 @@ trait OAPStrategies extends Logging {
...
+    }
+
+    private def calcChildPlan(
+        groupExpressions: Seq[NamedExpression],
+        aggExpressions: Seq[AggregateExpression],
+        resultExpressions: Seq[NamedExpression],
+        child : LogicalPlan
+    ) : SparkPlan = child match {
+      case PhysicalOperation(
+      projectList, filters, relation@LogicalRelation(
```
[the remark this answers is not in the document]
yueawang: changed.

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
yueawang: changed.

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -166,11 +274,15 @@ trait OAPStrategies extends Logging {
...
-      projects: Seq[NamedExpression], filters: Seq[Expression],
-      l: LogicalRelation, _fsRelation : HadoopFsRelation,
+      projects: Seq[NamedExpression],
+      filters: Seq[Expression],
+      l: LogicalPlan,
+      _fsRelation: HadoopFsRelation,
       table: Option[CatalogTable],
-      OapOption: Map[String, String]): Option[SparkPlan] = {
-    // Filters on this relation fall into four categories based on where we can use them to avoid
+      OapOption: Map[String, String],
```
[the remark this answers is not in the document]
yueawang: changed!

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregateExec.scala
```diff
@@ -0,0 +1,135 @@
...
+    "spillSize" -> SQLMetrics.createSizeMetric(sparkContext, "spill size"),
+    "aggTime" -> SQLMetrics.createTimingMetric(sparkContext, "aggregate time"))
+
+  override def output: Seq[Attribute] = resultExpressions.map(_.toAttribute)
+
+  override def outputPartitioning: Partitioning = child.outputPartitioning
+
+  // This is for testing. We force TungstenAggregationIterator to fall back to the unsafe row hash
+  // map and/or the sort-based aggregation once it has processed a given number of input rows.
+  private val testFallbackStartsAt: Option[(Int, Int)] = None
```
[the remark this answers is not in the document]
yueawang: Deleted, as this one is for compatible usage. We don't need it anymore.

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregateExec.scala
```diff
@@ -0,0 +1,135 @@
...
+        }
+      }
+    }
+  }
+
+  override def verboseString: String = toString(verbose = true)
+
+  override def simpleString: String = toString(verbose = false)
+
+  private def toString(verbose: Boolean): String = {
```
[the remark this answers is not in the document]
yueawang: This follows the origin Spark code, I just remove irrelevant code inside the function.
Why not change toString??

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
yueawang: This is a variable who saves the start point of next group and used to tell if current group is completed. Actually, I learned the similiar logic from Spark's sort hash. Is var a performance penalty to scala??

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
yueawang: The reason I don't use inputIter.hasNext is that the group head is already read out to tell if the previous group is completed, once the group head is the last element, the inputIter has no next, but the last group is not processed. So we need to process group until currGroupHead is null.

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
yueawang: Here the caller's next means next group, we need a wrapper hasNext call to differentiate these 2 situations.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/OapDataReaderWriter.scala
```diff
@@ -217,18 +217,19 @@ private[oap] class OapDataReader(
...
         if (options.contains(OapFileFormat.OAP_INDEX_SCAN_NUM_OPTION_KEY)) {
           fs.setScanNumLimit(
-            options.get(OapFileFormat.OAP_INDEX_SCAN_NUM_OPTION_KEY).get.toInt
+            options(OapFileFormat.OAP_INDEX_SCAN_NUM_OPTION_KEY).toInt
           )
         }
 
-        val isFastIndexQuery : Boolean =
-          limit > 0 || options.contains(OapFileFormat.OAP_INDEX_SCAN_NUM_OPTION_KEY)
+        val forceIndexScan : Boolean = limit > 0 ||
```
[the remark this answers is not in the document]
yueawang: Sure, changed.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
```diff
@@ -180,4 +183,34 @@ class OapPlannerSuite
...
+    checkKeywordsExist(sql("explain " + sqlString), "*OapAggregationFileScanExec")
+    val oapDF = sql(sqlString)
+
+    spark.experimental.extraStrategies = Nil
+    checkKeywordsNotExist(sql("explain " + sqlString), "OapAggregationFileScanExec")
+    val baseDF = sql(sqlString)
+
+    checkAnswer(oapDF, baseDF)
+
+    sql("drop oindex index1 on oap_fix_length_schema_table")
```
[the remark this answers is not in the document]
yueawang: OK, I can do a suite refactory later.

src/main/scala/org/apache/spark/storage/BlockId.scala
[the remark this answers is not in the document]
yueawang: The reason should be the force rebase of the master which happens by mistake last night.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
yueawang: It is for query who uses group by without filter. Actually it is a intersect part between optimization strategy and index selection. but if I don't add it. the only workable case is to select rows with group by and filter on the same column, which is a rare case.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/DataSourceMeta.scala
yueawang: No, this is because the mis-update on master

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
yueawang: I think it's hard to tell the behavior of caller. so far the caller/shuffle has both paths(copy & non-copy) and its own logic to decide which way. But I did not find flag to control that. So this copy is a kind of fall safe.
Maybe I can add a reverse logic here to switch between them.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
chenghao-intel: Is it a bug in master? if it's, then probably we need to fix that first before this PR been merged.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/DataSourceMeta.scala
chenghao-intel: I see, most likely a document bug in master, let's do that in a separated PR, and merge that PR first. Sounds good?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
adrian-wang: this is not a bug, but a workaround to enable
`select x from t order by x limit n` to be a fake `select x from t where x is not null order by x limit n`

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
chenghao-intel: Yes, I know, `sorted record iterator` --> `group iterator`. Usually in Java/Scala `Iterator` implementation, we don't need to check the `hasNext`, we should assume it always `hasNext=true`. And in the other hand, we need to guarantee the `group iterator`.`hasNext` has the right mean(is it has more groups?)

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
chenghao-intel: No performance penalty, I just wondering if we can simplify the code by removing the extra internal status.

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregateExec.scala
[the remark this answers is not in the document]
chenghao-intel: Every class in Java/Scala has the default function called `def toString(): String`, if we overwrite this method, probably cause confusion. anyway, this is not mandatory.

src/main/scala/org/apache/spark/storage/BlockId.scala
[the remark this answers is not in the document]
chenghao-intel: OK, I see, seems github doesn't detect the code conflict.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/DataSourceMeta.scala
yueawang: Yes, it was merged, I think I need to rebase again to get rid of these changes.

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
yueawang: OK, I can rename it to hasNextGroup to make it more meaningful.

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
chenghao-intel: As I remember, the principle is the parent operator should knows if it needs to cache the row, if it's true, then it will call the `.copy()`. but sometimes bugs there, ideally we should change the code in the parent operators.

you're right `.copy()` is safe, but may cause performance penalty, can you please add a comment there? and leave it for the further optimization.

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
chenghao-intel: Just talked with @yueawang offline, the OapAggregateIterator needs prefeching one record to check if it's the end of the aggregate group. So it's reasonable to me. thank you @yueawang for the explanation. I am OK with this code change.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
chenghao-intel: @adrian-wang what's your opinion? should we change the code in this PR?

adrian-wang: @chenghao-intel I think by redesign index selection we can avoid changes here, but for now it is ok to keep.

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
```diff
@@ -0,0 +1,205 @@
...
+  private def processOneGroupInputs(): (UnsafeRow, UnsafeRow) = {
+    // Init agg buffer for a new group.
+    groupAggregationBuffer.copyFrom(initialAggregationBuffer)
+
+    // Set the group key and process the first row in current group.
+    val currGroupingKey = groupingProjection.apply(currGroupHead).copy()
+    processRow(groupAggregationBuffer, currGroupHead)
+
+    var isSameGroup = true
+    while(isSameGroup && inputIter.hasNext) {
```
linhong-intel: Is it possible to change to 
```
inputIter.takeWhile(i => currGroupingKey.equals(groupingProjection.apply(i)))
        .foreach(processRow(groupAggregationBuffer, _))
```

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -129,33 +139,125 @@ trait OAPStrategies extends Logging {
...
       case _ => Nil
     }
 
     def calcChildPlan(
         child: LogicalPlan,
         order: Seq[SortOrder]): SparkPlan = child match {
-      case PhysicalOperation(
-      projectList, filters, relation@LogicalRelation(file: HadoopFsRelation, _, table)) =>
+      case PhysicalOperation(projectList, filters,
+        relation@LogicalRelation(
```
linhong-intel: Nit: add space before and after @

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -129,33 +139,125 @@ trait OAPStrategies extends Logging {
...
         val filterAttributes = AttributeSet(ExpressionSet(filters))
         val orderAttributes = AttributeSet(ExpressionSet(order.map(_.child)))
         if ((orderAttributes.size == 1 &&
-          (filterAttributes.isEmpty || filterAttributes == orderAttributes)) &&
-          (file.fileFormat.isInstanceOf[OapFileFormat])) {
+          (filterAttributes.isEmpty || filterAttributes == orderAttributes))) {
           val oapOption = new CaseInsensitiveMap(file.options +
             (OapFileFormat.OAP_INDEX_SCAN_NUM_OPTION_KEY -> "1"))
+
+          val indexHint = {
```
linhong-intel: brace `{` and `}` is unnecessary.

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -75,18 +82,23 @@ trait OAPStrategies extends Logging {
...
-            projectList, filters, relation@LogicalRelation(file: HadoopFsRelation, _, table)) =>
+      case PhysicalOperation(projectList, filters,
+        relation @ LogicalRelation(
+          file @ HadoopFsRelation(_, _, _, _, _ : OapFileFormat, _), _, table)) =>
         val filterAttributes = AttributeSet(ExpressionSet(filters))
         val orderAttributes = AttributeSet(ExpressionSet(order.map(_.child)))
         if ((orderAttributes.size == 1 &&
-            (filterAttributes.isEmpty || filterAttributes == orderAttributes)) &&
-            (file.fileFormat.isInstanceOf[OapFileFormat])) {
+            (filterAttributes.isEmpty || filterAttributes == orderAttributes))) {
```
linhong-intel: you have one unnecessary pair of `()`

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -129,33 +139,125 @@ trait OAPStrategies extends Logging {
...
+            // TODO: support distinct in future.
+            Nil
+          }
+
+        aggregateOperator
+
+      case _ => Nil
+    }
+
+    private def calcChildPlan(
```
linhong-intel: All `calcChildPlan` are similar. Please add a TODO here to abstract the common code.

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
yueawang: you mean call takeWhile inside the while loop? I think takeWhile could stop correctly on the first row of next group, but I guess it does not save that row, so it may lead to omit one row between groups. Just guess, I can double confirm that.

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
yueawang: OK, Yes, these 3 are very much alike. 
According to my previous design (if exist :P) all strategies implement their own childPlan method. So I wrote 3 similiar functions which actually do the FileStrategy's job (a plan decide how to scan a file) with different options.
 In future, I think we eventually have a OapFileScan strategy, at that time I think this similarity will be gone, as childPlan will be calc by Spark automatically (if possible)
So in my mind, if we have more similar optimizations in near future, I can abstract the common code then.
Otherwise, we can add OapFileStrategy to get a complete oap optimization strategy module.

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
yueawang: Looks unacceptable.
scala> val a = 1 to 20
a: scala.collection.immutable.Range.Inclusive = Range(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20)
scala> val iter = a.iterator
iter: Iterator[Int] = non-empty iterator
scala> iter.takeWhile(_ <= 10).map(_.toString).toArray
res9: Array[String] = Array(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
scala> iter.next
res10: Int = 12 <-- Here 11 is skipped because of condition check, I am not sure if there is a way to catch this failure row to do something (update group key) we wanted.

linhong-intel: Keep your change then.

adrian-wang:
Let's merge this to avoid conflicts.

PR #332, opened 2017-09-22
Another way to resolve table identifier

linhong-intel:

CreateIndex is a sub class of LeafNode which has no children.
But we use the children to resolve table identifer. This is simple but weird.

So, here is a another way to do this.

## How was this patch tested?

Unit tests


linhong-intel:
@yueawang @adrian-wang 
Hi, This is a proposal to eliminate overriding `children` and `containChild`

Similar code is in `DescribeTableCommand` and `AnalyzeTableCommand`

Just a proposal, we can still keep current one.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -20,19 +20,21 @@ package org.apache.spark.sql.execution.datasources.oap.index
 import org.apache.hadoop.fs.Path
 import org.apache.hadoop.mapreduce.Job
 
-import org.apache.spark.SparkException
 import org.apache.spark.internal.Logging
+import org.apache.spark.SparkException
```
adrian-wang: wrong order

adrian-wang:
LGTM

PR #344, opened 2017-10-19
allow write byte in oap index output format

adrian-wang:

refine the code and add byte written support to oap index output format. This is part 1 of #328 .

## How was this patch tested?

existed tests.


adrian-wang:
close this since #340 removed this.

PR #345, opened 2017-10-23
BTree Index Writer with Partial Loading

linhong-intel:

BTree Index Writer

## How was this patch tested?

Unit test

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
@@ -65,9 +65,12 @@ private[index] class BTreeIndexRecordWriter(
       val order = keySchema.zipWithIndex.map {
         case (field, index) => SortOrder(
           BoundReference(index, field.dataType, nullable = true),
-          if (field.metadata.getBoolean("isAscending")) Ascending else Descending)
+          if (!field.metadata.contains("isAscending") || field.metadata.getBoolean("isAscending")) {
+            Ascending
+          } else {
+            Descending
+          })
       }
-
       GenerateOrdering.generate(order, keySchema.toAttributes)
     }
 
@@ -270,7 +273,7 @@ private[index] class BTreeIndexRecordWriter(
         val nodeUniqueKeys = uniqueKeys.slice(startPosInKeyList, startPosInKeyList + keyCount)
         val rowCount = nodeUniqueKeys.map(multiHashMap.get(_).size()).sum
 
-        val nodeBuf = serializeNode(nodeUniqueKeys, startPosInKeyList)
+        val nodeBuf = serializeNode(nodeUniqueKeys, startPosInRowList)
         fileWriter.writeNode(nodeBuf)
         startPosInKeyList += keyCount
         startPosInRowList += rowCount
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
@@ -316,6 +316,7 @@ private[index] class BTreeIndexRecordWriter(
     buffer.toByteArray ++ keyBuffer.toByteArray
   }
 
+  // TODO: BTreeNode can be re-write. It doesn't carry any values.
   private def sumKeyCount(node: BTreeNode): Int = {
     if (node.children.nonEmpty) node.children.map(sumKeyCount).sum
     else node.root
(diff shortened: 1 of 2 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
```diff
@@ -230,4 +240,182 @@ private[index] class BTreeIndexRecordWriter(
...
+      output.writeInt(rowPos)
+      writeBasedOnSchema(keyOutput, key, keySchema)
+      rowPos += multiHashMap.get(key).size()
+    }
+    buffer.toByteArray ++ keyBuffer.toByteArray
+  }
+
+  private def sumKeyCount(node: BTreeNode): Int = {
+    if (node.children.nonEmpty) node.children.map(sumKeyCount).sum
+    else node.root
```
yueawang: Just thought a int called root sounds weird, is it meaningful??

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
```diff
@@ -24,19 +24,22 @@ import scala.collection.JavaConverters._
...
 import org.apache.spark.sql.execution.datasources.oap.io.IndexFile
 import org.apache.spark.sql.execution.datasources.oap.statistics.StatisticsManager
 import org.apache.spark.sql.execution.datasources.oap.utils.{BTreeNode, BTreeUtils}
-import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.types._
 
 private[index] class BTreeIndexRecordWriter(
     configuration: Configuration,
-    writer: OutputStream,
+    writer: FSDataOutputStream,
```
yueawang: Any reason to use sub-class FSDataOutputStream?

linhong-intel: `FSDataOutputStream.getPos` is helpful.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
linhong-intel: Yes, it's weird~~~. This is from the legacy code. So we have to refactor current code step by step to make a good project.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
```diff
@@ -230,4 +240,182 @@ private[index] class BTreeIndexRecordWriter(
...
+   * Min Key Pos in Key Data          4 Bytes
+   * Max Key Pos in Key Data          4 Bytes
+   * Key Data - Variable Bytes      M
+   * Min Key For Child #1 - Min
+   * Max Key For Child #1
+   * ...
+   * Min Key For Child #N
+   * Max Key For Child #N - Max
+   */
+  private def serializeFooter(nodes: Seq[BTreeNodeMetaData]): Array[Byte] = {
```
yueawang: i see there is a private case class BTreeFooter(buf: Array[Byte]) to de-serialize the footer, just think ser/de-ser in same obj looks better.

linhong-intel: In our design, This Footer's member is accessed like: `Platform.getInt(buf, offset)` since the data is in offHeap memory. So needn't and can't use an object to store footer.

yueawang: No, I mean a class to store ser/de-ser logic, as you already  have a BTreeFooter which does the de-ser job.

linhong-intel: OK, I got your idea.Actually I have thought about it.
But since we don't de-ser a real object (it's a buffer and set of get function), It looks a lit strange. Anyway, it's a good idea to put them together, I will think about it.
```
object FooterDerSer {
  def serialize(obj: FooterObj): Array[Byte] => { ... }
  // Good deserialize function
  def deserialize(buf: Array[Byte]): FooterObj => { ... }
  // Actually we need below function
  def deserializeWrapper(buf: Array[Byte]):
    (buffer: array[Byte], getFunc: () => Obj) = { ... }
}
```

linhong-intel:
This PR should be good to merge. I plan to merge it to reduce pending PRs.

PR #348, opened 2017-10-23
BTree Index with Partial Loading Integration

linhong-intel:

Based on #345 #346 #347
Remove dead code

## How was this patch tested?

Unit Test.


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
```diff
@@ -0,0 +1,258 @@
...
+  private def binarySearch(
+      start: Int, length: Int,
+      keys: Int => Key, candidate: Key,
+      compare: (Key, Key) => Int): (Int, Boolean) = {
+
+    var s = 0
+    var e = length - 1
+    var found = false
+    var m = s
+    while (s <= e & !found) {
```
yueawang: A rare exception happens here if e --> Int.MaxValue, as m = e/2, then (e + e/2) overflows.
I think we could add a few assert/warning sentences either here or in caller.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
```diff
@@ -39,177 +33,21 @@ private[oap] class BPlusTreeScanner(idxMeta: IndexMeta) extends IndexScanner(idx
   var currentKeyIdx = 0
   var indexFiber: IndexFiber = _
   var indexData: CacheResult = _
+  var recordReader: BTreeIndexRecordReader = _
 
   def initialize(dataPath: Path, conf: Configuration): IndexScanner = {
```
yueawang: IndexScanner is now Iterator[Long], Had we been talking on change it to Int?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
```diff
@@ -39,177 +33,21 @@ private[oap] class BPlusTreeScanner(idxMeta: IndexMeta) extends IndexScanner(idx
...
-      } else {
-        while (m<node.length-1 && order.compare(node.keyAt(m + 1), candidate) == 0)
-          m += 1
-      }
-    }
-
-    if (node.isLeaf) {
-      // here currentKey is equal to candidate or the last key on the left side
-      // which is less than the candidate
-      val currentKey = new CurrentKey(node, m, 0, getLimitScanNum())
```
yueawang: getLimitScanNum related logic can be moved to indexScanner, but I think this PR could skip this.
Can you help to disable the OapSemiJoin Strategy? Just comment it out in oapStrategies with a TODO.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
linhong-intel: Yes, we need to change Long to Int to save space. But this can only be done after all Indexes are merged. Or there will be compile error.

PR #350, opened 2017-10-23
Bitmap index partial loading r/w implementation

lee-lei:

Based on #349. The read and write and organized into one unified patch, since both of them are not big changes and independent to other modules.

Below is the bitmap index general layout and sections.
#section id || section size(B) || section description
      1            ||    4                  ||    header
      2            ||   varied           ||     sorted unique key list (index column unique values)
      3            ||   varied           ||     bitmap entry list
      4            ||   varied           ||     bitmap entry offset list
      5            ||   varied           ||     keep the original statistics, not changed than before.
      6            ||   40                 ||     save total key list size and length, total entry list size and
                                                 total offset list size, and also the original footer(not changed than before).

Several key features are below:
    1. refine the bitmap index file layout.
    2. cache the bitmap index each section after first loading.
    3. Use more efficient data structure to reduce both index file size and query time.
    3. refactor bitmap index r/w logic to be modularity-friendly.

The performance result will be uploaded as #351.

## How was this patch tested?
mvn clean test package

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapScanner.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapScanner.scala
@@ -19,9 +19,13 @@ package org.apache.spark.sql.execution.datasources.oap.index
 import java.io.{ByteArrayInputStream, ObjectInputStream}
 
 import scala.collection.mutable
+import scala.collection.immutable
+import scala.collection.JavaConverters._
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
+import org.roaringbitmap.buffer.ImmutableRoaringBitmap
+import org.roaringbitmap.buffer.MutableRoaringBitmap
 import sun.nio.ch.DirectBuffer
 
 import org.apache.spark.sql.catalyst.expressions.codegen.GenerateOrdering
@@ -37,9 +41,9 @@ private[oap] case class BitMapScanner(idxMeta: IndexMeta) extends IndexScanner(i
 
   override def canBeOptimizedByStatistics: Boolean = true
 
-  @transient var internalItr: Iterator[Int] = Iterator[Int]()
+  @transient var internalItr: Iterator[Integer] = Iterator[Integer]()
   var empty: Boolean = _
-  var internalBitSet: BitSet = _
+  var internalBitSet: MutableRoaringBitmap = _
   var indexFiber: IndexFiber = _
   var indexData: CacheResult = _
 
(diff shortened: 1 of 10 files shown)


linhong-intel:
**Partial Loading** means DO NOT load the whole index but only load needed parts. (For example: Only load header plus needed entries of bitmap)
But if you don't change below code:
`indexData = FiberCacheManager.getOrElseUpdate(indexFiber, conf)`
The index file will always be loaded as a whole, so there will be no partial loading

lee-lei:
Yes. You are right. I noticed that. The first level is to partially load one or several parts of the index file. This is beneficial for both disk and network IO. The second level is to partially load the specific bitmap entries whithin the bitmap index file, which are implemented already. This is beneficial for network IO. The first level is actually the common code or interfaces used by both BTree and Bitmap indexes.  I regarded it as part of common interfaces before. Thanks.

linhong-intel:
So, please implement the real partial loading. This is not part of common interface.

chenghao-intel:
Should we rebase the code first? I saw lots of unnecessary code changes here, which is probably in the master branch already.

linhong-intel:
Hi Lei,
If you don't remove below lines, partial loading will never be achieved.
```
indexFiber = IndexFiber(indexFile)
indexData = FiberCacheManager.getOrElseUpdate(indexFiber, conf)
```
The purpose of partial loading is to avoid loading unnecessary data from disk.
But the 2 lines code will read ALL data from disk. It's contradictory to partial loading.

Another thing is:
Bitmap should not support range scan. This will make your code looks simple.

lee-lei:
Thanks for your suggestion. For the first thing, my rathional is that if we can cache either by off-heap memory or on heap memory(getOrElseUpdate returned true), we just directly load from the cache; if it can't cache successfully(getOrElseUpdate returned false), then we switch to partial loading. Of course, if we want to use partial loading by default, that's totally OK to me.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapScanner.scala
```diff
@@ -59,82 +64,182 @@ private[oap] case class BitMapScanner(idxMeta: IndexMeta) extends IndexScanner(i
...
+        bitmapArray.foreach(bm => internalBitSet.or(bm))
+        internalItr = internalBitSet.iterator.asScala.toIterator
       }
       empty = false
     } else {
       empty = true
     }
   }
 
+  def partiallyLoadFromFile(indexPath: Path, conf: Configuration): Unit = {
```
chenghao-intel: Originally, my thought is to load the large of data block into offheap memory, and then using the `Platform.getXXX` API, which seems more efficient and can avoid the GC (smaller objects), and most importantly, we can leverage the potential Intel's new memory technology.

So please DO think about how to cache the loaded index/data in memory, not just parsing the source data directly from the file by Java/HDFS I/O API.

lee-lei: Sure. No problem. Thanks for your suggestion! I may not understand well your thoughts before. I will add the fine cache code to separately fine cache key list, bitmap list and offset list based on the existing partial loading soon. Thanks.

lee-lei:
The updated version with cache will be uploaded soon. Thanks.

lee-lei:
The code is finished. The debugging is WIP. Thanks.

lee-lei:
Some minor finalization will be updated soon. Thanks.

adrian-wang:
Can you divide this into smaller PRs and have some of them merged early?

lee-lei:
just rebased, since #349 is updated. Thanks.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapScanner.scala
```diff
@@ -37,104 +42,250 @@ private[oap] case class BitMapScanner(idxMeta: IndexMeta) extends IndexScanner(i
...
+
+  private var bmUniqueKeyListFiber: BitmapFiber = _
+  private var bmUniqueKeyListCache: CacheResult = _
+  private var bmUniqueKeyListBuffer: Array[Byte] = _
+
+  private var bmOffsetListFiber: BitmapFiber = _
+  private var bmOffsetListCache: CacheResult = _
+  private var bmOffsetListBuffer: Array[Byte] = _
+
+  private var bmEntryListFiber: BitmapFiber = _
```
yueawang: There are so many var, but I find most of them are immutable after the first assignment. Would val better??

lee-lei: Right now every query will call the initialize function of index scanner, since it's not guaranteed that the index file is not changed between the consecutive queries. That indicates that each initialization will assign these global variables. Actually, in most cases, in the same context, the queries are many times while the index change is not many. The possible opportunity in the future is to add a flag or whatever to check if the index file is changed or not and bypass these global virables reassignment if not changed. Thanks.

linhong-intel:
If you can split this into several small PRs, that will be better

lee-lei:
The above check failure log shows that mvn install unexpectedly failed. My local build is successful. Thanks.

lee-lei:
The check error details shows below error:
bigdata/OAP/src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapMicroBenchmarkSuite.scala:52: value OAP_IS_TESTING is not a member of object org.apache.spark.sql.internal.SQLConf
[ERROR]     sqlContext.conf.setConf(SQLConf.OAP_IS_TESTING, true)
I am not sure if the master is impacted or not. Thanks.

adrian-wang:
This error is because @yueawang  has removed the config of OAP_IS_TESTING in master branch. Can you update it?

yueawang:
Yes, ths IS_TESTING flag has no effect after index selection PR merged, so I removed it. We will have more fine grit flags in future. For now you can rebase to last master and update the commit.

lee-lei:
I see. Thanks.

linhong-intel:
merged to unblock 0.3.0 release QA

PR #352, opened 2017-10-24
 BTree Index Partial Loading with Fiber Cache

linhong-intel:

BTree Index Partial Loading with Fiber Cache

## How was this patch tested?

Unit Test

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -65,11 +65,7 @@ object FiberCacheManager extends Logging {
   private val dataFileIdMap = new TimeStampedHashMap[String, DataFile](updateTimeStampOnGet = true)
 
   private def toByteBuffer(buf: Array[Byte]): ChunkedByteBuffer = {
-    // TODO: make it configurable
-    val cbbos = new ChunkedByteBufferOutputStream(buf.length, ByteBuffer.allocate)
-    cbbos.write(buf)
-    cbbos.close()
-    cbbos.toChunkedByteBuffer
+    new ChunkedByteBuffer(ByteBuffer.wrap(buf))
   }
 
   def fiber2Block(fiber: Fiber): BlockId = {
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -17,6 +17,7 @@
 
 package org.apache.spark.sql.execution.datasources.oap.filecache
 
+import java.nio.ByteBuffer
 import java.util.concurrent.TimeUnit
 
 import scala.collection.mutable
@@ -34,7 +35,7 @@ import org.apache.spark.storage.{BlockId, FiberBlockId, StorageLevel}
 import org.apache.spark.unsafe.Platform
 import org.apache.spark.util.TimeStampedHashMap
 import org.apache.spark.util.collection.BitSet
-import org.apache.spark.util.io.ChunkedByteBuffer
+import org.apache.spark.util.io.{ChunkedByteBuffer, ChunkedByteBufferOutputStream}
 
 
 // TODO need to register within the SparkContext
@@ -63,6 +64,14 @@ object FiberCacheManager extends Logging {
 
   private val dataFileIdMap = new TimeStampedHashMap[String, DataFile](updateTimeStampOnGet = true)
 
+  private def toByteBuffer(buf: Array[Byte]): ChunkedByteBuffer = {
+    // TODO: make it configurable
+    val cbbos = new ChunkedByteBufferOutputStream(buf.length, ByteBuffer.allocate)
+    cbbos.write(buf)
+    cbbos.close()
+    cbbos.toChunkedByteBuffer
+  }
+
   def fiber2Block(fiber: Fiber): BlockId = {
     fiber match {
       case DataFiber(file, columnIndex, rowGroupId) =>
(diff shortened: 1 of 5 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -63,6 +64,14 @@ object FiberCacheManager extends Logging {
 
   private val dataFileIdMap = new TimeStampedHashMap[String, DataFile](updateTimeStampOnGet = true)
 
+  private def toByteBuffer(buf: Array[Byte]): ChunkedByteBuffer = {
+    // TODO: make it configurable
+    val cbbos = new ChunkedByteBufferOutputStream(buf.length, ByteBuffer.allocate)
```
chenghao-intel: `new ChunkedByteBuffer(ByteBuffer.wrap(buf))`? Should we copy the data from `buf`?

chenghao-intel: or `new ChunkedByteBuffer(ByteBuffer.allocateDirect(buf.length).put(buf, 0, buf.length))`, but this may cause the direct memory not in well managed.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -211,3 +223,10 @@ case class DataFiber(file: DataFile, columnIndex: Int, rowGroupId: Int) extends
 
 private[oap]
 case class IndexFiber(file: IndexFile) extends Fiber
+
+private[oap]
+case class BTreeFiber(
+    getFiberData: () => Array[Byte],
```
chenghao-intel: Can we use the `()=> ByteBuffer` instead?

chenghao-intel: As Array[Byte] probably cause additional data copy and GC overhead.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexFileReader.scala
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+package org.apache.spark.sql.execution.datasources.oap.index
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.Path
+
+import org.apache.spark.unsafe.Platform
+
+private[oap] case class BTreeIndexFileReader(
```
chenghao-intel: Probably we'd better to put the IndexFile layout as comment in the source code.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
```diff
@@ -0,0 +1,303 @@
...
+    val values = schema.map(_.dataType).map { dataType =>
+      val (value, length) = readBasedOnDataType(baseObj, pos, dataType)
+      pos += length
+      value
+    }
+    InternalRow.fromSeq(values)
+  }
+
+  private def readBasedOnDataType(baseObj: Object, offset: Long, dataType: DataType) = {
+
```
chenghao-intel: remove the empty line or even:
```scala
private def readBasedOnDataType(baseObj: Object, offset: Long, dataType: DataType) = dataType match {
  ...
}
```

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
```diff
@@ -87,147 +87,183 @@ private[index] class BTreeIndexRecordWriter(
...
     }
-    writer.write(keyBuf.toByteArray)
-    subOffset += keyBuf.size
-    map.put(writeList.head, updateOffset)
-    var rmCount = tree.root
-    while (rmCount > 1) {
-      rmCount -= 1
-      keysList.remove(keysList.size - listOffsetFromEnd - rmCount)
+    buffer.toByteArray ++ keyBuffer.toByteArray
+
```
chenghao-intel: Nit: remove the empty line

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
```diff
@@ -87,147 +87,183 @@ private[index] class BTreeIndexRecordWriter(
...
+        writer.write(bytes)
+      case BinaryType =>
+        val bytes = row.getBinary(ordinal)
+        writer.writeInt(bytes.length)
+        writer.write(bytes)
+      case _ => throw new OapException("not support data type")
+    }
   }
 }
+private case class
```
chenghao-intel:
```scala
private case class BTreeNodeMetaData(
  rowCount: Int, byteSize: Int, min: InternalRow, max: InternalRow)
```

src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeMultiColumnSearchSuite.scala
```diff
@@ -1,339 +0,0 @@
-/*
```
chenghao-intel: Will those unit test will be added back in later PRs?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
```diff
@@ -0,0 +1,303 @@
...
+    reader.close()
+  }
+
+  /**
+   * TODO: if this hasNext doesn't reach false, the resource can't be released
+   * For example:
+   *   Assume recordReader.size = 100, Someone called `recordReader.take(10)`.
+   */
+  override def hasNext: Boolean = {
+    if (internalIterator.hasNext) true
```
chenghao-intel: Nit: if we used the `{}` in `else` block, we'd better to use it in the `if` block as well.

linhong-intel:
Thanks @chenghao-intel 
1. Unit test is WIP, I was thinking about uploading another PR for test. and I'll add back the removed tests
2. Memory do need some changes. But I prefer a bit more discussion.

chenghao-intel:
Yes, let's leave the memory things for the further PR(s), most likely we need to use(integrated into) the Spark existing memory management mechanism.

PR #353, opened 2017-10-24
 fix a bug when read from hdfs

adrian-wang:

When we create an oap folder in hdfs, and read it back, then in index output writer, we would produce a task output dir like /path/to/dst//temporary/_0/hdfs://server:port/partVal=some, which is not as expected.

```
Caused by: java.io.IOException: Mkdirs failed to create file:/testOap/_temporary/0/_temporary/attempt_20171024022302_0001_m_000000_0hdfs:/10.0.2.40:9000/testOap
               at org.apache.hadoop.fs.ChecksumFileSystem.create(ChecksumFileSystem.java:438)
               at org.apache.hadoop.fs.ChecksumFileSystem.create(ChecksumFileSystem.java:424)
               at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:905)
               at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:886)
               at org.apache.hadoop.fs.FileSystem.create(FileSystem.java:783)
               at org.apache.spark.sql.execution.datasources.oap.index.OapIndexOutputFormat.getRecordWriter(OapIndexOutputFormat.scala:58)
               at org.apache.spark.sql.execution.datasources.oap.index.OapIndexOutputWriter.initWriter(OapIndexOutputWriter.scala:78)
               at org.apache.spark.sql.execution.datasources.oap.index.OapIndexOutputWriter.checkStartOfNewFile(OapIndexOutputWriter.scala:94)
               at org.apache.spark.sql.execution.datasources.oap.index.OapIndexOutputWriter.writeInternal(OapIndexOutputWriter.scala:66)
               at org.apache.spark.sql.execution.datasources.FileFormatWriter$SingleDirectoryWriteTask.execute(FileFormatWriter.scala:248)

```

## How was this patch tested?

add a test in IndexUtilsSuite

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexUtils.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexUtils.scala
@@ -27,7 +27,7 @@ import org.apache.spark.sql.execution.datasources.oap.OapFileFormat
 /**
  * Utils for Index read/write
  */
-object IndexUtils {
+private[oap] object IndexUtils {
   // TODO remove this and corresponding test
   def writeInt(out: OutputStream, v: Int): Unit = {
     out.write((v >>>  0) & 0xFF)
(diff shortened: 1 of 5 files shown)


adrian-wang:
Thanks @Moonlit-Sailor for reporting this issue.

yueawang:
Is this always happen?? I did not meet this issue.

adrian-wang:
```
val spark = SparkSession.builder().master("local").appName("testSql").getOrCreate()
    import spark.implicits._
    val df = spark.sparkContext.parallelize(1 to 100, 3).map(i => (i, i + 100, s"this is row $i"))
      .toDF("a", "b", "c")

    df.write.format("oap").mode(SaveMode.Overwrite).save("hdfs://10.0.2.40:9000/testOap")

    val oapDf = spark.read.format("oap").load("hdfs://10.0.2.40:9000/testOap")
    oapDf.registerTempTable("t")

    spark.sql("create oindex indexa on t(a)")
    spark.sql("show oindex from t").show()
//    spark.sql("drop oindex ida on t")
    val result = spark.sql("select * from t where a<7")
    result.show()

```

linhong-intel:
Thanks, Daoyuan. Look good to me.
I filed a issue #357 to track the disabled case.

PR #358, opened 2017-10-27
add DataSourceStrategy with like support

adrian-wang:

This is to enable spark push down STRING_LIKE operator

## How was this patch tested?

N/A

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
@@ -1,0 +1,626 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


adrian-wang:
If we extend it, it would be much simplified, but need to merge with OAPStrategies. @yueawang

chenghao-intel:
Maybe you can talk to @yueawang offline, and figure out if there any simple solution to push down more predictions into the data source, this is probably a common issues for the further optimization in OAP.

adrian-wang:
Actually I only changed the `def translateFilters` in the big file. Should be easy if extend.

yueawang:
Yes, If we create a base scan strategy class for OAP e.g. OapFileStrategy, we can control all these things in it.
This Daoyuan's change adds a case branch to spark which does not give a interface to register such case, that's why he need to extract out & modify the DataSourceStrategy from spark.
We can discuss later if we are able to implement the OapFileStrategy now.

src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala
```diff
@@ -0,0 +1,626 @@
...
+  private[this] def toCatalystRDD(relation: LogicalRelation, rdd: RDD[Row]): RDD[InternalRow] = {
+    toCatalystRDD(relation, relation.output, rdd)
+  }
+
+  /**
+   * Tries to translate a Catalyst [[Expression]] into data source [[Filter]].
+   *
+   * @return a `Some[Filter]` if the input [[Expression]] is convertible, otherwise a `None`.
+   */
+  protected[sql] def translateFilter(predicate: Expression): Option[Filter] = {
```
adrian-wang: The function I want to modify is in an scala object, it seems I cannot derive an object from another...

adrian-wang:
I'll merge this to unblock future work.

PR #361, opened 2017-10-29
[TRIE] utils to build trie in memory

adrian-wang:

build trie index in memory, will use as util when build trie.

## How was this patch tested?

add unit test.


adrian-wang:
I'm merging this in

PR #365, opened 2017-11-01
use Integer.SIZE / 8 instead of Integer.BYTE

adrian-wang:

as we talked in the meeting, Integer.BYTE is JVM 8 only, so use Integer.SIZE / 8 instead.

## How was this patch tested?

existing test.


adrian-wang:
merging this since this is quite simple and already discussed

PR #366, opened 2017-11-01
[Trie] trie read/write

adrian-wang:

implement read/write for trie index

## How was this patch tested?

unit test


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -78,6 +78,17 @@ case class CreateIndex(
     // this may impact index updating. It may also fail index existence check. Should put index
     // info at table level also.
     val time = System.currentTimeMillis().toHexString
+    val innerIndexType =
```
adrian-wang: I'll file a separate PR for this change...

PR #367, opened 2017-11-01
[bug] should use actual index type to invoke proper index writer

adrian-wang:

Since we may create permuterm index when user specifying btree index, we should use actual index type to invoke proper index writer

## How was this patch tested?

unit tests.


adrian-wang:
I'll merge this by end of today.

PR #368, opened 2017-11-01
fix FileMeta's data file name of Parquet

Moonlit-Sailor:
to make `DataFileName` field in meta file of Parquet file format only store data file's name, excluding path.

## How was this patch tested?
unit test

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/OapIndexOutputWriter.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/OapIndexOutputWriter.scala
@@ -85,7 +85,8 @@ private[index] class OapIndexOutputWriter(
       recordWriter.close(context)
       recordWriter = null
       results = results :+
-          IndexBuildResult(inputFileName, rowCount, "", new Path(inputFileName).getParent.toString)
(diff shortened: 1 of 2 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/OapIndexOutputWriter.scala
```diff
@@ -85,7 +85,8 @@ private[index] class OapIndexOutputWriter(
       recordWriter.close(context)
       recordWriter = null
       results = results :+
-          IndexBuildResult(inputFileName, rowCount, "", new Path(inputFileName).getParent.toString)
+        IndexBuildResult(new Path(inputFileName).getName,
+          rowCount, "", new Path(inputFileName).getParent.toString)
```
adrian-wang: Nit: better break around `(` instead of `,` or `:+`.

linhong-intel:
code is good to me. But what's the purpose @Moonlit-Sailor ?

adrian-wang:
@linhong-intel when the data format is Oap, the filename appears to only contain the filename, while for parquet, the filename contains the full uri. Since the meta file has already contained the path to the folder, we only need to put the filename in filemeta.

PR #374, opened 2017-11-07
Check integrity(index,data,meta) for specified partition(s)

Moonlit-Sailor:
Modify DML to enable checking for specified partition(s):
use
`CHECK OINDEX ON table PARTITION(partitionVal1, partitionVal2,...,partitionValN)`
 to check integrity of specified partition(s)

## How was this patch tested?
OapCheckIndexSuite

### DIFF
--- a/src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapCheckIndexSuite.scala
+++ b/src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapCheckIndexSuite.scala
@@ -331,4 +331,123 @@ class OapCheckIndexSuite extends QueryTest with SharedSQLContext with BeforeAndA
       "\nAmbiguous Index(different indices have the same name):\nindex name:idx1"))
 
   }
+
+  test("check index on partitioned table for a specified partition: Missing meta file") {
+    val data: Seq[(Int, Int)] = (1 to 10).map { i => (i, i) }
+    data.toDF("key", "value").createOrReplaceTempView("t")
+
+    sql(
+      """
+        |INSERT OVERWRITE TABLE oap_partition_table
+        |partition (b=1, c='c1')
+        |SELECT key from t where value < 4
+      """.stripMargin)
+
+    sql(
+      """
+        |INSERT INTO TABLE oap_partition_table
+        |partition (b=2, c='c2')
+        |SELECT key from t where value == 4
+      """.stripMargin)
+
+    val partitionPath =
+      new Path(spark.sqlContext.conf.warehousePath + "/oap_partition_table/b=2/c=c2")
+    checkAnswer(
+      sql("check oindex on oap_partition_table partition(b=2, c='c2')"),
+      Row(s"Meta file not found in partition: ${partitionPath.toUri.getPath}"))
+
+    sql("create oindex idx1 on oap_partition_table(a) partition(b=2, c='c2')")
+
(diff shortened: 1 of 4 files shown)


src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapCheckIndexSuite.scala
```diff
@@ -331,4 +331,123 @@ class OapCheckIndexSuite extends QueryTest with SharedSQLContext with BeforeAndA
...
+
+    val partitionPath =
+      new Path(spark.sqlContext.conf.warehousePath + "/oap_partition_table/b=2/c=c2")
+    // Delete a data file
+    val metaOpt = OapUtils.getMeta(sparkContext.hadoopConfiguration, partitionPath)
+    assert(metaOpt.nonEmpty)
+    assert(metaOpt.get.fileMetas.nonEmpty)
+
+    val dataFileName = metaOpt.get.fileMetas.head.dataFileName
+    Utils.deleteRecursively(new File(new Path(partitionPath, dataFileName).toUri.getPath))
```
linhong-intel: What will happen if I delete one entire partition dir? Can we detect this?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapCheckIndexSuite.scala
```diff
@@ -331,4 +331,123 @@ class OapCheckIndexSuite extends QueryTest with SharedSQLContext with BeforeAndA
...
+        |INSERT INTO TABLE oap_partition_table
+        |partition (b=2, c='c2')
+        |SELECT key from t where value == 4
+      """.stripMargin)
+
+    val partitionPath =
+      new Path(spark.sqlContext.conf.warehousePath + "/oap_partition_table/b=2/c=c2")
+    checkAnswer(
+      sql("check oindex on oap_partition_table partition(b=2, c='c2')"),
+      Row(s"Meta file not found in partition: ${partitionPath.toUri.getPath}"))
```
linhong-intel: If we don't create index, there is no meta file? I'm not sure this is expected. @adrian-wang

Moonlit-Sailor: For Parquet file format, there is no meta file if we don't create index previously.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapCheckIndexSuite.scala
Moonlit-Sailor: Currently, this will not be detected.
Some perspectives are listed to explain this issue:
- There is no global meta(table meta)  to keep track of every partition that the table has. Just maintain a plain directory path structure in file system.
- What comprises a table in physical layer? Within Spark SQL, each subdirectory of the table's rootpath(s) will be listed to find the data files when scanning the table. Then all data files under the rootpath(s) provided by table's catalog will be taken into account. These data files along with their subdirectories consist of the whole table. If we delete an entire subdirectory, then the deleted subdirectory does not belong to the table anymore, both in logical sense and in physical sense.
- `check oindex on table` command is to check integrity of indices not data. Nonetheless, it will check integrity of data files together with index when meta exists. And in the follow-up pr #382, all subdirectories that has meta files will be extracted as target paths for checking integrity. Thus missing data will be detected when all data files are removed but meta file is left behind in the subdirectory.

PR #394, opened 2017-11-15
fix bug when creating dup named index

Moonlit-Sailor:
Fixed a bug when creating duplicated named index.
`create oindex idxa on table(column)`
`create oindex if not exists idxa on table(column)`

to prevent from creating duplicated named index.

## How was this patch tested?
OapDDLSuite

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
@@ -107,6 +107,7 @@ case class CreateIndex(
               s"""Index $indexName exists on ${identifier.getOrElse(parent)}""")
           } else {
             logWarning(s"Dup index name $indexName")
+            return Nil
           }
         }
         if (existsData != null) existsData.foreach(metaBuilder.addFileMeta)
--- a/src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapDDLSuite.scala
+++ b/src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapDDLSuite.scala
@@ -165,5 +165,28 @@ class OapDDLSuite extends QueryTest with SharedSQLContext with BeforeAndAfterEac
       new Configuration()).globStatus(new Path(path,
       "oap_partition_table/b=2/c=c2/*.index")).length != 0)
   }
+
+  test("create duplicated name index") {
+    val data: Seq[(Int, String)] = (1 to 100).map { i => (i, s"this is test $i") }
+    val df = data.toDF("a", "b")
+    val pathDir = Utils.createTempDir().getAbsolutePath
+    df.write.format("oap").mode(SaveMode.Overwrite).save(pathDir)
+    val oapDf = spark.read.format("oap").load(pathDir)
+    oapDf.createOrReplaceTempView("t")
+    sql("create oindex idxa on t (a)")
+    val path = new Path(pathDir)
+    val fs = path.getFileSystem(sparkContext.hadoopConfiguration)
+    val indexFiles1 = fs.listStatus(path).collect { case fileStatus if fileStatus.isFile &&
+      fileStatus.getPath.getName.endsWith(OapFileFormat.OAP_INDEX_EXTENSION) =>
+        fileStatus.getPath.getName
+    }
+
+    sql("create oindex if not exists idxa on t (a)")
+    val indexFiles2 = fs.listStatus(path).collect { case fileStatus if fileStatus.isFile &&
+      fileStatus.getPath.getName.endsWith(OapFileFormat.OAP_INDEX_EXTENSION) =>
+      fileStatus.getPath.getName
+    }
+    assert(indexFiles1 === indexFiles2)
+  }
 }
 


adrian-wang:
LGTM

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -107,6 +107,7 @@ case class CreateIndex(
               s"""Index $indexName exists on ${identifier.getOrElse(parent)}""")
           } else {
             logWarning(s"Dup index name $indexName")
+            return Nil
```
chenghao-intel: Instead of return `Nil`, I would suggest to return some meaningful `string`s

chenghao-intel: Hence people can get the Error message from the console instead of immediate return without any explanation.

Moonlit-Sailor: add result display in console

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -107,6 +110,7 @@ case class CreateIndex(
               s"""Index $indexName exists on ${identifier.getOrElse(parent)}""")
           } else {
             logWarning(s"Dup index name $indexName")
+            return Seq(Row(s"Creating Dup index name:$indexName on table:${table.identifier}"))
```
chenghao-intel: Something like `Failed to create index for xxx as it has existed already.` :)

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
adrian-wang: This is a runable command, and it is a ddl, the standard output should be empty.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
adrian-wang: We should refer to MySQL for return...

adrian-wang: MySQL does not allow `if not exists` in `create index`, while for `create table` it does provide `create table if not exists`.

I tried it, looks like:
```
mysql> create table if not exists t1 (a1 int);
Query OK, 0 rows affected (0.10 sec)

mysql> create table if not exists t1 (a1 int);
Query OK, 0 rows affected, 1 warning (0.00 sec)

mysql> create table if not exists t1 (a1 int);
Query OK, 0 rows affected, 1 warning (0.00 sec)
```

adrian-wang: and
```
mysql> create table t1 (a1 int);
ERROR 1050 (42S01): Table 't1' already exists
```

chenghao-intel: And unit test for it.

chenghao-intel: `CREAT INDEX IF NOT EXISTS` doesn't related to this PR, right? I would suggest to create another PR if you feel we have to conform to MYSQL.

adrian-wang: Why not? This PR is to solve the case if user use `CREATE INDEX IF NOT EXISTS` while the index exists.

adrian-wang: I think return Nil here is everything we need.

chenghao-intel: if we returns Nil here, which means nothing to return in driver side, and will cause confusing if index build was not triggered.

adrian-wang: There's already a warning level log indicating that. And this is exactly how mysql handle this.

chenghao-intel: `CREATE INDEX IF NOT EXISTS` is a "feature" exists for some time, even if we need to remove it, let's do this in a seperated PR.

I am thinking this PR is focus on even the index name exists, it still will trigger the index build, which is not intended.

chenghao-intel: warning level log will not shows in `thriftserver mode` while connected with client like beeline.

adrian-wang: I don't think we are removing any feature in this pr. this pr is just fixing a bug that we create dup index, but `if not exists` means we just create when no index of this name presents.

chenghao-intel: then you need to remove the DDL support right?

adrian-wang: By using `if not exists`, the user should know that if it already exists, nothing will happen.
If user want to make sure the index is created, he should not use `if not exists`

adrian-wang: No, I am saying I insist on just return `Nil` here. Return any informative string here is not right for DDL. There's no other DDL actually returns anything. Just return Nil or throw exception.

chenghao-intel: Agree to throw exception. Logically:

I think it was originally by design allow user to create the index with given existing name. the difference probably one should throws exception, but the other just keep quiet. depends on how you make it more friendly.

adrian-wang: I wrote initial version of this part of code, the `if not exists` clause is for avoiding exception throwing, and do nothing(happens to have a bug here). If exception thrown, `if not exists` has no meaning at all. If we have to throw exception for this case, then we'd better just remove the feature.

chenghao-intel: OK, got it, let's revert to the first commit. thanks for the explanation.

Moonlit-Sailor:
Ok, now when creating duplicated name index using `create oindex if not exists indexName on table(column)`, there will be a warning log;
And when creating duplicated name index using `create oindex indexName on table(column)`, then will throw an exception

PR #395, opened 2017-11-15
Make bitmap index only support equal query

lee-lei:
1. Filter out range intervals before building bitmap scanner
2. If no equal value interval, bypass bitmap scanner
3. Add one test case to verify the above assumption
## How was this patch tested?
mvn clean test package

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
@@ -187,17 +187,29 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
 
         } // end for
       case BitMapIndex(entries) =>
-        keySchema = new StructType().add(meta.schema(entries(lastIdx)))
-        scanner = BitMapScanner(bestIndexer)
         val attribute = meta.schema(entries(lastIdx)).name
         val filterOptimizer = unapply(attribute).get
-        scanner.intervalArray =
+        val sortedIntervalArray =
           intervalMap(attribute).sortWith(filterOptimizer.compareRangeInterval)
+        val singleValueIntervalArray =
+          sortedIntervalArray.filter(filterOptimizer.isSingleValueInterval)
+        // Make sure that each interval is really equal query.
+        singleValueIntervalArray.foreach(interval => {
+          assert(interval.start == interval.end)
+        })
+        if (singleValueIntervalArray.nonEmpty) {
+          keySchema = new StructType().add(meta.schema(entries(lastIdx)))
+          scanner = BitMapScanner(bestIndexer)
+          logDebug("Bitmap index only supports equal query.")
+          scanner.intervalArray = singleValueIntervalArray
+        }
       case _ =>
     }
 
-    logDebug("Index Scanner Intervals: " + scanner.intervalArray.mkString(", "))
-    scanner.withKeySchema(keySchema)
+    if (scanner != null && keySchema != null) {
+      logDebug("Index Scanner Intervals: " + scanner.intervalArray.mkString(", "))
+      scanner.withKeySchema(keySchema)
+    }
   }
 
   def unapply(attribute: String): Option[FilterOptimizer] = {
@@ -223,6 +235,9 @@ private[oap] object DummyIndexContext extends IndexContext(null) {
 
 private[oap] class FilterOptimizer(keySchema: StructType) {
   val order = GenerateOrdering.create(keySchema)
+
+  def isSingleValueInterval(interval: RangeInterval): Boolean =
+    interval.start == interval.end && interval.startInclude && interval.endInclude
 
   // compare two intervals
   def compareRangeInterval(interval1: RangeInterval, interval2: RangeInterval): Boolean = {
--- a/src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexSelectionSuite.scala
+++ b/src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexSelectionSuite.scala
@@ -208,4 +208,20 @@ class IndexSelectionSuite extends SharedSQLContext with BeforeAndAfterEach{
     ScannerBuilder.build(filters, ic)
     assertIndexer(ic, 3, "idxACD", 2)
   }
+
+  test("Bitmap Scanner only supports equal query.") {
+    sql("create oindex idxa on oap_test(a) USING BITMAP")
+    val oapMeta = DataSourceMeta.initialize(path, new Configuration())
+    val ic = new IndexContext(oapMeta)
+    val nonEqualFilters: Array[Filter] = Array(
+      Or(And(GreaterThan("a", 9), LessThan("a", 14)),
+        And(GreaterThan("a", 23), LessThan("a", 166)))
+    )
+    ScannerBuilder.build(nonEqualFilters, ic)
+    // Non equal filters won't build BitmapScanner.
+    assert(ic.getScanner == None)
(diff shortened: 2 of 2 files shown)


lee-lei:
Please help review this PR. Thanks for your time.

lee-lei:
@adrian-wang @linhong-intel @yueawang Thanks.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -224,6 +230,11 @@ private[oap] object DummyIndexContext extends IndexContext(null) {
 private[oap] class FilterOptimizer(keySchema: StructType) {
   val order = GenerateOrdering.create(keySchema)
 
+  def isSingleValueInterval(interval: RangeInterval): Boolean = {
+    if (interval.start == interval.end && interval.startInclude && interval.endInclude) true
```
yueawang: return interval.start == interval.end && interval.startInclude && interval.endInclude

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -187,17 +187,23 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
...
-        keySchema = new StructType().add(meta.schema(entries(lastIdx)))
-        scanner = BitMapScanner(bestIndexer)
         val attribute = meta.schema(entries(lastIdx)).name
         val filterOptimizer = unapply(attribute).get
-        scanner.intervalArray =
+        val sortedIntervalArray =
           intervalMap(attribute).sortWith(filterOptimizer.compareRangeInterval)
+        val singleValueIntervalArray =
+          sortedIntervalArray.filter(filterOptimizer.isSingleValueInterval)
+        if (singleValueIntervalArray.size > 0) {
```
yueawang: use nonEmpty??

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -224,6 +230,11 @@ private[oap] object DummyIndexContext extends IndexContext(null) {
 private[oap] class FilterOptimizer(keySchema: StructType) {
   val order = GenerateOrdering.create(keySchema)
```
chenghao-intel: Will this be used?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -187,17 +187,23 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
...
+          scanner = BitMapScanner(bestIndexer)
+          logDebug("Bitmap index only supports equal query.")
+          scanner.intervalArray = singleValueIntervalArray
+        }
       case _ =>
     }
 
     logDebug("Index Scanner Intervals: " + scanner.intervalArray.mkString(", "))
-    scanner.withKeySchema(keySchema)
+    if (scanner != null && keySchema != null) scanner.withKeySchema(keySchema)
```
chenghao-intel: Is it possible for `scanner == null`? Is it supposed to set the keySchema while we create the scanner.

chenghao-intel: https://github.com/Intel-bigdata/OAP/pull/395/files#diff-c1beda8555c50f88113fa60658645d47R205

chenghao-intel: https://github.com/Intel-bigdata/OAP/pull/395/files#diff-c1beda8555c50f88113fa60658645d47L199

chenghao-intel:
Can you please add a unit test to verify weather the BitMapIndex will be triggered even if we specified a Range search?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
chenghao-intel: I think we need to refactor the `IndexContext` or `IndexScanner`, `ordering` or `intervalArray`, `pattern` probably not necessary to put into the base class.

@linhong-intel @adrian-wang

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
lee-lei: This is common for all the index types. Therefore it's better to be after building the specific index type as the above. Thanks.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -187,17 +187,23 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
...
       case BitMapIndex(entries) =>
-        keySchema = new StructType().add(meta.schema(entries(lastIdx)))
-        scanner = BitMapScanner(bestIndexer)
         val attribute = meta.schema(entries(lastIdx)).name
         val filterOptimizer = unapply(attribute).get
-        scanner.intervalArray =
+        val sortedIntervalArray =
           intervalMap(attribute).sortWith(filterOptimizer.compareRangeInterval)
+        val singleValueIntervalArray =
+          sortedIntervalArray.filter(filterOptimizer.isSingleValueInterval)
```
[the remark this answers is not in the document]
lee-lei: This is the entry point to build the specific OAP index types. Before here, interval maps is already built, combined and optimized. I agreee with you to assert all the filtered intervals to be single value intervals before building Bitmap Scanner. Thanks.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -224,6 +236,10 @@ private[oap] object DummyIndexContext extends IndexContext(null) {
 private[oap] class FilterOptimizer(keySchema: StructType) {
   val order = GenerateOrdering.create(keySchema)
 
+  def isSingleValueInterval(interval: RangeInterval): Boolean = {
+    return interval.start == interval.end && interval.startInclude && interval.endInclude
```
yueawang: return keyword is unnecessary.
def isSingleValueInterval(interval: RangeInterval) =
    interval.start == interval.end && interval.startInclude && interval.endInclude

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
yueawang: From original bitmap concept's point of view, maybe hash is better than ordering, especially now you support only equal query.
You may consider it and refine in future.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
lee-lei: It's more concise. As the public method, it needs explicit return type. So we can just remove the return keyword and keep Boolean keyword. Thanks.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
lee-lei: Thanks.

lee-lei:
@adrian-wang @linhong-intel Please help review and merge this PR if it's ready. Thanks.

PR #403, opened 2017-11-17
use readFully instead of read for HDFS file read

adrian-wang:

use readFully instead of read for HDFS file read.
According to 
https://stackoverflow.com/questions/12184975/why-hadoop-api-fsdatainputstream-read-less-than-buffer-size
we should avoid using read since it not always read fully bytes.

## How was this patch tested?

existed test


lee-lei:
Thanks for catching this. This is really important to be a product. Thanks.

PR #409, opened 2017-11-21
Remove range tests from bitmap index suite

lee-lei:
Remove the range query tests, since bitmap only supports equal query.
## How was this patch tested?
unit tests

### DIFF
--- a/src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapIndexSuite.scala
+++ b/src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapIndexSuite.scala
@@ -108,32 +108,20 @@ class BitMapIndexSuite extends QueryTest with SharedSQLContext with BeforeAndAft
     sql("drop oindex index_bf on oap_test")
   }
 
-  test("BitMap index for range predicate which over multi partitions") {
-    val data: Seq[(Int, String)] = (1 to 200).map { i => (i, s"this is test $i") }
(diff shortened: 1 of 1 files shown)


xuanyuanking:
The implementation of before version can support range filter, if this hard to implement in current code base, we can leave some comments in tests.
Currently in Baidu internal env we are implementing the statistics module to quantitate each user cases, so we may have detail data to support this feature need or not.
@adrian-wang What do you think？

adrian-wang:
@xuanyuanking We discussed this in internal weekly meeting yesterday, the implementation is not an issue, but we want to make user to easily figure out what type of index they are looking for. If this is found important in real-world work loads, we can talk about it and bring it back. 
\cc @chenghao-intel

lee-lei:
Actually the existing bitmap index implementation supports range filters. From my understanding, the plan is to make bitmap index only support equal query for column with not many unique keys. Therefore I am removing the redundant range query support. Thanks.

xuanyuanking:
> If this is found important in real-world work loads, we can talk about it and bring it back.
Yep, that's what I'm thinking, after getting detail user statistics later and decide this, just leave some comments now.

lee-lei:
Got it. Thanks for sharing your thoughts.

chenghao-intel:
thank you @xuanyuanking , we can talk about the range index case in next offline meeting with you. :)

lee-lei:
Close this PR now. If we need to remove the range tests in the future, then reopen it. Thanks.

PR #410, opened 2017-11-21
[Minor] eliminate IDEA warnings in indexPlans.scala

adrian-wang:

eliminate some IDEA code insight warnings.
By add () for `new FindDataSourceTable(sparkSession)`,
IDEA will know that `relation` is a `LogicalPlan` instead of `Logging`.
Also in a `map()` function we should not return with `Unit`.

## How was this patch tested?

existed tests


linhong-intel:
lgtm

Issue #406, opened 2017-11-20
BitMap index throws out of bound exception by self-created unit test.


yueawang:
Below test case throws out-of-bound exception.

`

    test("out-of-bound bitmap index") {
    val data = (1 to 300).map{ i => (i, s"this is test $i")}
    val dataRDD = spark.sparkContext.parallelize(data, 10)
    dataRDD.toDF("key", "value").createOrReplaceTempView("t")
    sql("insert overwrite table oap_test select * from t")
    sql("create oindex index1 on oap_test (a) using bitmap")
    val data1 = (1 to 300).map{ i => (i % 10, s"this is test $i")}
    val dataRDD1 = spark.sparkContext.parallelize(data1, 5)
    dataRDD1.toDF("key", "value").createOrReplaceTempView("t1")
    sql("insert overwrite table oap_test_1 select * from t1")
    sql("create oindex index1 on oap_test_1 (a) using bitmap")
    checkAnswer(
      sql("SELECT * " +
        "FROM oap_test t1 " +
        "WHERE EXISTS " +
        "(SELECT 1 FROM oap_test_1 t2 " +
        "WHERE t1.a = t2.a AND t2.a >= 1 AND t1.a < 5) " +
        "ORDER BY a"),
      Seq(
        Row(1, "this is test 1"),
        Row(2, "this is test 2"),
        Row(3, "this is test 3"),
        Row(4, "this is test 4")))
    sql("drop oindex index1 on oap_test")
    sql("drop oindex index1 on oap_test_1")
    }

`

Testing started at 7:43 PM ...
03:43:54.877 WARN org.apache.hadoop.util.NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
03:43:58.980 ERROR org.apache.spark.executor.Executor: Exception in task 1.0 in stage 4.0 (TID 20)
java.lang.IndexOutOfBoundsException
	at java.nio.Buffer.checkIndex(Buffer.java:546)
	at java.nio.HeapByteBuffer.getInt(HeapByteBuffer.java:365)
	at org.roaringbitmap.buffer.ImmutableRoaringArray.<init>(ImmutableRoaringArray.java:44)
	at org.roaringbitmap.buffer.ImmutableRoaringBitmap.<init>(ImmutableRoaringBitmap.java:969)
[...]

PR #411, opened 2017-11-21
[DO NOT MERGE] Add one bitmap index test case.

yueawang:

Add one bitmap index test case.

## How was this patch tested?

unit test

### DIFF
--- a/src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapIndexSuite.scala
+++ b/src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapIndexSuite.scala
@@ -39,10 +39,6 @@ class BitMapIndexSuite extends QueryTest with SharedSQLContext with BeforeAndAft
     sql(s"""CREATE TEMPORARY VIEW oap_test (a INT, b STRING)
             | USING oap
             | OPTIONS (path '$path')""".stripMargin)
-    val path1 = Utils.createTempDir().getAbsolutePath
-    sql(s"""CREATE TEMPORARY VIEW oap_test_1 (a INT, b STRING)
-           | USING oap
-           | OPTIONS (path '$path1')""".stripMargin)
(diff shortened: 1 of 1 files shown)


src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapIndexSuite.scala
```diff
@@ -39,9 +39,10 @@ class BitMapIndexSuite extends QueryTest with SharedSQLContext with BeforeAndAft
     sql(s"""CREATE TEMPORARY VIEW oap_test (a INT, b STRING)
             | USING oap
             | OPTIONS (path '$path')""".stripMargin)
-    sql(s"""CREATE TEMPORARY VIEW oap_test_date (a INT, b DATE)
-            | USING oap
-            | OPTIONS (path '$path')""".stripMargin)
+    val path1 = Utils.createTempDir().getAbsolutePath
+    sql(s"""CREATE TEMPORARY VIEW oap_test_1 (a INT, b STRING)
+           | USING oap
+           | OPTIONS (path '$path1')""".stripMargin)
```
chenghao-intel: Seems we are keep using the `path` in the rest of the function, why do we have to use `path1` in this particular case?

yueawang: Moved to testcase which accesses it, also renamed to a meaningful name.

lee-lei:
Thanks Yue to submit the test case to catch the issue #406. I am analyzing the root cause. Thanks.

linhong-intel:
Just for lei to debug, so we will not merge this.

lee-lei:
IMHO, it's a good test case to merge in order to track either feature dev or performance optimization patch to achieve the expected functionality in the future. Thanks.

linhong-intel:
close it and @lee-lei will upload this case in his PR

PR #414, opened 2017-11-22
Code refactory of oapStrategies and index scanner.

yueawang:

1. Add index requirement check of oapStrategies.
2. Reorganize a few logic of index & file scan.
3. Re-enable SemiJoin oap strategy which applies to bitmap index only.

## How was this patch tested?


src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -324,7 +314,8 @@ trait OapStrategies extends Logging {
         _fsRelation
     }
 
-    if (fsRelation.fileFormat.asInstanceOf[OapFileFormat].hasAvailableIndex(normalizedIndexHint)) {
+    if (fsRelation.fileFormat.asInstanceOf[OapFileFormat]
+      .hasAvailableIndex(normalizedIndexHint, indexRequirements)) {
```
linhong-intel: Maybe you can combine this `if()` with above `match case` to remove `asInstanceOf`
Something like:
```
_fsRelation.fileFormat match {
  case fileFormat : OapFileFormat =
    fileFormat.initialize()
    val fsRelation = _fsRelation
    if (fileFormat.hasAvailableIndex) {
      ....
    } else {
      ....
    }
}
```

linhong-intel: BTW, below style of `else` is wrong:
Correct:
```
if (...) {
  A
} else {
  B
}
```
Wrong:
```
if () {
  A
} else B
```

adrian-wang:
Could you resolve the conflicts?

linhong-intel:
Change the description.

> How How was this patch tested?

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -255,15 +263,10 @@ trait OapStrategies extends Logging {
...
 
         createOapFileScanPlan(
-          projectList, indexHint, relation, file, table, oapOption, indexHint) match {
-          case Some(fastScan) =>
-            OapAggregationFileScanExec(
-              aggExpressions,
-              projectList,
-              fastScan)
+          projectList, indexHint, relation, file, table, oapOption, indexHint, Nil) match {
+          case Some(fastScan) => OapAggregationFileScanExec(aggExpressions, projectList, fastScan)
```
chenghao-intel: indent

src/main/scala/org/apache/spark/sql/execution/datasources/oap/DataSourceMeta.scala
```diff
@@ -125,13 +126,13 @@ private[oap] case class BTreeIndex(entries: Seq[BTreeIndexEntry] = Nil) extends
     BitSet.fromBitMask(Array(INDEX_METRICS_KEY_ORDER_BIT_MASK | INDEX_METRICS_KEY_GROUP_BIT_MASK))
 }
 
-private[oap] case class BitMapIndex(entries: Seq[Int] = Nil) extends IndexType {
```
chenghao-intel: Any reason why we remove the `private[oap]`, seems all of the reference places are still under the package `oap`.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFileFormat.scala
```diff
@@ -231,7 +231,9 @@ private[sql] class OapFileFormat extends FileFormat
           if (supportFilters.nonEmpty) {
             // determine whether we can use index
             supportFilters.foreach(filter => logDebug("\t" + filter.toString))
-            ScannerBuilder.build(supportFilters, ic)
+            // get index options such as limit, order, etc.
+            val indexOptions = options.filterKeys(_.contains("oap.scan"))
```
chenghao-intel: Can we put this as a function? `oap.scan` seems a little bit tricky, and we'd better to place the function with the other contansts together.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
```diff
@@ -44,22 +44,22 @@ private[oap] abstract class IndexScanner(idxMeta: IndexMeta)
...
+
   protected var keySchema: StructType = _
 
   /**
    * Scan N items from each index entry.
    */
   private var limitScan : Int = 0
 
-  def setScanNumLimit(scanNum : Int) : Unit = {
+  def setInternalLimit(scanNum : Int) : Unit = {
```
chenghao-intel: In scala, usually we don't use the getter or setter method, at least the name conversion is not setXXX. probably you can try:
```scala
@BeanProperty
private var internalLimit: Int = _
```

chenghao-intel: https://stackoverflow.com/questions/26012039/scala-getters-setters-best-practice

chenghao-intel: https://www.dustinmartin.net/getters-and-setters-in-scala/

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -106,13 +108,18 @@ trait OapStrategies extends Logging {
             (OapFileFormat.OAP_QUERY_ORDER_OPTION_KEY -> order.head.isAscending.toString))
           val indexHint = if (filters.nonEmpty) filters
             else IsNotNull(orderAttributes.head) :: Nil
+          val indexRequirement = indexHint.map(_ => BTreeIndex())
 
           createOapFileScanPlan(
```
chenghao-intel: Will that be more readiable if we create a Pattern object just like the `PhysicalOperation`? As I see `createOapFileScanPlan` always along with `PhysicalOperation` in the code.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/DataSourceMeta.scala
yueawang: There is one exception oapStrategy which belongs to another package checks the indexType to apply its optimization method.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
yueawang: Will follow that.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFileFormat.scala
yueawang: OK, a check func can be applied here.

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
yueawang: Yes, PhysicalOperation always comes along with createOapFileScanPlan, and which is exactly what FileSourceStrategy does, so far I do not merge them together, or say abstract a common object is because the condition check between strategies are somehow different, and just 3 strategies looks few, I can combine it in future with more strategies implementations.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/DataSourceMeta.scala
adrian-wang: we can put oapStrategy back into package of [OAP]

adrian-wang: another option is to use `private[sql]` here. I prefer move `OapStrategy`

src/main/scala/org/apache/spark/sql/execution/OapStrategies.scala
```diff
@@ -31,7 +31,7 @@ import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
 import org.apache.spark.sql.catalyst.util.CaseInsensitiveMap
 import org.apache.spark.sql.execution.aggregate.OapAggUtils
 import org.apache.spark.sql.execution.datasources._
-import org.apache.spark.sql.execution.datasources.oap.OapFileFormat
+import org.apache.spark.sql.execution.datasources.oap.{BitMapIndex, BTreeIndex, IndexType, OapFileFormat}
```
adrian-wang: `BTreeIndex` should be placed in front of `BitMapIndex`.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/DataSourceMeta.scala
yueawang: OK, the origin thought was put all strategies in one package, but moving oapStrategy looks acceptable, let's move.

adrian-wang:
LGTM

PR #416, opened 2017-11-22
ISSUE#387 re-enable statistics for B-tree

adrian-wang:

This is to re-enable statistics for BTree.

- [x] re-enable statistics r/w in BTree
- [x] re-enable statistics tests
- [x] select index based on statistics for BTree
- [x] clean old code concerning array[byte]
- [x] clean old code concerning `UnsafeRow`

## How was this patch tested?

Unit tests

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
@@ -22,6 +22,8 @@ import org.apache.hadoop.fs.Path
 
 import org.apache.spark.sql.execution.datasources.oap._
 import org.apache.spark.sql.execution.datasources.oap.filecache._
+import org.apache.spark.sql.execution.datasources.oap.index.BTreeIndexRecordReader.BTreeFooter
+import org.apache.spark.sql.execution.datasources.oap.statistics.StatisticsManager
 
 // we scan the index from the smallest to the largest,
 // this will scan the B+ Tree (index) leaf node.
@@ -44,6 +46,20 @@ private[oap] class BPlusTreeScanner(idxMeta: IndexMeta) extends IndexScanner(idx
     recordReader = BTreeIndexRecordReader(conf, keySchema)
     recordReader.initialize(path, intervalArray)
     this
+  }
+
+  override protected def readStatistics(indexPath: Path, conf: Configuration): Double = {
+    // TODO decouple with btreeindexrecordreader
+    // This is called before the scanner call `initialize`
+    val reader = BTreeIndexFileReader(conf, indexPath)
+    val footerFiber = BTreeFiber(() => reader.readFooter(), reader.file.toString, 0, 0)
+    val footerCache = FiberCacheManager.get(footerFiber, conf)
+    val footer = BTreeFooter(footerCache)
+    val offset = footer.getStatsOffset
+
+    val statisticsManager = new StatisticsManager
+    statisticsManager.read(footerCache, offset, keySchema)
+    statisticsManager.analyse(intervalArray, conf)
   }
 
   override def hasNext: Boolean = recordReader.hasNext
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
@@ -246,7 +246,7 @@ private[index] object BTreeIndexRecordReader {
       fiberCache.getInt(nodeMetaStart + idx * nodeMetaByteSize + nodePosOffset)
     def getNodeSize(idx: Int): Int =
       fiberCache.getInt(nodeMetaStart + idx * nodeMetaByteSize + nodeSizeOffset)
-    def getStatsOffset: Int = Integer.SIZE * 3 / 8 + nodeMetaByteSize * getNodesCount
+    def getStatsOffset: Int = Integer.SIZE / 8 * 3 + nodeMetaByteSize * getNodesCount
     private def getStatsLength: Int = fiberCache.getInt(
       nodeMetaStart + nodeMetaByteSize * getNodesCount)
   }
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
@@ -46,6 +46,8 @@ private[index] case class BTreeIndexRecordReader(
   private lazy val ordering = GenerateOrdering.create(schema)
   private lazy val partialOrdering = GenerateOrdering.create(StructType(schema.dropRight(1)))
 
+  def getFooterFiber: FiberCache = footerCache
+
   def initialize(path: Path, intervalArray: ArrayBuffer[RangeInterval]): Unit = {
     reader = BTreeIndexFileReader(configuration, path)
 
@@ -219,12 +221,14 @@ private[index] case class BTreeIndexRecordReader(
 private[index] object BTreeIndexRecordReader {
 
   private[index] case class BTreeFooter(fiberCache: FiberCache) {
+    // TODO move to companion object
     private val nodePosOffset = Integer.SIZE / 8
     private val nodeSizeOffset = Integer.SIZE / 8 * 2
     private val minPosOffset = Integer.SIZE / 8 * 3
     private val maxPosOffset = Integer.SIZE / 8 * 4
     private val nodeMetaStart = Integer.SIZE / 8 * 2
     private val nodeMetaByteSize = Integer.SIZE / 8 * 5
+    private val statsLengthSize = Integer.SIZE / 8
 
     def getRecordCount: Int = fiberCache.getInt(0)
     def getNodesCount: Int = fiberCache.getInt(Integer.SIZE / 8)
@@ -234,14 +238,17 @@ private[index] object BTreeIndexRecordReader {
       IndexUtils.readBasedOnSchema(fiberCache, getMinValueOffset(idx), schema)
     def getMinValueOffset(idx: Int): Int =
       fiberCache.getInt(nodeMetaStart + nodeMetaByteSize * idx + minPosOffset) +
-          nodeMetaStart + nodeMetaByteSize * getNodesCount
+          nodeMetaStart + nodeMetaByteSize * getNodesCount + statsLengthSize + getStatsLength
     def getMaxValueOffset(idx: Int): Int =
       fiberCache.getInt(nodeMetaStart + nodeMetaByteSize * idx + maxPosOffset) +
-          nodeMetaStart + nodeMetaByteSize * getNodesCount
+          nodeMetaStart + nodeMetaByteSize * getNodesCount + statsLengthSize + getStatsLength
     def getNodeOffset(idx: Int): Int =
       fiberCache.getInt(nodeMetaStart + idx * nodeMetaByteSize + nodePosOffset)
     def getNodeSize(idx: Int): Int =
       fiberCache.getInt(nodeMetaStart + idx * nodeMetaByteSize + nodeSizeOffset)
+    def getStatsOffset: Int = Integer.SIZE * 3 / 8 + nodeMetaByteSize * getNodesCount
+    private def getStatsLength: Int = fiberCache.getInt(
+      nodeMetaStart + nodeMetaByteSize * getNodesCount)
   }
 
   private[index] case class BTreeRowIdList(fiberCache: FiberCache) {
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
@@ -17,8 +17,6 @@
 
 package org.apache.spark.sql.execution.datasources.oap.index
 
-import org.apache.spark.sql.execution.datasources.oap.index.BTreeIndexRecordReader.BTreeFooter
-
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 
@@ -28,9 +26,8 @@ import org.apache.hadoop.fs.Path
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.expressions.UnsafeRow
 import org.apache.spark.sql.execution.datasources.oap._
-import org.apache.spark.sql.execution.datasources.oap.filecache.{BTreeFiber, FiberCacheManager}
 import org.apache.spark.sql.execution.datasources.oap.io.OapIndexInfo
-import org.apache.spark.sql.execution.datasources.oap.statistics.{StaticsAnalysisResult, StatisticsManager}
+import org.apache.spark.sql.execution.datasources.oap.statistics.StaticsAnalysisResult
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types.StructType
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
@@ -17,6 +17,8 @@
 
 package org.apache.spark.sql.execution.datasources.oap.index
 
+import org.apache.spark.sql.execution.datasources.oap.index.BTreeIndexRecordReader.BTreeFooter
+
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 
@@ -26,7 +28,9 @@ import org.apache.hadoop.fs.Path
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.expressions.UnsafeRow
 import org.apache.spark.sql.execution.datasources.oap._
+import org.apache.spark.sql.execution.datasources.oap.filecache.{BTreeFiber, FiberCacheManager}
 import org.apache.spark.sql.execution.datasources.oap.io.OapIndexInfo
+import org.apache.spark.sql.execution.datasources.oap.statistics.{StaticsAnalysisResult, StatisticsManager}
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types.StructType
@@ -112,12 +116,11 @@ private[oap] abstract class IndexScanner(idxMeta: IndexMeta)
       val dataFileSize = dataPath.getFileSystem(conf).getContentSummary(dataPath).getLength
       val ratio = conf.getDouble(SQLConf.OAP_INDEX_FILE_SIZE_MAX_RATIO.key,
         SQLConf.OAP_INDEX_FILE_SIZE_MAX_RATIO.defaultValue.get)
-      indexFileSize <= dataFileSize * ratio
+      if (indexFileSize > dataFileSize * ratio) return false
 
       // Policy 2: statistics tells the scan cost
-      // TODO: Get statistics info after statistics is enabled.
-      // val statsAnalyseResult = tryToReadStatistics(indexPath, conf)
-      // statsAnalyseResult == StaticsAnalysisResult.USE_INDEX
+      val statsAnalyseResult = tryToReadStatistics(indexPath, conf)
+      statsAnalyseResult == StaticsAnalysisResult.USE_INDEX
 
       // More Policies
     } else {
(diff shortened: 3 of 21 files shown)


adrian-wang:
BitMap version will be updated after bit map index really use unsafe way to access data.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
```diff
@@ -128,30 +129,30 @@ class StatisticsManager {
...
-  def read(bytes: Array[Byte], s: StructType): Unit = {
-    var offset = 0L
-    val mask = Platform.getLong(bytes, Platform.BYTE_ARRAY_OFFSET + offset)
-    offset += 8
+  def read(fiberCache: FiberCache, offset: Int, s: StructType): Unit = {
+    var readOffset = 0
+    val mask = fiberCache.getLong(offset + readOffset)
+    readOffset += 8
     if (mask != StatisticsManager.STATISTICSMASK) {
       invalidStatistics = true
```
chenghao-intel: Why not using the `Option[Array[Statistics]]` instead?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
```diff
@@ -128,30 +129,30 @@ class StatisticsManager {
...
     } else {
-      val numOfStats = Platform.getInt(bytes, Platform.BYTE_ARRAY_OFFSET + offset)
-      offset += 4
+      val numOfStats = fiberCache.getInt(offset + readOffset)
+      readOffset += 4
       stats = new Array[Statistics](numOfStats)
 
       for (i <- 0 until numOfStats) {
-        Platform.getInt(bytes, Platform.BYTE_ARRAY_OFFSET + offset) match {
+        fiberCache.getInt(offset + readOffset) match {
```
chenghao-intel:
```scala
object Statistics {
...
  val TYPE_MIN_MAX: Int = 0
  val TYPE_SAMPLE_BASE: Int = 1
  val TYPE_PART_BY_VALUE: Int = 2
  val TYPE_BLOOM_FILTER: Int = 3

  def unapply(t: Int): Option[Statistics] = t match {
    case TYPE_MIN_MAX => Some(new MinMaxStatistics)
    case TYPE_SAMPLE_BASE => Some(new SampleBasedStatistics)
    case TYPE_PART_BY_VALUE => Some(new PartByValueStatistics)
    case TYPE_BLOOM_FILTER => Some(new BloomFilterStatistics)
    case _ => None
  }
...
}


...
      for (i <- 0 until numOfStats) {
        fiberCache.getInt(offset + readOffset) match {
          case Statistics(s) => stats(i) = s
          case _ => throw new UnsupportedOperationException("unsupport statistics id")
        }
        readOffset += 4
      }
```
Probably we don't need `SampleBasedStatisticsType.id` etc. any more and we definitely can simplify the code.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
```diff
@@ -46,6 +48,20 @@ private[oap] class BPlusTreeScanner(idxMeta: IndexMeta) extends IndexScanner(idx
     this
   }
 
+  override protected def readStatistics(indexPath: Path, conf: Configuration): Double = {
+    // TODO decouple with btreeindexrecordreader
+    // This is called before the scanner call `initialize`
+    val reader = BTreeIndexFileReader(conf, indexPath)
+    val footerFiber = BTreeFiber(() => reader.readFooter(), reader.file.toString, 0, 0)
```
chenghao-intel:
```scala
val footerFiber = BTreeFiber(() => BTreeIndexFileReader(conf, indexPath).readFooter(), indexPath.toString, 0, 0)
```

Once we initialize the `BTreeIndexFileReader` instance, and it will open the index file, but this is costy, and not necessary if the footerFiber has been cached. Let's make the fiber read as lazy.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
```diff
@@ -193,7 +194,9 @@ object StatisticsManager {
 
   val statisticsTypeMap: scala.collection.mutable.Map[AnyIndexType, Array[StatisticsType]] =
     scala.collection.mutable.Map(
-      BTreeIndexType -> Array.empty,
+      BTreeIndexType -> Array(
+        MinMaxStatisticsType, SampleBasedStatisticsType,
+        BloomFilterStatisticsType, PartByValueStatisticsType),
```
chenghao-intel: I know this exists long time ago, but code like:

```scala
object StatisticsManager {
  val STATISTICSMASK: Long = 0x20170524abcdefabL // a random mask for statistics begin

  val statisticsTypeMap: scala.collection.mutable.Map[AnyIndexType, Array[StatisticsType]] =
    scala.collection.mutable.Map(
      BTreeIndexType -> Array(
        MinMaxStatisticsType, SampleBasedStatisticsType,
        BloomFilterStatisticsType, PartByValueStatisticsType),
      BitMapIndexType -> Array.empty)

  /**
   * Using a static object to store parameter is not a good idea, some reasons:
   * 1. In local mode, driver and worker will use one same object, but in cluster they are different
   * 2. object.setXXX() then val a = object.XXX can be realized by Configuration.set and get
   * But
   * 1. current Statistics has no interface to set parameters by Configuration and many test suites
   * depend on this static object.
   * 2. the `initialize()` in each Statistics is depended by both `read` and `write` while `read`
   * should get parameter from disk and doesn't need Configuration.
   * 3. Some test suites only init specific Statistics and not init StatisticsManager, so remove
   * this object will need interface change in every Statistics sub classes, this means more changes
   * So, TODO: Let's address all of this latter.
   */
  var sampleRate: Double = SQLConf.OAP_STATISTICS_SAMPLE_RATE.defaultValue.get
  var partNumber: Int = SQLConf.OAP_STATISTICS_PART_NUM.defaultValue.get
  var bloomFilterMaxBits: Int = SQLConf.OAP_BLOOMFILTER_MAXBITS.defaultValue.get
  var bloomFilterHashFuncs: Int = SQLConf.OAP_BLOOMFILTER_NUMHASHFUNC.defaultValue.get

  var FULLSCANTHRESHOLD: Double = SQLConf.OAP_FULL_SCAN_THRESHOLD.defaultValue.get

  // TODO we need to find better ways to configure these parameters
  def setStatisticsType(indexType: AnyIndexType, statisticsType: Array[StatisticsType]): Unit =
    statisticsTypeMap.update(indexType, statisticsType)
  def setSampleRate(rate: Double): Unit = this.sampleRate = rate
  def setPartNumber(num: Int): Unit = this.partNumber = num
  def setFullScanThreshold(rate: Double): Unit = this.FULLSCANTHRESHOLD = rate
  def setBloomFilterMaxBits(maxBits: Int): Unit = this.bloomFilterMaxBits = maxBits
  def setBloomFilterHashFuncs(num: Int): Unit = this.bloomFilterHashFuncs = num
}
```
I don't think we can pass the value via an object, which will cause many performance in multiple thread mode. Please think about to use configuration instances for each query.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/PartByValueStatistics.scala
```diff
@@ -43,18 +44,17 @@ import org.apache.spark.unsafe.Platform
...
-  @transient private lazy val converter = UnsafeProjection.create(schema)
 
   private lazy val maxPartNum: Int = StatisticsManager.partNumber
   @transient private lazy val ordering = GenerateOrdering.create(schema)
   @transient private lazy val partialOrdering =
     GenerateOrdering.create(StructType(schema.dropRight(1)))
 
-  protected case class PartedByValueMeta(idx: Int, row: InternalRow,
-                                         curMaxId: Int, accumulatorCnt: Int)
+  protected case class PartedByValueMeta(
```
chenghao-intel: Please implement each statistic in a separated PR.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
```diff
@@ -46,6 +48,20 @@ private[oap] class BPlusTreeScanner(idxMeta: IndexMeta) extends IndexScanner(idx
     this
   }
 
+  override protected def readStatistics(indexPath: Path, conf: Configuration): Double = {
```
chenghao-intel: I will agree we'd better to put the statistic into a separated class, not binding with `BPlusTreeScanner`, basically there are different things.

Another thing, why not just place the `footerCache` as parameters? I believe when the code logic run into here, the index(or index footer) should be loaded already, and the statistic should just an optional optimization to enable/disable the index usage before we query on it. Please think about how to refactor the code and simplify the logic to avoid the duplication code.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
chenghao-intel: typo, I mean the conflict in multi-thread model.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
yueawang: how about change name to analyzeStatistics?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
```diff
@@ -126,6 +129,24 @@ private[oap] abstract class IndexScanner(idxMeta: IndexMeta)
     }
   }
 
+  /**
+   * Through getting statistics from related index file,
+   * judging if we should bypass this datafile or full scan or by index.
+   * return -1 means bypass, close to 1 means full scan and close to 0 means by index.
+   * called before invoking [[initialize]].
+   */
+  private def tryToReadStatistics(indexPath: Path, conf: Configuration): Double = {
```
yueawang: tryAnalyzeStatistics seems better as read does not tell the full functionality.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
```diff
@@ -46,6 +48,20 @@ private[oap] class BPlusTreeScanner(idxMeta: IndexMeta) extends IndexScanner(idx
...
+    // TODO decouple with btreeindexrecordreader
+    // This is called before the scanner call `initialize`
+    val reader = BTreeIndexFileReader(conf, indexPath)
+    val footerFiber = BTreeFiber(() => reader.readFooter(), reader.file.toString, 0, 0)
+    val footerCache = FiberCacheManager.get(footerFiber, conf)
+    val footer = BTreeFooter(footerCache)
+    val offset = footer.getStatsOffset
+
+    val statisticsManager = new StatisticsManager
```
yueawang: Should we reduce this kind of new?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
```diff
@@ -234,14 +238,17 @@ private[index] object BTreeIndexRecordReader {
...
+          nodeMetaStart + nodeMetaByteSize * getNodesCount + statsLengthSize + getStatsLength
     def getMaxValueOffset(idx: Int): Int =
       fiberCache.getInt(nodeMetaStart + nodeMetaByteSize * idx + maxPosOffset) +
-          nodeMetaStart + nodeMetaByteSize * getNodesCount
+          nodeMetaStart + nodeMetaByteSize * getNodesCount + statsLengthSize + getStatsLength
     def getNodeOffset(idx: Int): Int =
       fiberCache.getInt(nodeMetaStart + idx * nodeMetaByteSize + nodePosOffset)
     def getNodeSize(idx: Int): Int =
       fiberCache.getInt(nodeMetaStart + idx * nodeMetaByteSize + nodeSizeOffset)
+    def getStatsOffset: Int = Integer.SIZE * 3 / 8 + nodeMetaByteSize * getNodesCount
```
yueawang: These & above's magic number are confusing, why Integer.SIZE *3 / 8? Any connection to minPosOffset = Integer.SIZE / 8 * 3 above?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/BloomFilterStatistics.scala
```diff
@@ -81,27 +81,28 @@ private[oap] class BloomFilterStatistics extends Statistics {
...
-    var offset = baseOffset
-    val bitLength = Platform.getInt(bytes, Platform.BYTE_ARRAY_OFFSET + offset)
-    val numHashFunc = Platform.getInt(bytes, Platform.BYTE_ARRAY_OFFSET + offset + 4)
-    offset += 8
+  private def readBloomFilter(fiberCache: FiberCache, offset: Int): Int = {
+    var readOffset = offset
+
+    val bitLength = fiberCache.getInt(readOffset)
+    val numHashFunc = fiberCache.getInt(readOffset + 4)
+    readOffset += 8
```
yueawang: I think it's better to use code like type.defaultsize, any historical reason?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
adrian-wang: Actually, the index hasn't been loaded... This function is the first place that tries to load index footer.

adrian-wang: As I has observed that, I add the TODO below: "decouple with btreeindexrecordreader".

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
adrian-wang: I think there will not be duplicate loading...By using `FiberCacheManager.get`, we are reading from the underlying cache.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/PartByValueStatistics.scala
adrian-wang: Since the code related to each statistic is already in the repo, I only add API to read from cached data for each statistic.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
adrian-wang: Sure, will fix that

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
adrian-wang: Sure, thanks

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/BloomFilterStatistics.scala
adrian-wang: will improve code readability accordingly, thanks.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
linhong-intel: Since statistic is storing in index file, this should be part of index scan rather than out of index.
So `readStatistics()` should put in `BTreeIndexRecordReader.initialize()`
Something like:
```
def initialize() {
  footerCache = FiberCacheManager.get(footerFiber, configuration)
  ...
  statisticsManager.read(footerCache, offset, keySchema)
  statisticsManager.analyse(intervalArray, conf)
  if (SKIP_INDEX) Some(Iterator.Empty)
  else if (FULL_SCAN) None
  else read_and_scan_index_data()
}
```
Since we have to read index file anyway. It's better to do the reading in one place. then StatisticsManager can be treated as a Util class.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
linhong-intel: Integer.SIZE / 8 = IntegerType.defaultSize = 4.
Integer.SIZE * 3 / 8 means skip 3 INTs. I think Integer.SIZE / 8 * 3 is better

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
linhong-intel: Another reason, since write statistics is part of index writing already. read statistics in index reading is reasonable.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
linhong-intel: @adrian-wang , I proposed a commit (https://github.com/Intel-bigdata/OAP/pull/435/commits/bb88d5b5acf7d913be17a8539367496fc405bc7b) try to fix this kind of object, parameter, config problem. 
you can check if it's ok.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
```diff
@@ -17,6 +17,8 @@
 
 package org.apache.spark.sql.execution.datasources.oap.index
 
+import org.apache.spark.sql.execution.datasources.oap.index.BTreeIndexRecordReader.BTreeFooter
```
linhong-intel: import order is not correct

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/OapDataReaderWriter.scala
```diff
@@ -234,20 +234,4 @@ private[oap] class OapDataReader(
         iter
     }
   }
-
-  /**
-   * Through getting statistics from related index file,
-   * judging if we should bypass this datafile or full scan or by index.
-   * return -1 means bypass, close to 1 means full scan and close to 0 means by index.
-   */
-  private def tryToReadStatistics(indexPath: Path, conf: Configuration): Double = {
```
linhong-intel: after remove this, some import should be removed as well

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/MinMaxStatistics.scala
```diff
@@ -52,27 +55,34 @@ private[oap] class MinMaxStatistics extends Statistics {
     }
   }
 
-  override def write(writer: OutputStream, sortedKeys: ArrayBuffer[Key]): Long = {
+  override def write(writer: OutputStream, sortedKeys: ArrayBuffer[Key]): Int = {
     var offset = super.write(writer, sortedKeys)
     if (min != null) {
-      offset += Statistics.writeInternalRow(converter, min, writer)
```
linhong-intel: converter is unused

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
```diff
@@ -25,6 +25,7 @@ import org.apache.hadoop.conf.Configuration
 
 import org.apache.spark.sql.catalyst.expressions.codegen.GenerateOrdering
 import org.apache.spark.sql.execution.datasources.oap.Key
+import org.apache.spark.sql.execution.datasources.oap.filecache.FiberCache
 import org.apache.spark.sql.execution.datasources.oap.index._
 import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types._
```
linhong-intel: import org.apache.spark.unsafe.Platform is unused

linhong-intel: import org.apache.spark.unsafe.Platform is unused

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BPlusTreeScanner.scala
adrian-wang: Sounds reasonable, I'll refactor code relating to index reading, thanks.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/StatisticsManager.scala
adrian-wang: There has been an `OapEnv` in `hive-thriftserver`, I think the naming would be confusing. Better name it like `OapSession`, in accordance with `SparkSession`.

adrian-wang: Then you can decide if we need `MemoryManager` and `FiberCacheManager` to be singleton.

adrian-wang:
There are several remaining issues in the pr, I'll list and fix them in a new conversation. Will merge this first since this pr is the fundamental one and involves a lot of code.

PR #419, opened 2017-11-23
Change back to Guava Cache as Backend

linhong-intel:
## What changes were proposed in this pull request?

This is the first PR of Cache Refactor, contains:
* Change back to Guava Cache as Backend
* Refactor MemoryManager to acquire off-heap memory from Spark
* Generic data access function in FiberCache

Issues need to be resolved in following PR:
* [Important] Keep in-used FiberCache from being evicted
* Change FiberCacheManager, MemoryManger to class for better init
* Decode/Decompress DataFile in off-heap directly
* cache.status may be costly and not considering data eviction

## How was this patch tested?

Unit test passed, also passed new added test suite.

PR #420, opened 2017-11-24
Enable Guava as cache backend

linhong-intel:

This is a follow up of #419 
Enabled Guava
Created a series of get functions in FiberCache for generic accessing

## How was this patch tested?

Unit Test passed


linhong-intel:
merge this into #419

PR #432, opened 2017-11-29
move writeBasedOnSchema to IndexUtils

adrian-wang:

This is to move writeBasedOnSchema to IndexUtils.
For statistic manager, we need to write some key down,
and let's reuse the logic.

## How was this patch tested?

unit tests.


Moonlit-Sailor:
LGTM

PR #435, opened 2017-11-30
[DO NOT MERGE] Draft patch for Cache lock mechanism

linhong-intel:

1. Add a OapEnv for Oap initialize
2. Use Java's read write lock, to secure in-used cache not be freed
    2.1 hold a read lock before reading FiberCache (done in FiberCacheManager.get)
    2.2 Try to hold write lock in onRemove notify. if can't acquire, put cache back to Cache Manager
    2.3 create a dummy MemoryBlock without allocating memory when put into CacheManager
        -- 2.3.1 if put into cache success, hold read lock, allocate memory and load data
        -- 2.3.2 if put into cache fail, do nothing and return null to indicate no enough memory

## How was this patch tested?

Unit test pass

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -46,7 +46,7 @@ case class FiberInputStream(is: InputStream, offset: Long, length: Int)
  *
  * TODO: change object to class for better initialization
  */
-class FiberCacheManager(memorySize: Long) extends Logging {
+class FiberCacheManager(memorySize: Long, concurrency: Int) extends Logging {
 
   private val removalListener = new RemovalListener[Fiber, FiberCache] {
     override def onRemoval(notification: RemovalNotification[Fiber, FiberCache]): Unit = {
@@ -86,7 +86,7 @@ class FiberCacheManager(memorySize: Long) extends Logging {
 
   private val cache: Cache[Fiber, FiberCache] = CacheBuilder.newBuilder()
       .recordStats()
-      .concurrencyLevel(4)
+      .concurrencyLevel(concurrency)
       .removalListener(removalListener)
       .maximumWeight(MAX_WEIGHT)
       .weigher(weigher)
@@ -128,6 +128,11 @@ class FiberCacheManager(memorySize: Long) extends Logging {
   }
 
   def getStats: CacheStats = cache.stats()
+}
+
+object FiberCacheManager {
+  val OAP_FIBER_CACHE_CONCURRENCY = "spark.oap.fiberCache.concurrency"
+  val OAP_FIBER_CACHE_CONCURRENCY_DEFAULT = 4
 }
 
 private[oap] object DataFileHandleCacheManager extends Logging {
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -17,6 +17,7 @@
 
 package org.apache.spark.sql.execution.datasources.oap.filecache
 
+import java.io.InputStream
 import java.util.concurrent.{Callable, TimeUnit}
 
 import scala.collection.JavaConverters._
@@ -28,6 +29,7 @@ import org.apache.spark.internal.Logging
 import org.apache.spark.sql.execution.datasources.oap.OapEnv
 import org.apache.spark.sql.execution.datasources.oap.io._
 import org.apache.spark.sql.execution.datasources.oap.utils.CacheStatusSerDe
+import org.apache.spark.unsafe.memory.MemoryBlock
 import org.apache.spark.util.collection.BitSet
 
 // TODO need to register within the SparkContext
@@ -36,6 +38,8 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
     OapEnv.get.fiberCacheManager.status
   }
 }
+
+case class FiberInputStream(is: InputStream, offset: Long, length: Int)
 
 /**
  * Fiber Cache Manager
@@ -49,7 +53,13 @@ class FiberCacheManager(memorySize: Long) extends Logging {
       // TODO: Change the log more readable
       logDebug(s"Removing Cache ${notification.getKey}")
       // TODO: Investigate lock mechanism to secure in-used FiberCache
-      notification.getValue.dispose()
+      val fiberCache = notification.getValue
+      if (fiberCache.lock.writeLock().tryLock()) {
+        notification.getValue.dispose()
+        fiberCache.lock.writeLock().unlock()
+      } else {
+        cache.put(notification.getKey, notification.getValue)
+      }
     }
   }
   private val weigher = new Weigher[Fiber, FiberCache] {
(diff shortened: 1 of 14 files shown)


linhong-intel:
@chenghao-intel 
Can you review the second commit in this PR? https://github.com/Intel-bigdata/OAP/pull/435/commits/08a7033fd07caf837a291c2103b5f90b4cad32c7
Let's first check if this implementation is acceptable or not.

linhong-intel:
## Current implementation in master code

no-lock-master-code

## proposed implementation in commit https://github.com/Intel-bigdata/OAP/commit/08a7033fd07caf837a291c2103b5f90b4cad32c7
has-lock-pr-code

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -59,31 +68,41 @@ object FiberCacheManager extends Logging {
...
-    cache.get(fiber, cacheLoader(fiber, conf))
+  def get(fiber: Fiber, is: => FiberInputStream): FiberCache = {
+    val fiberCache = cache.get(fiber, cacheLoader(fiber, is.length))
+    fiberCache.lock.readLock().lock()
+    if (fiberCache.isDisposed) {
+      fiberCache.lock.readLock().unlock()
+      null
+    } else {
+      // we have inserted FiberCache into Cache Manager, so we've acquire memory. load data
+      fiberCache.loadData(is)
```
chenghao-intel: is it possible `onRemove` called right before we enter this function for the same Fiber?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -17,32 +17,53 @@
...
+import org.apache.spark.sql.execution.datasources.oap.OapEnv
 import org.apache.spark.storage.{BlockManager, TestBlockId}
 import org.apache.spark.unsafe.Platform
 import org.apache.spark.unsafe.memory.{MemoryAllocator, MemoryBlock}
 import org.apache.spark.unsafe.types.UTF8String
 
 // TODO: make it an alias of MemoryBlock
 trait FiberCache {
+  def loadData(is: FiberInputStream): Unit = {
+    val bytes = new Array[Byte](is.length)
```
chenghao-intel: probably we also need to leave some buffers for this kind of usage.

1) Request the maximum memory OAP(GUAVA_CACHE_SIZE + PARALLEM_TMP_BUFFER) used from MemoryManager of Spark
2) Create an Guava cache with weight GUAVA_CACHE_SIZE.

And PARALLEM_TMP_BUFFER can be parallel core * MAX(or AVG) Fiber size.

What do you think?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -17,32 +17,53 @@
...
 import org.apache.spark.sql.execution.datasources.oap.ColumnValues
+import org.apache.spark.sql.execution.datasources.oap.OapEnv
 import org.apache.spark.storage.{BlockManager, TestBlockId}
 import org.apache.spark.unsafe.Platform
 import org.apache.spark.unsafe.memory.{MemoryAllocator, MemoryBlock}
 import org.apache.spark.unsafe.types.UTF8String
 
 // TODO: make it an alias of MemoryBlock
 trait FiberCache {
+  def loadData(is: FiberInputStream): Unit = {
```
chenghao-intel: what if 2 separated thread request the same Fiber loading? Thread-safety.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -17,32 +17,53 @@
...
+    val bytes = new Array[Byte](is.length)
+    is.is match {
+      case fsInputStream: FSDataInputStream => fsInputStream.readFully(is.offset, bytes)
```
chenghao-intel: Actually I think we can leave the `loadData` to its sub classes. 
like:
```scala
def loadCache(): Unit = {
  lock the write lock
  try {
    if the data not loaded {
       load()
    }
  } finally {
    unlock the write lock
  }
}

protected[this] def load: Unit
```

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -17,32 +17,53 @@
...
   def isDisposed: Boolean = disposed
   def dispose(): Unit = {
-    if (!disposed) MemoryManager.free(fiberData)
+    if (!disposed && fiberData.getBaseOffset != -1) OapEnv.get.memoryManager.free(fiberData)
     disposed = true
   }
+  val lock = new ReentrantReadWriteLock()
 
   /** For debug purpose */
   def toArray: Array[Byte] = {
```
chenghao-intel: Add the `Deprecated` annotation.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
chenghao-intel: Lock the write lock? and double check if the data loaded when we got the write lock?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/OapDataFile.scala
```diff
@@ -104,11 +113,12 @@ private[oap] case class OapDataFile(path: String, schema: StructType,
     // We have to read Array[Byte] from file and decode/decompress it before putToFiberCache
     // TODO: Try to finish this in off-heap memory
     val data = fiberParser.parse(decompressor.decompress(bytes, uncompressedLen), rowCount)
-    MemoryManager.putToDataFiberCache(data)
+    FiberInputStream(new ByteArrayInputStream(data), 0, data.length)
```
chenghao-intel: Since we got the `byte[]` already, probably we should avoid to wrap it as the InputStream again.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -59,31 +68,41 @@ object FiberCacheManager extends Logging {
...
+      .concurrencyLevel(concurrency)
       .removalListener(removalListener)
       .maximumWeight(MAX_WEIGHT)
       .weigher(weigher)
       .build[Fiber, FiberCache]()
 
-  def get(fiber: Fiber, conf: Configuration): FiberCache = {
-    // Used a flag called disposed in FiberCache to indicate if this FiberCache is removed
-    cache.get(fiber, cacheLoader(fiber, conf))
+  def get(fiber: Fiber, is: => FiberInputStream): FiberCache = {
```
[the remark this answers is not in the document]
linhong-intel: Here is the implementation of cache.get(). https://github.com/google/guava/blob/v14.0.1/guava/src/com/google/common/cache/LocalCache.java#L2234
`waitForLoadingValue()` and `lockedGetOrLoad()` will make sure the cache won't load twice.

But code in `fiberCache.loadData()` should be made thread-safety. I'll fix it.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: two scenarios:
1. `onRemove` first then `fiberCache.loadData` right after that:
  `onRemove` will hold write lock. so `fiberCache.loadData` will wait until `onRemove` is done. then `fiberCache.isDisposed`, then just return null
2. `fiberCache.loadData` first then `onRemove` right after that:
  `loadData` will hold read lock. `onRemove` will failed on acquiring write lock. so just return and do nothing

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
linhong-intel: yes, there is a problem here. I think a `AtomicBoolean.compareAndSet()` can fix this

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
linhong-intel: See if my understanding is correct:
`PARALLEM_TMP_BUFFER` is used to store the data read from file and will be copied to cache.
if there is no enough space in cache, we can still use this temp buffer directly.

If so, I totally agree.
but need think about the allocate and free of temp buffer. Also, need to think about the proper buffer size.

linhong-intel:
close it.

PR #438, opened 2017-11-30
Support querying null/non-null key using B+ Tree index

Moonlit-Sailor:
Support query for isNull/isNotNull in B+ tree index:
`attribute is null` , `attribute is not null` now is supported by B+ tree index
## How was this patch tested?
unit test

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
@@ -27,6 +27,7 @@ import org.apache.spark.sql.catalyst.expressions.codegen.GenerateOrdering
 import org.apache.spark.sql.execution.datasources.oap.filecache.{BTreeFiber, FiberCache, FiberCacheManager}
 import org.apache.spark.sql.types._
 
+
 private[index] case class BTreeIndexRecordReader(
     configuration: Configuration,
     schema: StructType) extends Iterator[Int] {
@@ -72,7 +73,7 @@ private[index] case class BTreeIndexRecordReader(
     val (nodeIdxForEnd, isEndFound) = findNodeIdx(interval.end, isStart = false)
 
     if (nodeIdxForStart == nodeIdxForEnd && !isStartFound && !isEndFound) {
-      (0, 0)
+      (0, 0) // not found in B+ tree
     } else {
       val start = if (interval.start == IndexScanner.DUMMY_KEY_START) 0
       else {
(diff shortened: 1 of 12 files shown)


Moonlit-Sailor:
please help reviewing the `BTreeIndexRecordWriter` and `BTreeIndexRecordWriter` files,
since I worked on it based on the `ChunkedBuffer` version before you refactored to Guava cache version.
Then I merged with the newest version, I concern that if there are some underlying bugs
@linhong-intel 
thanks  : )

yueawang:
Just one question, Does intervalRange (DummyStart, DummyEnd) include those NULL value??

Moonlit-Sailor:
@yueawang      
> Does intervalRange (DummyStart, DummyEnd) include those NULL value?

nope, [DummyStart , DummyEnd] is all non-null keys

Moonlit-Sailor:
B+Tree Index File Structure
\************************************************
\*　　　　　　　 **Version   info**　　　　　　
\*———————————————————
\* 　　　　　　　**Node Data**　　　　
\*———————————————————
\* 　Row ID (with Non-Null key) List　　　　
\* 　Row ID (with Null key) List　　　　　<=======
\*———————————————————
\*　　　　　　　 **Footer**　　　
\* 　Row Count with Non-Null Key
\* 　Row Count With Null Key　　　　　<======
\* 　Node Count 
\*　 Nodes Meta Data
\*　 Min/Max Key Data for each node
\*———————————————————
\* 　Row Id List Section Size
\* 　Footer Section Size
\***********************************************

yueawang:
Thanks @Moonlit-Sailor , that's what I have expected.

linhong-intel:
LGTM but need some test cases.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
```diff
@@ -125,9 +130,10 @@ private[index] case class BTreeIndexRecordWriter(
         BTreeNodeMetaData(rowCount, nodeBuf.length, nodeUniqueKeys.head, nodeUniqueKeys.last)
       }
     // Write Row Id List
-    fileWriter.writeRowIdList(serializeRowIdLists(uniqueKeys))
+    fileWriter.writeRowIdList(serializeRowIdLists(nonNullUniqueKeys ++ nullKeys))
```
adrian-wang: combine 2 arrays will involve memory consumption, can we modify the interface to avoid that?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
```diff
@@ -139,11 +143,13 @@ private[index] case class BTreeIndexRecordReader(
    */
   private def findNodeIdx(candidate: InternalRow, isStart: Boolean): (Option[Int], Boolean) = {
     val idxOption = (0 until footer.getNodesCount).find { idx =>
-      rowOrdering(candidate, footer.getMaxValue(idx, schema), isStart) <= 0
+      footer.getRowCountOfNode(idx) > 0 && // ensure this node is not an empty node
```
linhong-intel: A node has zero row count? can you give some details here?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/UnsafeIndexTree.scala
```diff
@@ -210,15 +210,25 @@ private[oap] class CurrentKey(node: IndexNode, keyIdx: Int, valueIdx: Int, index
...
 }
 
-private [oap] class RangeInterval(s: Key, e: Key, includeStart: Boolean, includeEnd: Boolean)
-  extends Serializable {
+private [oap] class RangeInterval(
+    s: Key,
+    e: Key,
+    includeStart: Boolean,
+    includeEnd: Boolean,
+    isNull: Boolean) extends Serializable {
```
linhong-intel: In my understanding, isNull and s, e, includeStart, includeEnd is exclusive.
So it's better to have two kinds of RangeInterval. one is for range, one is for null.
Please think about if we can make another PR for this

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
Moonlit-Sailor: all the keys are null

adrian-wang:
LGTM pending tests.

PR #441, opened 2017-12-04
Avoid finding ending key index for bitmap equal query.

lee-lei:
    1. Use binary search to find starting or ending index.
    2. If equal query, no need to find ending index.
    3. Correct starting index or ending index for dummy starting
       key and dummy ending key.
    4. Add one unit test.
## How was this patch tested?
unit tests all passed.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapScanner.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapScanner.scala
@@ -172,34 +172,35 @@ private[oap] case class BitMapScanner(idxMeta: IndexMeta) extends IndexScanner(i
     startIdxOffset
   }
 
-  private def getEndIdxOffset(fiberCache: FiberCache, baseOffset: Long, endIdx: Int): Int = {
-    val idxOffset = baseOffset + endIdx * 4
-    val endIdxOffset = Platform.getInt(FiberCache, idxOffset)
-    endIdxOffset
-  }
-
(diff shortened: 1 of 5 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapScanner.scala
```diff
@@ -159,33 +160,32 @@ private[oap] case class BitMapScanner(idxMeta: IndexMeta) extends IndexScanner(i
...
     } else {
-      // find first key which >= start key, can't find return -1
+      // Below will use binary search, which is more efficient than traversing from the first key.
+      // If no found, return -1.
       if (range.startInclude) {
-        keySeq.indexWhere(ordering.compare(range.start, _) <= 0)
+        keySeq.search(range.start)(ordering).insertionPoint
       } else {
-        keySeq.indexWhere(ordering.compare(range.start, _) < 0)
+        keySeq.search(range.start)(ordering).insertionPoint + 1
```
linhong-intel: if no found, return -1?
-1 + 1 = 0?

linhong-intel: Seems the code is not correct and current test can't find this problem.
Can you add some test cases?

lee-lei:
The new version will be uploaded soon. Thanks.

lee-lei:
Done, Please help review when you are available. Thanks. @linhong-intel @adrian-wang @yueawang

lee-lei:
You are right. I will correct and upload it at once. Thanks.

PR #453, opened 2017-12-07
follow up #416 clean up

adrian-wang:

- [ ] configurations for statisticmanager should support multi-thread scenario
- [ ] read statistics should move to index reader
- [x] simplify the code of statistic manager
- [x] simplify `StatisticsType.id`
- [x] function rename from `readStatistics` to `analyzeStatistics`

## How was this patch tested?

unit test.


adrian-wang:
I'll merge this to unblock benchmark.

PR #458, opened 2017-12-10
Directly decode bitmap from FiberCache

lee-lei:
Creat BitmapDataInputStream to directly decode bitmap from
FiberCache which is either in offheap/onheap memory.

## How was this patch tested?
All of tests passed.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapScanner.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapScanner.scala
@@ -16,6 +16,9 @@
  */
 package org.apache.spark.sql.execution.datasources.oap.index
 
+import java.io.DataInput
+import java.io.EOFException
+import java.lang.IndexOutOfBoundsException
 import java.nio.ByteBuffer
 
 import scala.collection.mutable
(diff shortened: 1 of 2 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapScanner.scala
```diff
@@ -200,35 +203,30 @@ private[oap] case class BitMapScanner(idxMeta: IndexMeta) extends IndexScanner(i
...
+     val bmStream = new BitmapDataInputStream(byteCache)
+     bmStream.skipBytes(position)
     (startIdx until endIdx).map( idx => {
-      // Below is directly constructed from byte buffer rather than deserializing into java object.
-      val bmEntry = new ImmutableRoaringBitmap(rawBb)
-      curPosition += bmEntry.serializedSizeInBytes
-      rawBb.position(curPosition)
+      val bmEntry = new RoaringBitmap()
+      // Below is directly reading from byte array rather than deserializing into java object.
+      bmEntry.deserialize(bmStream)
```
adrian-wang: I think by doing this, the data from FiberCache will still be moved to heap space.

lee-lei: IMHO, actually, the bits (bitmap data) doesn't need to be copied to on heap any more. The BitmapDataInputStream will directly read the bits from FiberCache. But the bitmap object itself will be created on heap with the foundation of bits from FiberCache, since bitmap is java object. This is the same mechanism if we want to use BitSet. If we want to directly get row IDs, one approach is to just save row IDs directly rather than bitmap. The other approach is to first decode bitmap and save row IDs to cache and next time directly get row IDs from FiberCache. Thanks.

PR #466, opened 2017-12-14
Add null support to bitmap index.

yueawang:

Add null value support to bitmap index.
1. Add null value index read & write.
2. Add test cases for normal and corner cases.

## How was this patch tested?
mvn test passes.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapScanner.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapScanner.scala
@@ -207,7 +207,10 @@ private[oap] case class BitMapScanner(idxMeta: IndexMeta) extends IndexScanner(i
     // If invalid starting index, just return.
     if (startIdx == -1 || startIdx == keyLength) return (-1, -1)
     // If equal query, no need to find endIdx.
-    if (range.start == range.end) return (startIdx, startIdx)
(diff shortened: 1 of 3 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitMapScanner.scala
```diff
@@ -224,17 +242,22 @@ private[oap] case class BitMapScanner(idxMeta: IndexMeta) extends IndexScanner(i
...
+          if (startIdx == -1 || endIdx == -1) {
+            // range not fond in cur bitmap, return empty for performance consideration
+            Seq.empty[ImmutableRoaringBitmap]
+          } else {
+            val startIdxOffset = getStartIdxOffset(bmOffsetListCache, 0L, startIdx)
+            val curPosition = startIdxOffset - bmEntryListOffset
+            getDesiredBitmaps(bmEntryListCache.toArray, curPosition, startIdx, endIdx)
+          }
+        case range if range.isNullPredicate =>
+          getDesiredBitmaps(bmNullListCache.toArray, 0, 0, 0)
```
lee-lei: Can we avoid to use toArray in line 258? The latest master doesn't use toArrary in line 255. Thanks.

lee-lei:
Please help rebase on the latest master. Thanks.

adrian-wang:
LGTM, thanks!

PR #469, opened 2017-12-18
Correct the starting and ending index for range query in bitmap index.

lee-lei:

1. Binary search will get the exact index to match the start or end of
   the specific range.
2. If not, the range start may be less than the first key. Or the range
   end may be greater than the last key.
3. Otherwise, it's invalid range.

## How was this patch tested?
All of unit test passed.


chenghao-intel:
Is it a bug fixing or optimization? And please add the unit test if it's a bug.

lee-lei:
It's a bug. The existng bitmap range test case in BitMapIndexSuite can be used to verify it. Thanks.

lee-lei:
@adrian-wang @linhong-intel If no question, please help merge it to the master. Thanks.

PR #470, opened 2017-12-19
throw exception when indexing more than 2g rows

### LINKED ISSUE #448: Fail index building if count exceed limit
We only support row id < Int.MaxValue

xuyangit:

Throw OapException when indexing more than 2g rows. Currently just check the recordCount in index record writer and if it exceeds Int.MaxValue, we throw an OapException.

## How was this patch tested?

Tested on tpcds1000 dataset with more than 2g rows for only one partition. 

Fixes #448

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
@@ -29,6 +29,7 @@ import org.apache.hadoop.mapreduce.{RecordWriter, TaskAttemptContext}
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.expressions.codegen.GenerateOrdering
+import org.apache.spark.sql.execution.datasources.OapException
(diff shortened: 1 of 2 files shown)


linhong-intel:
is it possible to create a test case for this? I'm not sure if unit test have bear 2g rows (but it should).

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
```diff
@@ -52,6 +53,9 @@ private[index] case class BTreeIndexRecordWriter(
     val v = genericProjector(value).copy()
     multiHashMap.put(v, recordCount)
     statisticsManager.addOapKey(v)
+    if (recordCount == Int.MaxValue) {
```
chenghao-intel: As we discussed previously, it will OOM even before we hit the maximize value, like 5M records. Can we have a hard code check for this? @yueawang

xuyangit: @chenghao-intel ,currently i'm working on removing code which limits us from writing more than 2g rows. It can write successfully now but there is an issue with reading index. I'm trying to fix it now. 
So I think this check is appropriate but should be merged when the full test for 2g rows reading and writing is passed.

Issue #472, opened 2017-12-19
B+ Tree Row Id List Fiber size can be 8GB maximum


linhong-intel:
Should split it into several pieces. For example 2M each.
Then load needed fiber on demand.

linhong-intel:
fixed by #473

PR #473, opened 2017-12-19
Partial Load RowIdList Fiber

linhong-intel:

For 2G rows index file, RowIdList Fiber size is 8GB. This is too large.
With partial loading, each fiber contains 1M rows maximum.
So fiber size is 4MB maximum. This is better.

## How was this patch tested?

Unit test passed

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexFileReader.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexFileReader.scala
@@ -64,7 +64,12 @@ private[oap] case class BTreeIndexFileReader(
 
   def readRowIdList(partIdx: Int): FiberCache = {
     val partSize = rowIdListSizePerSection * Integer.SIZE / 8
-    MemoryManager.putToIndexFiberCache(reader, rowIdListIndex + partIdx * partSize, partSize)
+    val readLength = if (partIdx * partSize + partSize > rowIdListLength) {
+      rowIdListLength % rowIdListSizePerSection
+    } else {
+      partSize
+    }
+    MemoryManager.putToIndexFiberCache(reader, rowIdListIndex + partIdx * partSize, readLength)
   }
 
   @deprecated("no need to read the whole row id list", "v0.3")
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexFileReader.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexFileReader.scala
@@ -36,6 +36,8 @@ private[oap] case class BTreeIndexFileReader(
   val rowIdListSectionId: Int = 1
   val nodeSectionId: Int = 2
 
+  val rowIdListSizePerSection: Int = 1024 * 1024
+
   private val (reader, fileLength) = {
     val fs = file.getFileSystem(configuration)
     (fs.open(file), fs.getFileStatus(file).getLen)
@@ -60,6 +62,12 @@ private[oap] case class BTreeIndexFileReader(
   def readFooter(): FiberCache =
     MemoryManager.putToIndexFiberCache(reader, footerIndex, footerLength)
 
+  def readRowIdList(partIdx: Int): FiberCache = {
+    val partSize = rowIdListSizePerSection * Integer.SIZE / 8
+    MemoryManager.putToIndexFiberCache(reader, rowIdListIndex + partIdx * partSize, partSize)
+  }
+
+  @deprecated("no need to read the whole row id list", "v0.3")
   def readRowIdList(): FiberCache =
     MemoryManager.putToIndexFiberCache(reader, rowIdListIndex, rowIdListLength)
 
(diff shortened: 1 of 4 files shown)


linhong-intel:
fixes #472

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
```diff
@@ -58,14 +56,23 @@ private[index] case class BTreeIndexRecordReader(
...
+          val rowIdListFiber = BTreeFiber(
+            () => reader.readRowIdList(partIdx),
+            reader.file.toString,
+            reader.rowIdListSectionId, partIdx)
+
+          val rowIdListCache = FiberCacheManager.get(rowIdListFiber, configuration)
+          val rowIdList = BTreeRowIdList(rowIdListCache)
+          val iterator =
+            subPosList.toIterator.map(i => rowIdList.getRowId(i % reader.rowIdListSizePerSection))
+          CompletionIterator[Int, Iterator[Int]](iterator,
```
yueawang: Does release make cache useless, as it reloads (from file??) at next access time.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexFileReader.scala
```diff
@@ -60,6 +62,17 @@ private[oap] case class BTreeIndexFileReader(
   def readFooter(): FiberCache =
     MemoryManager.putToIndexFiberCache(reader, footerIndex, footerLength)
 
+  def readRowIdList(partIdx: Int): FiberCache = {
+    val partSize = rowIdListSizePerSection * Integer.SIZE / 8
```
adrian-wang: Use `IndexUtils.INT_SIZE`

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexFileReader.scala
```diff
@@ -60,6 +62,17 @@ private[oap] case class BTreeIndexFileReader(
...
+    val readLength = if (partIdx * partSize + partSize > rowIdListLength) {
+      rowIdListLength % rowIdListSizePerSection
+    } else {
+      partSize
+    }
+    MemoryManager.putToIndexFiberCache(reader, rowIdListIndex + partIdx * partSize, readLength)
```
adrian-wang: what if the total length of row id list is much smaller than `readLength`?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexFileReader.scala
```diff
@@ -36,6 +36,8 @@ private[oap] case class BTreeIndexFileReader(
   val rowIdListSectionId: Int = 1
   val nodeSectionId: Int = 2
 
+  val rowIdListSizePerSection: Int = 1024 * 1024
```
yueawang: Use config? By that, we can do few benchmarks to find a better value later.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
linhong-intel: Yes, it will be reloaded at next access time. But in my opinion, since we are using cache, we should release the cache when unused, and acquire it when need again. Most time there will be no IO read and this can give cache a chance to free this memory if needed.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexFileReader.scala
linhong-intel: `rowIdListLength % rowIdListSizePerSection` should be changed to `rowIdListLength % partSize`
If `rowIdListLength` is much smaller, say, only 1 row, 4 Bytes, then `readLength = 4 Bytes % 8M Bytes = 4 Bytes`

linhong-intel:
> I think we should also revise the index writer to write down a row id list section table in the index footer. This can help make those sections more even.

@adrian-wang I'm not sure what does it mean.

In this PR, no change needed in write side, since each entry in rowIdList fiber is fixed-size (Int size). So no extra extra offset for each sub-part is needed.
Not clear about "make sections more even"

linhong-intel:
@yueawang @adrian-wang 
PR updated

adrian-wang:
@linhong-intel If not changing the writer side, the sub lists can be [1M, 1], while if we change that, the lists will be [500k, 500k], which is more even.

linhong-intel:
@adrian-wang I see.
This means we need to write a `partitionSize` of rowIdList Fiber into footer. Right?
Currently, I plan to use a configuration to control partitionSize. This is simple.
Can we implement your idea in another PR?

adrian-wang:
@linhong-intel  sure, just a thought , not sure if that can be beneficial for performance.

PR #474, opened 2017-12-20
refactor statisticmanager configuration

adrian-wang:

some follow-up in #453 .
configurations for statisticmanager should support multi-thread scenario.

## How was this patch tested?

unit tests


chenghao-intel:
@adrian-wang I have some high level comments on this:

1) Add detailed description on how to support the multiple thread mode.
2) Put the unrelated code change into separated PR(s);
3) Why there are so many code delete/add as most likely they are exactly the same? Please minimize the code change and will be great helpful for peer review.

Thanks,
Hao

adrian-wang:
I have updated the code.

adrian-wang:
basically I divided the old statisticsManager into StatisticsReadManager and StatisticsWriteManager, since we only need the configs in writing. So maybe you will see some code are exactly the same, yes.

Issue #468, opened 2017-12-18
Invalidate index caches if the index is dropped


linhong-intel:
For now, we use a timestamp in index name to WA this.
But a formal fix is still needed to secure v0.3 release

linhong-intel:
fixed by #480

PR #480, opened 2017-12-26
 Invalidate Cache After Index is Dropped

linhong-intel:

Index Data will be cached during using. After index is dropped, the related
cache should be removed.
Or the outdated cache may be used again and cause unexpected problem.

## How was this patch tested?

Unit test pass


linhong-intel:
Note 1:
Used java reflection to get access for private member

Note 2:
How to test:
Start a cluster mode spark with OAP loaded.
enable debug log and start a spark-shell and run:
```
val path = "/Users/linhongl/temp/data"
spark.sql(s"""CREATE TEMPORARY VIEW oap_test (a INT, b STRING) using oap options (path '$path')""".stripMargin)
val data: Seq[(Int, String)] = (1 to 300).map { i => (i, s"this is test $i") }
data.toDF("key", "value").createOrReplaceTempView("t")
spark.sql("insert overwrite table oap_test select * from t")
spark.sql("select * from oap_test").show()
spark.sql("create oindex index1 on oap_test (a)")
spark.sql("select * from oap_test where a = 1").show()
spark.sql("drop oindex index1 on oap_test")
```
Check log and looking for:
```
FiberCacheManager: going to remove cache of index1, executor: 0
FiberCacheManager: cache size before remove: 13
FiberCacheManager: cache size after remove: 10 <-- after remove, cache size should decrease
```
After this, restart another spark-shell and run again, WITHOUT`spark.sql("select * from oap_test where a = 1").show()`
Check the log again, cache size should be same before and after remove.

linhong-intel:
Close #468

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,6 +46,18 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
...
 
+  def removeIndexCache(indexName: String): Unit = {
+    logDebug(s"going to remove cache of $indexName, executor: ${SparkEnv.get.executorId}")
+    logDebug("cache size before remove: " + cache.size())
+    val fiberToBeRemoved = cache.asMap().keySet().asScala.filter {
+      case BTreeFiber(_, file, _, _) => file.contains(indexName)
+      case BitmapFiber(_, file, _, _) => file.contains(indexName)
+      case _ => false
+    }.asJava
+    cache.invalidateAll(fiberToBeRemoved)
```
chenghao-intel: Will the file handler be released also?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -213,6 +219,25 @@ case class DropIndex(
...
+        executorDataMapField.setAccessible(true)
+        val executorDataMap =
+          executorDataMapField.get(scheduler).asInstanceOf[mutable.HashMap[String, AnyRef]]
+        for ((_, executorData) <- executorDataMap) {
+          val c = Utils.classForName("org.apache.spark.scheduler.cluster.ExecutorData")
+          val executorEndpointField = c.getDeclaredField("executorEndpoint")
+          executorEndpointField.setAccessible(true)
+          val executorEndpoint =
+            executorEndpointField.get(executorData).asInstanceOf[RpcEndpointRef]
+          executorEndpoint.send(CacheDrop(indexName))
```
chenghao-intel: Can we make the above code as a function? I am wondering if we will post more events in the future.

src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedClusterMessage.scala
```diff
@@ -0,0 +1,116 @@
...
+  // Internal messages in driver
+  case object ReviveOffers extends CoarseGrainedClusterMessage
+
+  case object StopDriver extends CoarseGrainedClusterMessage
+
+  case object StopExecutor extends CoarseGrainedClusterMessage
+
+  case object StopExecutors extends CoarseGrainedClusterMessage
+
+  case class CacheDrop(indexName: String) extends CoarseGrainedClusterMessage
```
chenghao-intel: I thought we just need to add this into a separated file. can't we?

src/main/scala/org/apache/spark/executor/CoarseGrainedExecutorBackend.scala
```diff
@@ -0,0 +1,311 @@
...
+import org.apache.spark.deploy.worker.WorkerWatcher
+import org.apache.spark.internal.Logging
+import org.apache.spark.rpc._
+import org.apache.spark.scheduler.{ExecutorLossReason, TaskDescription}
+import org.apache.spark.scheduler.cluster.CoarseGrainedClusterMessages._
+import org.apache.spark.serializer.SerializerInstance
+import org.apache.spark.sql.execution.datasources.oap.filecache.FiberCacheManager
+import org.apache.spark.util.{ThreadUtils, Utils}
+
+private[spark] class CoarseGrainedExecutorBackend(
```
chenghao-intel: Can we simply take out the code of  `class CoarseGrainedExecutorBackend` and keep the else?

linhong-intel: Actually we can't, this file has only 2 classes.
`class CoarseGrainedExecutorBackend ` and
`object CoarseGrainedExecutorBackend `
if we only keep `class` and remove `object`, the code in `object` will be lost even if we don't overwrite it. The reason can be found in:
https://stackoverflow.com/questions/41570148/why-does-scala-place-a-dollar-sign-at-the-end-of-class-names

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: Yes it's released, if you mean the "Footer" Fiber.
But we don't cache the `FSDataInputStream` for index file as Data file.

linhong-intel:
@chenghao-intel 
updated.
Remove one spark file.
Move send/handle message function into a Util class.

src/main/scala/org/apache/spark/executor/CoarseGrainedExecutorBackend.scala
```diff
@@ -0,0 +1,311 @@
...
+      ).onFailure { case e =>
+        logWarning(s"Unable to notify the driver due to " + e.getMessage, e)
+      }(ThreadUtils.sameThread)
+    }
+
+    System.exit(code)
+  }
+}
+
+private[spark] object CoarseGrainedExecutorBackend extends Logging {
```
chenghao-intel: Is there any code change compared to the original one? Do we have to copy the source code?

linhong-intel:
```
     case Shutdown =>
       stopping.set(true)
       new Thread("CoarseGrainedExecutorBackend-stop-executor") {
         override def run(): Unit = {
           // executor.stop() will call `SparkEnv.stop()` which waits until RpcEnv stops totally.
           // However, if `executor.stop()` runs in some thread of RpcEnv, RpcEnv won't be able to
           // stop until `executor.stop()` returns, which becomes a dead-lock (See SPARK-14180).
           // Therefore, we put this line in a new thread.
            executor.stop()
          }
        }.start()
 +
 +    case message: OapMessage =>
 +      OapMessageUtils.handleOapMessage(message)
```

chenghao-intel: I mean for object `CoarseGrainedExecutorBackend`, not the class `CoarseGrainedExecutorBackend`

chenghao-intel: and I would also encourage you to update the Spark upstream, to support the extensible message handling.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -111,9 +72,26 @@ object FiberCacheManager extends Logging {
...
-  def cacheStats: CacheStats = cache.stats()
-
-  def cacheSize : Long = _cacheSize.get()
+  def cacheStats: CacheStats = cache.cacheStats
+
+  def cacheSize : Long = cache.cacheSize
+
+  private val cache: OapCache = {
+    val sparkEnv = SparkEnv.get
+    if (sparkEnv == null) {
```
chenghao-intel: is it possible be a `null` or can we simple put an assert here?

src/main/scala/org/apache/spark/executor/CoarseGrainedExecutorBackend.scala
linhong-intel: Actually we can't remove the `object CoarseGrainedExecutorBackend `
if we only keep class and remove object, the code in object will be lost even if we don't overwrite it. The reason can be found in:
https://stackoverflow.com/questions/41570148/why-does-scala-place-a-dollar-sign-at-the-end-of-class-names

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -0,0 +1,147 @@
...
+
+  override def get(fiber: Fiber, conf: Configuration): FiberCache = {
+    fiber.fiber2Data(conf)
+  }
+
+  override def getDataFibers: mutable.Set[DataFiber] = {
+    mutable.Set.empty
+  }
+
+  override def getIndexFibers(indexName: String): mutable.Set[Fiber] = {
```
chenghao-intel: The return value has to be mutable?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -0,0 +1,147 @@
...
+  def cacheStats: CacheStats
+}
+
+class SimpleOapCache extends OapCache with Logging {
+
+  override def get(fiber: Fiber, conf: Configuration): FiberCache = {
+    fiber.fiber2Data(conf)
+  }
+
+  override def getDataFibers: mutable.Set[DataFiber] = {
```
chenghao-intel: same here.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: Sorry, this code is not belong to this PR.
But it's a good point, I'll change it.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
linhong-intel: Sorry, this code is not belong to this PR.
The guava cache gives us a mutable.Set. If we want to return an immutable, we have to convert it manually, just not sure if this is necessary.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -0,0 +1,147 @@
...
+    cache.asMap().keySet().asScala.collect {
+      case fiber: DataFiber => fiber
+    }
+  }
+
+  override def getIndexFibers(indexName: String): mutable.Set[Fiber] = {
+    cache.asMap().keySet().asScala.filter {
```
chenghao-intel: asScala will cause duplicated the set, maybe simply using the for - loop.

Another issue I am not sure will it cause problem in multiple threads? I mean a new Fiber added in another thread. A lock is necessary? Maybe we can identify the issue in #494

src/main/scala/org/apache/spark/executor/CoarseGrainedExecutorBackend.scala
chenghao-intel: OK, got it.

chenghao-intel:
Thank you @linhong-intel , I will suggest to separated the unrelated code to another PR (maybe #494). The other looks good to me.

chenghao-intel:
and #482 :)

PR #481, opened 2017-12-26
check Index File version when reading index file

### LINKED ISSUE #444: Check index version and ignore not supported ones
Currently, we don't check index version and will read index data even if not support. Then exception happened.

Moonlit-Sailor:
Check index file Version before using index.

## How was this patch tested?
unit tests

Fixes #444


Moonlit-Sailor:
alright...
modify it later

adrian-wang:
I think when we write the version to footer, maybe you don't have to write PREFIX.

adrian-wang:
LGTM

Issue #486, opened 2017-12-28
Move Bitmap Single Column Assertion from Executor to Driver.


songzhan01:

songzhan01:
There are some advantages:
1. fast fail
2. avoid waste of resources

PR #487, opened 2017-12-28
The assert(BitMap index supports the column with one single field) in…#486

songzhan01:
… Driver is better than in Executor.

Add  assert in indexPlans.scala
Delete assert in BitmapIndexRecordWriter.scala

## How was this patch tested?
existing tests all passed.

add a new test in BitMapIndexSuite.scala 
test("BitMap index supports the column with one single field")


xuanyuanking:
Just close this, give the PR based on clean branch.

PR #494, opened 2018-01-02
Create CacheGuardian To Handle Fiber Removal

linhong-intel:

Use a `refCount` to track fiber usage number
Only add fiber to a pending queue in onRemoval
Create a CacheGuardian to handle pending removal fibers
CacheGuardian will Wait fiber release if refCount > 0
Changed some code to ensure Fiber is release after using

## How was this patch tested?

Unit test pass, add a new test case

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -96,6 +96,7 @@ object FiberCacheManager extends Logging {
       // TODO: Change the log more readable
       logDebug(s"Add Cache ${notification.getKey} into removal list")
       removalPendingQueue.offer(notification.getValue)
+      // TODO: Reduce cache size until CacheGuardian removed it?
       _cacheSize.addAndGet(-notification.getValue.size())
     }
   }
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -56,25 +56,15 @@ object FiberCacheManager extends Logging {
       logDebug(s"Removing Cache ${notification.getKey}")
       // TODO: Investigate lock mechanism to secure in-used FiberCache
       val fiberCache = notification.getValue
-      if (fiberCache.refCount.get() > 0) {
-        fiberCache.lock.lock()
-        try {
-          // lock will be released in await and acquired after thread wake up.
-          if (fiberCache.nonUsed.await(3000, TimeUnit.MILLISECONDS)) {
-            fiberCache.dispose()
-            _cacheSize.addAndGet(-notification.getValue.size())
-          } else {
-            logWarning(s"Remove Cache ${notification.getKey} failed")
-            exceptionFlag = true
-          }
-        } finally {
-          fiberCache.lock.unlock()
-        }
+      if (fiberCache.tryDispose(3000, TimeUnit.MILLISECONDS)) {
+        _cacheSize.addAndGet(-notification.getValue.size())
       } else {
-        fiberCache.dispose()
+        logWarning(s"Remove Cache ${notification.getKey} failed")
+        exceptionFlag = true
       }
     }
   }
+
   private val weigher = new Weigher[Fiber, FiberCache] {
     override def weigh(key: Fiber, value: FiberCache): Int =
       math.ceil(value.size() / MB).toInt
@@ -116,7 +106,7 @@ object FiberCacheManager extends Logging {
       cache.asMap().values().asScala.foreach(_.dispose())
       throw new OapException("Can't remove in-used fiber within 3 seconds")
     }
-    fiberCache.refCount.incrementAndGet()
+    fiberCache.occupy()
     fiberCache
   }
 
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -128,15 +128,15 @@ object FiberCacheManager extends Logging {
   private val cache = CacheBuilder
     .newBuilder()
     .recordStats()
-    .concurrencyLevel(4)
+    .concurrencyLevel(4) // TODO: Make 4 configurable
     .removalListener(removalListener)
     .maximumWeight(MAX_WEIGHT)
     .weigher(weigher)
     .build[Fiber, FiberCache]()
 
   def get(fiber: Fiber, conf: Configuration): FiberCache = {
-    // TODO: Should avoid loading a fiber larger than MAX_WEIGHT
     val fiberCache = cache.get(fiber, cacheLoader(fiber, conf))
+    // Avoid loading a fiber larger than MAX_WEIGHT / 4, 4 is concurrency number
     assert(fiberCache.size() <= MAX_WEIGHT * MB / 4, "Can't cache fiber larger than MAX_WEIGHT / 4")
     fiberCache.occupy()
     fiberCache
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -21,8 +21,10 @@ import java.util.concurrent.{Callable, TimeUnit}
 import java.util.concurrent.atomic.AtomicLong
 
 import scala.collection.JavaConverters._
+
 import com.google.common.cache._
 import org.apache.hadoop.conf.Configuration
+
 import org.apache.spark.SparkConf
 import org.apache.spark.executor.custom.CustomManager
 import org.apache.spark.internal.Logging
@@ -45,6 +47,9 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
  */
 object FiberCacheManager extends Logging {
 
+  // Set this to true if we need to throw an exception in removalListener
+  private[filecache] var exceptionFlag = false
+
   private val removalListener = new RemovalListener[Fiber, FiberCache] {
     override def onRemoval(notification: RemovalNotification[Fiber, FiberCache]): Unit = {
       // TODO: Change the log more readable
@@ -52,12 +57,21 @@ object FiberCacheManager extends Logging {
       // TODO: Investigate lock mechanism to secure in-used FiberCache
       val fiberCache = notification.getValue
       if (fiberCache.refCount.get() > 0) {
-        if (fiberCache.nonUsed.await(3000, TimeUnit.MILLISECONDS)) {
-          fiberCache.dispose()
-          _cacheSize.addAndGet(-notification.getValue.size())
-        } else {
-          throw new OapException("not release in 3 second")
+        fiberCache.lock.lock()
+        try {
+          // lock will be released in await and acquired after thread wake up.
+          if (fiberCache.nonUsed.await(3000, TimeUnit.MILLISECONDS)) {
+            fiberCache.dispose()
+            _cacheSize.addAndGet(-notification.getValue.size())
+          } else {
+            logWarning(s"Remove Cache ${notification.getKey} failed")
+            exceptionFlag = true
+          }
+        } finally {
+          fiberCache.lock.unlock()
         }
+      } else {
+        fiberCache.dispose()
       }
     }
   }
@@ -94,9 +108,10 @@ object FiberCacheManager extends Logging {
       .weigher(weigher)
       .build[Fiber, FiberCache]()
 
-  def get(fiber: Fiber, conf: Configuration): FiberCache = {
+  def get(fiber: Fiber, conf: Configuration): FiberCache = synchronized {
     // Used a flag called disposed in FiberCache to indicate if this FiberCache is removed
     val fiberCache = cache.get(fiber, cacheLoader(fiber, conf))
+    if (exceptionFlag) throw new OapException("Can't remove in-used fiber within 3 seconds")
     fiberCache.refCount.incrementAndGet()
     fiberCache
   }
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -56,7 +56,7 @@ object FiberCacheManager extends Logging {
         val fiberCache = queue.take()
         logDebug(s"Removing fiber $fiberCache ...")
         // Block if fiber is in use.
-        while (!fiberCache.dispose(3000, TimeUnit.MILLISECONDS)) {
+        while (!fiberCache.tryDispose(3000, TimeUnit.MILLISECONDS)) {
           // Check memory usage every 3s while we are waiting fiber release.
           if (queue.asScala.map(_.size()).sum > maxMemory) {
             logWarning("Fibers pending on removal use too much memory")
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -104,7 +104,7 @@ object FiberCacheManager extends Logging {
   }
 
   // Used by test suite
-  private[filecache] def removeFiber(fiber: TestFiber): Unit = {
+  private[filecache] def removeFiber(fiber: TestFiber): Unit = synchronized {
     // cache may be removed by other thread before invalidate
     // but it's ok since only used by test to simulate race condition
     if (cache.getIfPresent(fiber) != null) cache.invalidate(fiber)
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -17,7 +17,7 @@
 
 package org.apache.spark.sql.execution.datasources.oap.filecache
 
-import java.util.concurrent.{Callable, TimeUnit}
+import java.util.concurrent.{Callable, PriorityBlockingQueue, TimeUnit}
 import java.util.concurrent.atomic.AtomicLong
 
 import scala.collection.JavaConverters._
@@ -28,7 +28,6 @@ import org.apache.hadoop.conf.Configuration
 import org.apache.spark.SparkConf
 import org.apache.spark.executor.custom.CustomManager
 import org.apache.spark.internal.Logging
-import org.apache.spark.sql.execution.datasources.OapException
 import org.apache.spark.sql.execution.datasources.oap.io._
 import org.apache.spark.sql.execution.datasources.oap.utils.CacheStatusSerDe
 import org.apache.spark.util.collection.BitSet
@@ -47,21 +46,41 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
  */
 object FiberCacheManager extends Logging {
 
-  // Set this to true if we need to throw an exception in removalListener
-  private[filecache] var exceptionFlag = false
+  private val DEFAULT_QUEUE_INITIAL_CAPACITY = 11
+  private val removalPendingQueue =
+    new PriorityBlockingQueue[FiberCache](DEFAULT_QUEUE_INITIAL_CAPACITY, Ordering.by(_.refCount))
+
+  private class CacheGuardian(queue: PriorityBlockingQueue[FiberCache]) extends Thread {
+    override def run(): Unit = {
+      logWarning("started ...")
+      while (true) {
+        // Block until queue is not empty
+        val fiberCache = queue.take()
+        logWarning("get one entry")
+        // Block until fiberCache is disposed
+        while (!fiberCache.dispose(3000, TimeUnit.MILLISECONDS)) {
+          // logDebug("Can't remove in 3s, will try again")
+          logWarning("Can't remove in 3s, will try again")
+          val usedMemory = queue.asScala.map(_.size()).sum
+          if (usedMemory > 200 * MB) {
+            logWarning("Removal Pending Fiber used too much memory")
+          }
+        }
+        // logDebug("Fiber removed successfully")
+        logWarning("Fiber removed successfully")
+      }
+    }
+  }
+
+  private val cacheGuardian = new CacheGuardian(removalPendingQueue)
+  cacheGuardian.start()
 
   private val removalListener = new RemovalListener[Fiber, FiberCache] {
     override def onRemoval(notification: RemovalNotification[Fiber, FiberCache]): Unit = {
       // TODO: Change the log more readable
       logDebug(s"Removing Cache ${notification.getKey}")
-      // TODO: Investigate lock mechanism to secure in-used FiberCache
-      val fiberCache = notification.getValue
-      if (fiberCache.tryDispose(3000, TimeUnit.MILLISECONDS)) {
-        _cacheSize.addAndGet(-notification.getValue.size())
-      } else {
-        logWarning(s"Remove Cache ${notification.getKey} failed")
-        exceptionFlag = true
-      }
+      removalPendingQueue.offer(notification.getValue)
+      _cacheSize.addAndGet(-notification.getValue.size())
     }
   }
 
@@ -90,22 +109,17 @@ object FiberCacheManager extends Logging {
       }
     }
 
-  private val cache = CacheBuilder.newBuilder()
-      .recordStats()
-      .concurrencyLevel(4)
-      .removalListener(removalListener)
-      .maximumWeight(MAX_WEIGHT)
-      .weigher(weigher)
-      .build[Fiber, FiberCache]()
+  private val cache = CacheBuilder
+    .newBuilder()
+    .recordStats()
+    .concurrencyLevel(4)
+    .removalListener(removalListener)
+    .maximumWeight(MAX_WEIGHT)
+    .weigher(weigher)
+    .build[Fiber, FiberCache]()
 
-  def get(fiber: Fiber, conf: Configuration): FiberCache = synchronized {
-    // Used a flag called disposed in FiberCache to indicate if this FiberCache is removed
+  def get(fiber: Fiber, conf: Configuration): FiberCache = {
     val fiberCache = cache.get(fiber, cacheLoader(fiber, conf))
-    if (exceptionFlag) {
-      // Here we don't check the refCount since we will throw exception
-      cache.asMap().values().asScala.foreach(_.dispose())
-      throw new OapException("Can't remove in-used fiber within 3 seconds")
-    }
     fiberCache.occupy()
     fiberCache
   }
@@ -130,7 +144,7 @@ object FiberCacheManager extends Logging {
 
   def cacheStats: CacheStats = cache.stats()
 
-  def cacheSize : Long = _cacheSize.get()
+  def cacheSize: Long = _cacheSize.get()
 }
 
 private[oap] object DataFileHandleCacheManager extends Logging {
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -111,7 +111,11 @@ object FiberCacheManager extends Logging {
   def get(fiber: Fiber, conf: Configuration): FiberCache = synchronized {
     // Used a flag called disposed in FiberCache to indicate if this FiberCache is removed
     val fiberCache = cache.get(fiber, cacheLoader(fiber, conf))
-    if (exceptionFlag) throw new OapException("Can't remove in-used fiber within 3 seconds")
+    if (exceptionFlag) {
+      // Here we don't check the refCount since we will throw exception
+      cache.asMap().values().asScala.foreach(_.dispose())
+      throw new OapException("Can't remove in-used fiber within 3 seconds")
+    }
     fiberCache.refCount.incrementAndGet()
     fiberCache
   }
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -21,13 +21,12 @@ import java.util.concurrent.{Callable, TimeUnit}
 import java.util.concurrent.atomic.AtomicLong
 
 import scala.collection.JavaConverters._
-
 import com.google.common.cache._
 import org.apache.hadoop.conf.Configuration
-
 import org.apache.spark.SparkConf
 import org.apache.spark.executor.custom.CustomManager
 import org.apache.spark.internal.Logging
+import org.apache.spark.sql.execution.datasources.OapException
 import org.apache.spark.sql.execution.datasources.oap.io._
 import org.apache.spark.sql.execution.datasources.oap.utils.CacheStatusSerDe
 import org.apache.spark.util.collection.BitSet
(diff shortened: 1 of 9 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,13 +47,32 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
...
+      val fiberCache = notification.getValue
+      if (fiberCache.refCount.get() > 0) {
+        fiberCache.lock.lock()
+        try {
+          // lock will be released in await and acquired after thread wake up.
+          if (fiberCache.nonUsed.await(3000, TimeUnit.MILLISECONDS)) {
+            fiberCache.dispose()
+            _cacheSize.addAndGet(-notification.getValue.size())
+          } else {
+            logWarning(s"Remove Cache ${notification.getKey} failed")
```
LuciferYang: in this else branch, fiberCache not dispose , _cacheSize not decrement , but the notification already remove from the cache, Is there a memory leak ?

linhong-intel: Thanks, good finding. But the memory leak is not only on this fiber, we need cleanup all the caches before exception. Or the cached memory will never be released.

linhong-intel: updated, the fix is in FiberCacheManager.get()

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -88,9 +108,16 @@ object FiberCacheManager extends Logging {
...
 
-  def get(fiber: Fiber, conf: Configuration): FiberCache = {
+  def get(fiber: Fiber, conf: Configuration): FiberCache = synchronized {
     // Used a flag called disposed in FiberCache to indicate if this FiberCache is removed
-    cache.get(fiber, cacheLoader(fiber, conf))
+    val fiberCache = cache.get(fiber, cacheLoader(fiber, conf))
+    if (exceptionFlag) {
+      // Here we don't check the refCount since we will throw exception
+      cache.asMap().values().asScala.foreach(_.dispose())
+      throw new OapException("Can't remove in-used fiber within 3 seconds")
```
chenghao-intel: Shall we set the `exceptionFlag` back to false?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -88,9 +108,16 @@ object FiberCacheManager extends Logging {
...
+    }
+    fiberCache.refCount.incrementAndGet()
```
chenghao-intel: Or can we use the function instead of call the refCount directly? `fiberCache.open()` or `fiberCache.occupy()` etc.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -88,9 +108,16 @@ object FiberCacheManager extends Logging {
...
       .build[Fiber, FiberCache]()
 
-  def get(fiber: Fiber, conf: Configuration): FiberCache = {
+  def get(fiber: Fiber, conf: Configuration): FiberCache = synchronized {
     // Used a flag called disposed in FiberCache to indicate if this FiberCache is removed
-    cache.get(fiber, cacheLoader(fiber, conf))
+    val fiberCache = cache.get(fiber, cacheLoader(fiber, conf))
+    if (exceptionFlag) {
+      // Here we don't check the refCount since we will throw exception
+      cache.asMap().values().asScala.foreach(_.dispose())
```
chenghao-intel: as we will throw exception, which will fails the current task, but doens't mean we have to release all of the cache isn't? even if we release all of the cache, don't forget to reset the `_cacheSize`.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,13 +47,32 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
...
       // TODO: Investigate lock mechanism to secure in-used FiberCache
-      notification.getValue.dispose()
-      _cacheSize.addAndGet(-notification.getValue.size())
+      val fiberCache = notification.getValue
+      if (fiberCache.refCount.get() > 0) {
+        fiberCache.lock.lock()
+        try {
+          // lock will be released in await and acquired after thread wake up.
+          if (fiberCache.nonUsed.await(3000, TimeUnit.MILLISECONDS)) {
+            fiberCache.dispose()
```
chenghao-intel: This is really a danger operation, as will cause the null pointer exception (segment fault) in offheap memory access, and executor will crash without any hint.

linhong-intel:
@chenghao-intel @carsonwang @yueawang 
Let's review this PR. Thanks.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -92,17 +126,21 @@ object FiberCacheManager extends Logging {
...
+  private val cache = CacheBuilder
+    .newBuilder()
+    .recordStats()
+    .concurrencyLevel(4) // TODO: Make 4 configurable
+    .removalListener(removalListener)
+    .maximumWeight(MAX_WEIGHT)
+    .weigher(weigher)
+    .build[Fiber, FiberCache]()
 
   def get(fiber: Fiber, conf: Configuration): FiberCache = {
```
yueawang: Does this func need synchronized?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,6 +46,39 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
...
+      queue: PriorityBlockingQueue[FiberCache],
+      maxMemory: Long) extends Thread {
+    override def run(): Unit = {
+      // Loop forever
+      while (true) {
+        // Block if there is no pending fibers
+        val fiberCache = queue.take()
+        logDebug(s"Removing fiber $fiberCache ...")
+        // Block if fiber is in use.
+        while (!fiberCache.dispose(3000, TimeUnit.MILLISECONDS)) {
```
yueawang: Maybe name like disposeWithTimer or disposeWait sound better??

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,6 +46,39 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
...
+          // Check memory usage every 3s while we are waiting fiber release.
+          if (queue.asScala.map(_.size()).sum > maxMemory) {
```
yueawang: I thought the memory size in CacheGuardian just works as a temp buffer to prevent easy overflow of cache memory. Correct me if wrong.
Does the total size of those buffers in queue really matter here?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: sure

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: This log is useful for debugging.
We just don't want to use memory more than we requested from spark.
If this happens, usually something is wrong, like fiber not release for a very long time, or off-heap memory is too small.
The worst case is, we use too much memory and yarn will kill the executor.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: Guava cache is thread safe, there is a lock inside.
If we use synchronized here, we will lose the concurrency.
synchronized is same to change `concurrencyLevel(4)` to `concurrencyLevel(1)`

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,6 +46,39 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
...
+        logDebug(s"Fiber $fiberCache removed successfully")
+      }
+    }
+  }
+
+  // Default parameter comes from PriorityBlockQueue constructor
+  private val DEFAULT_QUEUE_INITIAL_CAPACITY = 11
+  // Ordering.by() makes sure non-used FiberCache dequeue first, used by test suite.
+  private[filecache] val removalPendingQueue = new PriorityBlockingQueue[FiberCache](
+    DEFAULT_QUEUE_INITIAL_CAPACITY, Ordering.by(_.refCount))
```
carsonwang: When you release a FiberCache, its refCount is updated. But its position in the priority queue will not be updated. Next time you take an element from the queue you are not going to get the FiberCache with lowest refCount.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,6 +46,39 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
...
+      queue: PriorityBlockingQueue[FiberCache],
+      maxMemory: Long) extends Thread {
+    override def run(): Unit = {
+      // Loop forever
+      while (true) {
+        // Block if there is no pending fibers
+        val fiberCache = queue.take()
+        logDebug(s"Removing fiber $fiberCache ...")
+        // Block if fiber is in use.
+        while (!fiberCache.tryDispose(3000, TimeUnit.MILLISECONDS)) {
```
carsonwang: It seems we will block at here forever if this fiberCache can not be disposed? It is possible that other fiberCaches be added to the queue whose refCount is 0. So can we go through all the elements in the queue and only dispose the fiberCache whose refCount is 0?

linhong-intel: Yes, this is what we talked at Monday. We have 2 options:
1. check pending fibers one by one and wait until refCount is 0
2. check pending fibers one by one and skip fibers with refCount > 0

We have chosen option#1, but option#2 is also acceptable.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: I checked the implementation of `PriorityBlockingQueue`
sort happens both in `offer()` and `take()`. So, `take()` will make sure the the poped entry is the lowest at current queue.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
carsonwang: The problem of #1 is that we may wait for a long time or forever, and the thread does nothing. At the same time, more fibers (refCount can be 0) are going to be added to the removal queue and we have more chance to be OOM.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -33,9 +36,53 @@ import org.apache.spark.unsafe.types.UTF8String
 
 // TODO: make it an alias of MemoryBlock
 trait FiberCache {
```
chenghao-intel:
```scala
trait FiberCache {
  private var _refCount = 0
  def synchronized occupy(): Unit = { _refCount += 1 }
  def synchronized release(): Unit = { 
    _refCount -= 1
    notifyAll()
  }

  def synchronized tryDispose(): Unit = {
    while(_refCount > 0) {
      wait(100) // 100 ms or even less
      log.warning(s"Fiber Cache Dispose waiting detected for ${this.toString()}")
    }
    realDispose()
  }

  private[this] def realDispose(): Unit = { release the memory }
}
```

Do you think if above code works?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,6 +46,39 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
  */
 object FiberCacheManager extends Logging {
 
+  private class CacheGuardian(
+      queue: PriorityBlockingQueue[FiberCache],
+      maxMemory: Long) extends Thread {
+    override def run(): Unit = {
+      // Loop forever
+      while (true) {
```
chenghao-intel: We probably need a way to release the CacheGuardian / FiberCacheManager just for ease of unit testing, but can leave this for the next PR or release.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
carsonwang: Really? Getting the element with highest priority should only be O(1) and removing it should be O(logN). Check this discussion  https://stackoverflow.com/questions/1871253/updating-java-priorityqueue-when-its-elements-change-priority

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
linhong-intel: of course it works.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -92,17 +126,21 @@ object FiberCacheManager extends Logging {
...
+    .weigher(weigher)
+    .build[Fiber, FiberCache]()
 
   def get(fiber: Fiber, conf: Configuration): FiberCache = {
-    // Used a flag called disposed in FiberCache to indicate if this FiberCache is removed
-    cache.get(fiber, cacheLoader(fiber, conf))
+    val fiberCache = cache.get(fiber, cacheLoader(fiber, conf))
+    // Avoid loading a fiber larger than MAX_WEIGHT / 4, 4 is concurrency number
+    assert(fiberCache.size() <= MAX_WEIGHT * MB / 4, "Can't cache fiber larger than MAX_WEIGHT / 4")
+    fiberCache.occupy()
```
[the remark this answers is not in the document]
linhong-intel: Yes, this is a problem.
How about:
```
def get(fiber: Fiber, conf: Configuration): FiberCache = synchronized {
}
```
Since `onRemoval` is called inside `get()`, the operation is sequential.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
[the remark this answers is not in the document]
linhong-intel: you mean `ConcurrentLinkedQueue` ? yes, we can.
But then the `CacheGuardian` should change from signal wake up to periodically wake up.

linhong-intel: Thanks @carsonwang , you are right. I see, the sort in remove is after pop.
Anyway, Let's change to a simple queue without priority and optimize this later as @chenghao-intel  said.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
linhong-intel: As I know, synchronized function is locked the whole object. I mean, `occupy`, `release`, `tryDispose` can't happen at the same time.
Then, if we enter `tryDispose() -> while(_refCount > 0)`, it will loop forever, since `release` cannot be called.

linhong-intel: In another way, if synchronized only lock it's own function, then `occupy` and `tryDispose` can happen at the same time, right after `while(_refCount > 0)` and before `realDispose()`. So we have to introduce another lock to make sure this won't happen.
Maybe it's `def get(fiber: Fiber, conf: Configuration): FiberCache = synchronized`

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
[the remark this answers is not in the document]
chenghao-intel: so probably we need to add the `synchronized` for the `get()` method?

I mean `onRemove` can be called by another thread as many thread can run the `get()` simultaneously.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
chenghao-intel: yes, that's by design. and once `wait(100)` called, the `synchronized` lock will be released. see: https://docs.oracle.com/javase/tutorial/essential/concurrency/guardmeth.html

the `get()` and `put()` will be called in different threads.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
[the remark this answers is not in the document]
chenghao-intel: As you see below, we've added the wake up in the `FiberCache.tryDispose()`, so it's not a problem.

BTW, since the queue actually keep adding the blocks by LRU from the Guava, hence probably a normal `ConcurrentLinkedQueue` is good enough for now, and it supposedly that the Fiber in the head of the queue will be the longest time not been used.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
linhong-intel: Ah, I see. Ok, I'll change to this.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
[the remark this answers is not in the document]
linhong-intel: sure.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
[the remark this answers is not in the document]
linhong-intel: if we add `synchronized` for `get()`, `onRemoval` can NOT be called by another thread. It's always called by `get()` and we have locked it.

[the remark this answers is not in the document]
chenghao-intel: yes, I agreed with u, sorry for the confusion. :)

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: Yes, it's synchronized :)

linhong-intel:
@chenghao-intel , @carsonwang @yueawang  for review. I've addressed most reviews.
here is a summary:
1. record pending cache size -> done
2. Change to `synchronized FiberCache` for `occupy`, `release`, `dispose` - > done
3. `synchronized FiberCacheManager.get()` -> done
4. Use `ConcurrentLinkedQueue` -> done
5. release function for CacheGuardian and FiberCacheManager -> added a TODO
6. More test case -> WIP
7. OOM may happen in current PR -> Shall we fix this in release 0.3 or just leave a log?

chenghao-intel:
Thank you @linhong-intel for the summary, for 5), probably you have to change if you meet any problem in writing the unit tests, but I am OK if you decide to put it into next release. for 7) let's just leave a warning log in this release.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,6 +46,55 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
...
+    def addRemovalFiber(fiberCache: FiberCache): Unit = {
+      _pendingFiberSize.addAndGet(fiberCache.size())
+      removalPendingQueue.add(fiberCache)
+    }
+
+    def removalPendingQueueSize: Int = removalPendingQueue.size()
+
+    override def run(): Unit = {
+      // Loop forever, TODO: provide a release function
+      while (true) {
```
chenghao-intel: Sorry, probably `LinkedBlockingQueue` is much better:
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/LinkedBlockingQueue.html

It provide the 
Is the take API a block API? Or the code like:
```scala
while (true) {
  val cache = removalPendingQueue.poll(200, TimeUnit.MILLISECONDS)
  if (cache != null) {
    ...
  }
  ...
}
```

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,6 +46,55 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
...
+    private val _pendingFiberSize: AtomicLong = new AtomicLong(0)
+
+    private val removalPendingQueue = new ConcurrentLinkedQueue[FiberCache]()
+
+    def addRemovalFiber(fiberCache: FiberCache): Unit = {
+      _pendingFiberSize.addAndGet(fiberCache.size())
+      removalPendingQueue.add(fiberCache)
+    }
+
+    def removalPendingQueueSize: Int = removalPendingQueue.size()
```
chenghao-intel: A better name will be `pendingSize` for short?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,6 +46,55 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
...
+           s"current used memory: ${_pendingFiberSize.get()}")
+        while (!removalPendingQueue.isEmpty) {
+          val fiberCache = removalPendingQueue.poll()
+          if (fiberCache != null) {
+            logDebug(s"Removing fiber $fiberCache ...")
+            // Block if fiber is in use.
+            while (!fiberCache.tryDispose(3000)) {
+              // Check memory usage every 3s while we are waiting fiber release.
+              if (_pendingFiberSize.get() > maxMemory) {
+                logError("Fibers pending on removal use too much memory")
```
chenghao-intel: put this log into the `addRemovalFiber` function also?

chenghao-intel: And it's better to print out the current and maximum memory in the log, which should be more informative to the end user.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,6 +46,55 @@ class OapFiberCacheHeartBeatMessager extends CustomManager with Logging {
...
+              }
+            }
+            // TODO: Make log more readable
+            _pendingFiberSize.addAndGet(-fiberCache.size())
+            logDebug(s"Fiber $fiberCache removed successfully")
+          }
+        }
+        // Sleep 1s if there is no pending fibers
+        logDebug("Processed all pending removal fibers, sleep 1 seconds")
+        Thread.sleep(1000)
```
chenghao-intel: this is not necessary if we used the `BlockedQueue`

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -92,17 +141,21 @@ object FiberCacheManager extends Logging {
...
-      .weigher(weigher)
-      .build[Fiber, FiberCache]()
-
-  def get(fiber: Fiber, conf: Configuration): FiberCache = {
-    // Used a flag called disposed in FiberCache to indicate if this FiberCache is removed
-    cache.get(fiber, cacheLoader(fiber, conf))
+  private val cache = CacheBuilder
+    .newBuilder()
+    .recordStats()
+    .concurrencyLevel(4) // TODO: Make 4 configurable
```
chenghao-intel: I know this is from history reason, but is there any meaningful constant value name instead of `4` in Guava?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -32,14 +33,44 @@ import org.apache.spark.unsafe.memory.{MemoryAllocator, MemoryBlock}
...
+    _refCount += 1
+  }
+
+  def release(): Unit = synchronized {
+    assert(refCount > 0, "release a non-used fiber")
+    _refCount -= 1
+    notifyAll()
+  }
+
+  def tryDispose(timeout: Long): Boolean = synchronized {
```
chenghao-intel: Nit: 
As most of `condition wait` will be a condition right before the `wait`, so I would like to suggest the code like:
```scala
val startTime = System.currentTimeMillis();
  if (_refCount > 0) {
    try { wait(100) } catch {..}
  } else {
    realDispose()
    return true;
  }
}
return false;
```

chenghao-intel:
@linhong-intel I leave some more comments, maybe you can take some of the suggestions in the updated code. please at me when you have done. I didn't take too much code in code review, but it's worth to well designed.  :)

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: agree, I also thought about `LinkedBlockingQueue`

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: sure

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: sure

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: there is a `DEFAULT_CONCURRENCY_LEVEL` in guava, and it is 4. :(
Anyway, i'm considering to remove it since use `synchronized get` will lose concurrency

linhong-intel:
@chenghao-intel 
updated.
1 note:
I removed `.concurrencyLevel(4)`. for 2 reasons:
cache default concurrency level is 4 and use `synchronized get` will lose concurrency anyway

linhong-intel:
Also updated some test cases focused on race condition, like:
multiple threads get 1 fiber
multiple threads get multiple fiber
multiple threads release 1 fiber
multiple threads get/release fiber at the same time
multiple threads get/remove fiber at the same time

src/test/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManagerSuite.scala
```diff
@@ -47,41 +51,182 @@ class FiberCacheManagerSuite extends SharedOapContext {
...
+    assert(FiberCacheManager.pendingSize == 0)
+  }
+
+  // refCount should be correct
+  test("release same fiber simultaneously") {
+    val pool = Executors.newCachedThreadPool()
+    val data = generateData(kbSize)
+    val fiber = TestFiber(() => MemoryManager.putToDataFiberCache(data), s"release test")
+    val fiberCaches = (1 to 5).map(_ => FiberCacheManager.get(fiber, configuration))
+    fiberCaches.foreach{ fiberCache =>
```
carsonwang: nit: need one space before {

src/test/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManagerSuite.scala
```diff
@@ -47,41 +51,182 @@ class FiberCacheManagerSuite extends SharedOapContext {
...
+        val data = generateData(memorySizeInMB / 4 * mbSize)
+        val fiber = TestFiber(() => MemoryManager.putToDataFiberCache(data), s"test fiber #2.$i")
+        val fiberCache = FiberCacheManager.get(fiber, configuration)
+        Thread.sleep(2000)
+        fiberCache.release()
+      }
+    }
+    val threads = (0 until 5).map(i => new FiberTestRunner(i))
+    threads.foreach(_.start())
+    threads.foreach(_.join())
```
carsonwang: No assert needed in this test? Do we want to fail it after a timeout?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManagerSuite.scala
```diff
@@ -47,41 +51,182 @@ class FiberCacheManagerSuite extends SharedOapContext {
...
+    val getWork: Unit = {
+      (1 to 1000).foreach { _ =>
+        val fiberCache = FiberCacheManager.get(fiber, configuration)
+        assert(!fiberCache.isDisposed)
+        fiberCache.release()
+      }
+    }
+    val removeWork: Unit = {
+      (1 to 1000).foreach { _ =>
+        FiberCacheManager.removeFiber(fiber)
```
carsonwang: Do we need `removeFiber` exactly after `FiberCacheManager.get` and check the fiber is still not disposed after a few seconds? Can we do that in a single thread as `removeFiber` only invalidate the cache and move the fiber to the pending removal queue? It seems we can't guarantee when `removeFilber` is called if it is in another thread? I mean `removeFiber` may be called after our assert.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManagerSuite.scala
```diff
@@ -47,41 +51,182 @@ class FiberCacheManagerSuite extends SharedOapContext {
...
+    val pool = Executors.newCachedThreadPool()
+    val data = generateData(kbSize)
+    val fiber = TestFiber(() => MemoryManager.putToDataFiberCache(data), s"release test")
+    val fiberCaches = (1 to 5).map(_ => FiberCacheManager.get(fiber, configuration))
+    fiberCaches.foreach{ fiberCache =>
+      pool.execute(new TestRunner(fiberCache.release()))
+    }
+    pool.shutdown()
+    pool.awaitTermination(1000, TimeUnit.MILLISECONDS)
+    assert(fiberCaches.head.refCount == 0)
```
carsonwang: Do we need assert the refCount is 5 initially?

linhong-intel: done

src/test/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManagerSuite.scala
linhong-intel: In out, "get" fiber from cache and "occupy" fiber are not atomic.
I just want to make sure these 2 steps are protected and no remove can happen between them.
I'll talk with you about this offline.

PR #495, opened 2018-01-03
Modify btree index reader and writer to support 2g rows

xuyangit:

1. Previously we write row id into a buffer and use writer to write them in oneshot, which may overflow for as many as 2g rows. Currently modified to writing them directly. 
2. Index file format. The row id list size set to long.
3. Reading index fiber in memory manager. Now use a multi-read style to avoid overflowing.

## How was this patch tested?

Test on 2g rows table in one partition for multi queries. There should be a test suite for this but the least memory requirement is 8G, so this needs some discussion here.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
@@ -19,6 +19,7 @@ package org.apache.spark.sql.execution.datasources.oap.filecache
 
 import java.util
 import java.util.concurrent.atomic.AtomicLong
+import javax.annotation.concurrent.GuardedBy
 
 import org.apache.hadoop.fs.FSDataInputStream
 import org.apache.spark.SparkEnv
@@ -32,14 +33,47 @@ import org.apache.spark.unsafe.memory.{MemoryAllocator, MemoryBlock}
 import org.apache.spark.unsafe.types.UTF8String
 
 // TODO: make it an alias of MemoryBlock
-trait FiberCache {
+trait FiberCache extends Logging {
+
   // In our design, fiberData should be a internal member.
   protected def fiberData: MemoryBlock
 
-  // TODO: need a flag to avoid accessing disposed FiberCache
+  @GuardedBy("FiberCache.this")
+  private var _refCount: Long = 0L
+  def refCount: Long = _refCount
+
+  def occupy(): Unit = synchronized {
+    _refCount += 1
+  }
+
+  def release(): Unit = synchronized {
+    assert(refCount > 0, "release a non-used fiber")
+    _refCount -= 1
+    notifyAll()
+  }
+
+  def tryDispose(timeout: Long): Boolean = synchronized {
+    val startTime = System.currentTimeMillis()
+    // Give caller a chance to deal with the long wait case.
+    while (System.currentTimeMillis() - startTime <= timeout) {
+      if (_refCount > 0) {
+        try {
+          wait(100)
+        } catch {
+          case _: InterruptedException =>
+            logWarning(s"Fiber Cache Dispose waiting detected for ${this}")
+        }
+      } else {
+        realDispose()
+        return true
+      }
+    }
+    false
+  }
+
   private var disposed = false
   def isDisposed: Boolean = disposed
-  def dispose(): Unit = {
+  private[filecache] def realDispose(): Unit = {
     if (!disposed) MemoryManager.free(fiberData)
     disposed = true
   }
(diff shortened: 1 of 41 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -165,19 +165,28 @@ private[oap] object MemoryManager extends Logging {
     logDebug(s"freed ${memoryBlock.size()} memory, used: $memoryUsed")
   }
 
-  // Used by IndexFile
+  // Used by IndexFile, modified as multi-read to load large row list (2G)
```
linhong-intel: Since row id list is partial loaded now. I don't think there will be a fiber has 2gb size.
Can we just put an assert here and fix it if we really have this problem.

xuyangit: @linhong-intel ,can u give more details about putting assert. And will we remove reading row id list completely from code in the future?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
```diff
@@ -198,11 +197,10 @@ private[index] case class BTreeIndexRecordWriter(
...
    */
-  private def serializeRowIdLists(uniqueKeys: Seq[InternalRow]): Array[Byte] = {
-    val buffer = new ByteArrayOutputStream()
-    uniqueKeys.flatMap(key =>
-      multiHashMap.get(key).asScala).foreach(IndexUtils.writeInt(buffer, _))
-    buffer.toByteArray
+  private def serializeAndWriteRowIdLists(uniqueKeys: Seq[InternalRow]): Unit = {
+    uniqueKeys.foreach(
+      key => multiHashMap.get(key).asScala.foreach(
+      x => fileWriter.writeRowIdList(IndexUtils.writeInt(x))))
```
linhong-intel: So you are writing row id one by one.
So, either change the function name to `writeRowId`, or keep the function and write rowIdList as a group, like the way you used in `putToIndexFiberCache`

Also, change you indent.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexUtils.scala
```diff
@@ -74,6 +74,10 @@ private[oap] object IndexUtils {
     out.write(v >>> 8 & 0xFF)
   }
 
+  def writeInt(v: Int): Array[Byte] = {
```
linhong-intel: maybe `toBytes` is better.
Anyway, this is not a "write" so change the name

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
linhong-intel: something like `assert(length < 100MB, "Fiber Size is larger than expected")`
Why we will remove reading row id list?
My suggestion is:
the row id list is partial loaded into multiple fibers, so each row id list fiber size should be less than 10MB. (By default, it's 1M * 4 Bytes = 4MB)

xuyangit:
@linhong-intel ,I have put an assert in read row id function instead. So if user set partition size too large, there can be a check here.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -205,7 +205,6 @@ private[oap] object MemoryManager extends Logging {
   def putToIndexFiberCache(in: FSDataInputStream, position: Long, length: Int): IndexFiberCache = {
     val bytes = new Array[Byte](length)
     in.readFully(position, bytes)
-
```
linhong-intel: I prefer keep this unchanged if no code style doc suggests this.

linhong-intel:
@carsonwang 
How do we manage bug fix PR after code freeze?

linhong-intel:
Fixes #457

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
xuyangit: Yeah. Maybe I don't change it back in my code. Can we just merge other files except this? Or I need to commit again. @linhong-intel

carsonwang:
@linhong-intel, let's create a 0.3 branch this week and bump the master's version to 0.4.0-snapshot. Critical bug fixes should be merged to both 0.3 and master while new features will only be merged to master. Before that, let's just merge to master.

carsonwang:
@linhong-intel , is this ready to merge?

Issue #484, opened 2017-12-27
"assertion failed: No plan for MetastoreRelation" with spark-sql


linhong-intel:
use spark-shell, there is no problem.

Suspect this is caused by `OapEnv`

My findings:
In `SparkSQLEnv`, `HiveSessionState` is used, and `HiveStrategies` is applied to `SparkPlanner`
In `OapEnv`, `OapSessionState` is used, and `OapStrategies` is applied to `SparkPlanner`

`OapStrategies` DOES NOT contain `HiveStrategies`
In `HiveStrategies` there is a `HiveTableScans` used to plan `MetastoreRelation`

linhong-intel:
found in baidu's production environment, don't have a way to easily reproduce in local env.

adrian-wang:
`OapStrategies` should include `HiveStrategies`, if we enable hive support.

linhong-intel:
fixed #501

PR #498, opened 2018-01-04
[Draft] another way to resolve relation

adrian-wang:

another way to resolve relation fully utilizing Spark's analyzer.

## How was this patch tested?

unit test


adrian-wang:
@Moonlit-Sailor for your information on #484

Issue #485, opened 2017-12-27
CreateIndex can't resolve MetastoreRelation


linhong-intel:
If use HiveCatalog, lookupRelation will return MetastoreRelation.
Now we only can resolve SimpleCatalogRelation.

Maybe below code can fix this:
```
   val relation =
      EliminateSubqueryAliases(sparkSession.sessionState.catalog.lookupRelation(table)) match {
        case r: SimpleCatalogRelation => (new FindDataSourceTable(sparkSession))(r)
        case other => sparkSession.sessionState.catalog match {
          case c: HiveSessionCatalog => c.ParquetConversions(other)
          case _ => other
        }
      }
```

adrian-wang:
I propose shift the resolve and conversion to Spark itself, as other operations do. In this case, Let's get rid of `RunnableCommand`, create corresponding `LogicalPlan` and `PhysicalPlan` for index operations. We only create `LogicalPlan` containing the relation as its child, and Spark will resolve all kinds of relations and do all possible conversion. Then we construct `PhysicalPlan` in `SparkPlanner`.

Assign this to @Moonlit-Sailor

adrian-wang:
closed via #499

PR #499, opened 2018-01-04
use spark QueryExecution to resolve relation during running index command

Moonlit-Sailor:
Before performing index operations(create/drop/refresh/show/check index),
create `QueryExecution` to resolve `unresolvedRelations`

## How was this patch tested?
HiveOapIndexDDLSuite

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
@@ -24,7 +24,6 @@ import org.antlr.v4.runtime.tree.TerminalNode
 
 import org.apache.spark.sql.{AnalysisException, SaveMode}
 import org.apache.spark.sql.catalyst.{FunctionIdentifier, TableIdentifier}
-import org.apache.spark.sql.catalyst.analysis.UnresolvedRelation
 import org.apache.spark.sql.catalyst.catalog._
 import org.apache.spark.sql.catalyst.parser._
 import org.apache.spark.sql.catalyst.parser.SqlBaseParser._
@@ -1413,7 +1412,7 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
     withOrigin(ctx) {
       CreateIndexCommand(
         ctx.IDENTIFIER.getText,
-        UnresolvedRelation(visitTableIdentifier(ctx.tableIdentifier())),
+        visitTableIdentifier(ctx.tableIdentifier()),
         visitIndexCols(ctx.indexCols),
         ctx.EXISTS != null, visitIndexType(ctx.indexType),
         Option(ctx.partitionSpec).map(visitNonOptionalPartitionSpec))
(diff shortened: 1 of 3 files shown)


Moonlit-Sailor:
Fixed #485

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -42,20 +40,19 @@ import org.apache.spark.sql.types._
 /**
  * Creates an index for table on indexColumns
  */
-case class CreateIndex(
+case class CreateIndexCommand(
     indexName: String,
-    table: TableIdentifier,
+    unresolvedRelation: LogicalPlan,
```
adrian-wang: don't use unresolvedRelation as name...

adrian-wang: Also, parameter type set to `LogicalPlan`

adrian-wang:
Closes #498

adrian-wang:
Better add a test for this.

src/test/scala/org/apache/spark/sql/hive/execution/HiveOapIndexDDLSuite.scala
```diff
@@ -0,0 +1,188 @@
...
+import org.apache.spark.sql.catalyst.catalog.CatalogTableType
+import org.apache.spark.sql.execution.datasources.oap.utils.OapUtils
+import org.apache.spark.sql.hive.test.TestHiveSingleton
+import org.apache.spark.sql.test.SQLTestUtils
+import org.apache.spark.util.Utils
+
+
+// test OAP Index DDL&DML on Hive tables
+class HiveOapIndexDDLSuite
+  extends QueryTest with SQLTestUtils with TestHiveSingleton with BeforeAndAfterEach {
```
adrian-wang: TestHiveSingleton will use object `TestHive`, which is not defined in our project, hence NullPointer....

yueawang: It's much similar to issue I met days ago, and I think there is no easy way to enable more than 1 Hive related suites because if we don't stop the hive context, next non-hive suite will report multi-context warning, once we stop is, next new hive suite would say no way to open metastore_db (as this JVM holds the file lock when the last hive context starts.).
Spark unit test actually splits the hive & non-hive casese by organizing them into different modules like sql/core & sql/hive, but OAP does not.
Let's see if we can find a better work around.

src/main/scala/org/apache/spark/sql/execution/SparkSqlParser.scala
```diff
@@ -1410,23 +1411,26 @@ class SparkSqlAstBuilder(conf: SQLConf) extends AstBuilder {
    */
   override def visitOapCreateIndex(ctx: OapCreateIndexContext): LogicalPlan =
     withOrigin(ctx) {
-      CreateIndex(
-        ctx.IDENTIFIER.getText, visitTableIdentifier(ctx.tableIdentifier()),
-        visitIndexCols(ctx.indexCols), ctx.EXISTS != null, visitIndexType(ctx.indexType),
+      CreateIndexCommand(
+        ctx.IDENTIFIER.getText,
+        UnresolvedRelation(visitTableIdentifier(ctx.tableIdentifier())),
```
linhong-intel: can we keep passing `TableIdentifier` here and Warp it by `UnresolvedRelation` in `CreateIndex`.
Not sure which one is better.
If no reason to change, I prefer the `TableIdentifier`

adrian-wang:
LGTM

PR #501, opened 2018-01-05
Add hive strategies to oap session

yueawang:

1. Add hive related strategies to OapSessionState as it inherits from HiveSessionStage.
2. Create TestOap (just like TestHive in Spark) which takes the role of oap test context.
3. Modify related unit test to see if OapSession works fine.

## How was this patch tested?
mvn test passes.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/TestOap.scala
+++ b/src/main/scala/org/apache/spark/sql/TestOap.scala
@@ -20,8 +20,7 @@ import org.apache.spark.SparkConf
 import org.apache.spark.sql.internal.SQLConf
 
 
-object TestOap
-  extends TestOapContext(
+object TestOap extends TestOapContext(
   OapSession.builder.config(
     (new SparkConf).set("spark.master", "local[2]")
       .set("spark.app.name", "test-oap-context")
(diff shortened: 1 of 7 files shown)


src/main/scala/org/apache/spark/sql/TestOap.scala
```diff
@@ -0,0 +1,45 @@
...
+ */
+package org.apache.spark.sql
+
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.internal.SQLConf
+
+
+object TestOap
+  extends TestOapContext(
+  OapSession.builder.config(
```
carsonwang: nit: need more indent here or move `extends TestOapContext(` to the upper line?

src/main/scala/org/apache/spark/sql/hive/OapSessionState.scala
```diff
@@ -35,18 +35,26 @@ class OapSessionState(sparkSession: OapSession) extends HiveSessionState(sparkSe
 
   override def planner: SparkPlanner = {
     new SparkPlanner(sparkSession.sparkContext, conf, experimentalMethods.extraStrategies)
-    with OapStrategies
+    with HiveStrategies with OapStrategies
```
carsonwang: nit: Indent

yueawang:
For concern
> 2. The test case is not cover #484 . I mean, no regression test.

Yi has already created hive context related test case, I will merge them during rebase as there is only one Hive context allowed.

linhong-intel:
>Yi has already created hive context related test case, I will merge them during rebase as there is only one Hive context allowed.

and .gitignore? then we can merge it.

yueawang:
Done!

PR #505, opened 2018-01-11
fix Enable/Disable using index test case of OapSuite

Moonlit-Sailor:
minor fix for Enable/Disable using OAP index Test Case:
Set `OAP_ENABLE_EXECUTOR_INDEX_SELECTION` to `false` in order to
use index forcely
## How was this patch tested?
OapSuite


Moonlit-Sailor:
closed due to #507

PR #507, opened 2018-01-11
Fix issue 502: all test should use configuration from ShareOapContext.

yueawang:
Fix issue 502, all case suites based on ShareOapContext should use
configuration in it.

## How was this patch tested?
mvn passes.


linhong-intel:
One thing, since index is partial loaded, there should be a better way to check only needed index data size.

yueawang:
Yes, we need a better way to predict the index scan size, and choose threshold accordingly, so far this file size is just a base (worst? :-)) model.

PR #521, opened 2018-01-16
Increase sleep time in test Simple Cache Strategy

linhong-intel:

To fix the flaky unit test "test Simple Cache Strategy"

## How was this patch tested?

Unit test pass


linhong-intel:
It's a very rare failure since I never see this on my local machine.
no side effect. merge directly.

carsonwang:
@linhong-intel, since we have a 0.3 branch. This need to be merged to both master and branch-0.3. Do you know a good way to do this or create a seperated PR?

linhong-intel:
Sorry, don’t know. If we want to track everything. Another PR should be the choice.

PR #530, opened 2018-01-17
Add support for testing in localClusterMode

gczsjdy:

This is to enable testing in `localClusterMode`, as Spark docs indicated: `Running the spark.deploy.master.Master and spark.deploy.worker.Workers in the same JVMs. Executors launched by the Workers still run in separate JVMs. This can be used to test distributed operation and
 fault recovery without spinning up a lot of processes.`

You can write any test suites as you like, just need to extend `SharedOapLocalClusterContext`.

## How was this patch tested?

org.apache.spark.sql.execution.datasources.oap.cluster.FilterSuite


linhong-intel:
Should we change the travis to test both local and cluster?

gczsjdy:
1. Good to me, we can make each testSuite choose its base class so as to decide running local or cluster test. But there's 1 issue: Is there any test cases we want to test both locally and on cluster? How to deal with them? 
2. & 3. Based on your advice: Two `OAPContext`s : `SharedOapContext` & `SharedClusterOapContext` are needed, and corresponding `LocalTestSession` & `ClusterTestSession`.
Compared to my original solution, we can't dynamically decide to run a certain case locally or on cluster. But this is minor?

gczsjdy:
@linhong-intel I have modified the code and tested manually, it should be no problem.

linhong-intel:
@gczsjdy . Sorry I missed your update. This PR is good to me now.
But I suggest more peer reviews. Since it's kind like a new feature. It's better everyone is aware of this PR.

@carsonwang @adrian-wang @lee-lei @yueawang

pom.xml
```diff
@@ -518,4 +520,25 @@
...
+        <local.cluster>false</local.cluster>
+      </properties>
+    </profile>
+    <profile>
+      <activation>
+        <activeByDefault>false</activeByDefault>
+      </activation>
+      <id>cluster</id>
+      <properties>
+        <local.cluster>true</local.cluster>
```
adrian-wang: Do we need a profile here, or just `-Dlocal.cluster=true` works as expected?

src/test/scala/org/apache/spark/sql/test/TestOapSession.scala
```diff
@@ -0,0 +1,43 @@
...
+        new SparkContext(s"local-cluster[$numExecutor, $cores, $memory]",
+          "test-cluster-sql-context",
+          sparkConf.set("spark.sql.testkey", "true"))
+      } else {
+        new SparkContext("local[2]",
+          "test-sql-context",
+          sparkConf.set("spark.sql.testkey", "true"))
+      })
+  }
+}
```
adrian-wang: Nit: new line at end

src/test/scala/org/apache/spark/sql/test/TestOapSession.scala
```diff
@@ -0,0 +1,43 @@
...
+private[sql] class TestOapSession(sc: SparkContext) extends TestSparkSession(sc) { self =>
+
+  def this(sparkConf: SparkConf) {
+    this(
+      if (sys.props.get("spark.test.localCluster.enabled") == Some("true")) {
+        val numExecutor = 3
+        val cores = 5
+        val memory = 1024
+        new SparkContext(s"local-cluster[$numExecutor, $cores, $memory]",
+          "test-cluster-sql-context",
```
adrian-wang: Nit: test-cluster-oap-context?

src/test/scala/org/apache/spark/sql/test/oap/SharedOapContext.scala
```diff
@@ -21,15 +21,21 @@ import scala.collection.mutable
...
-import org.apache.spark.sql.test.SharedSQLContext
+import org.apache.spark.sql.test.{SharedSQLContext, TestOapSession, TestSparkSession}
 
 trait SharedOapContext extends SharedSQLContext {
 
   // avoid the overflow of offHeap memory
   sparkConf.set("spark.memory.offHeap.size", "100m")
+
+  protected override def createSparkSession: TestSparkSession = {
+    new TestOapSession(sparkConf.set("spark.hadoop.fs.file.impl", classOf[DebugFilesystem].getName))
```
adrian-wang: better move the conf setting to where you initiate `SparkContext`

pom.xml
gczsjdy: Actually this is a profile, you can activate it by `-Pcluster` or `-Dlocal.cluster=true`

src/test/scala/org/apache/spark/sql/test/oap/SharedOapContext.scala
gczsjdy: Good catch!

gczsjdy:
Ping @linhong-intel @adrian-wang Any new comments?

adrian-wang:
There should also be some advice on how to make use of this in the Doc

adrian-wang:
LGTM

PR #531, opened 2018-01-18
Add few information to memory assert message.

yueawang:

Add few information to assert message to help debugging.

## How was this patch tested?
mvn passes.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
@@ -171,6 +171,7 @@ private[oap] object MemoryManager extends Logging {
     assert(SparkEnv.get != null, "Oap can't run without SparkContext")
     val memoryManager = SparkEnv.get.memoryManager
     // TODO: make 0.7 configurable
+    assert(memoryManager.maxOffHeapStorageMemory > 0, "Oap can't run without offHeap memory")
     val oapMemory = (memoryManager.maxOffHeapStorageMemory * 0.7).toLong
     if (memoryManager.acquireStorageMemory(
       DUMMY_BLOCK_ID, oapMemory, MemoryMode.OFF_HEAP)) {
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
@@ -26,6 +26,7 @@ import com.google.common.cache._
 import org.apache.hadoop.conf.Configuration
 
 import org.apache.spark.internal.Logging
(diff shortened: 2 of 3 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -171,6 +171,7 @@ private[oap] object MemoryManager extends Logging {
     assert(SparkEnv.get != null, "Oap can't run without SparkContext")
     val memoryManager = SparkEnv.get.memoryManager
     // TODO: make 0.7 configurable
+    assert(memoryManager.maxOffHeapStorageMemory > 0, "Oap can't run without offHeap memory")
```
linhong-intel: Thanks. this is what I always forgot.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -124,7 +124,8 @@ class GuavaOapCache(cacheMemory: Long, cacheGuardianMemory: Long) extends OapCac
   override def get(fiber: Fiber, conf: Configuration): FiberCache = {
     val fiberCache = cache.get(fiber, cacheLoader(fiber, conf))
     // Avoid loading a fiber larger than MAX_WEIGHT / 4, 4 is concurrency number
-    assert(fiberCache.size() <= MAX_WEIGHT * MB / 4, "Can't cache fiber larger than MAX_WEIGHT / 4")
+    assert(fiberCache.size() <= MAX_WEIGHT * MB / 4,
+      s"Can't cache fiber(${fiberCache.size()}) larger than MAX_WEIGHT($MAX_WEIGHT) / 4")
```
linhong-intel: Pleas use the Utils.bytesToString to format size.

linhong-intel: Also, please hep one of my mistake.
I remove the concurrencyLevel already, so we need to remove `4` as well.

yueawang: I remember guava has a default concurrency number 4?

linhong-intel: you are right..... my bad

carsonwang: Not related to this PR. I just have a little concern on the hard coded 4 here. If guava changes the default concurrency in the next version and we update to that version, it is very likely we will forget to update this number here. If there is no variable in guava that represents the default concurrency, we may still want to specify the concurrency ourselves and use the same variable here. @linhong-intel

adrian-wang:

PR #532, opened 2018-01-18
Change Cache Unit from MB to KB

linhong-intel:

If use MB as Unit, we'll lose too much memory due to ceil.
Say, 100KB will ceil to 1MB, 90% memory is wasted.

## How was this patch tested?

Unit test pass


linhong-intel:
Clone of #529

PR #534, opened 2018-01-19
support startswith in btree scanner

adrian-wang:

support startswith in btree scanner.
Still need to modify statisticsReader accordingly.

## How was this patch tested?

add a unit test.


adrian-wang:
bug is fixed, I'll refactor the code a little bit and then work on statistics.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/UnsafeIndexTree.scala
```diff
@@ -210,25 +210,28 @@ private[oap] class CurrentKey(node: IndexNode, keyIdx: Int, valueIdx: Int, index
...
     includeStart: Boolean,
     includeEnd: Boolean,
+    ignoreTail: Boolean = false,
     isNull: Boolean = false) extends Serializable {
   var start = s
   var end = e
   var startInclude = includeStart
   var endInclude = includeEnd
   val isNullPredicate = isNull
+  val isPrefixMatch = ignoreTail
```
linhong-intel: I'm thinking about creating a another Interval class, say `StartWithInterval`.
Then `RangeInterval` and `StartWithInterval` extends `Interval`.
Then we use pattern match to replace `if .. else ...` in your current code.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
```diff
@@ -187,6 +199,14 @@ private[index] case class BTreeIndexRecordReader(
     }
   }
 
+  private[index] def rowOrderingPattern(x: InternalRow, y: InternalRow, isStart: Boolean): Int = {
+    require(x.numFields == 1 && y.numFields == 1)
```
linhong-intel: How to satisfy this require since btree support multi-column index?
Should we add some check when building index scanner and contains string column?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -207,9 +208,10 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
 
   }
 
-  private def buildScanner(lastIdx: Int, bestIndexer: IndexMeta, intervalMap:
-    mutable.HashMap[String, ArrayBuffer[RangeInterval]],
-    options: Map[String, String] = Map.empty): IndexScanner = {
+  private def buildScanner(
+      lastIdx: Int, bestIndexer: IndexMeta,
```
linhong-intel: I'm a little bit confused about the indent of function parameters.
Should they be in separate lines?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexUtils.scala
```diff
@@ -114,19 +114,73 @@ private[oap] object IndexUtils {
...
       else e = m - 1
     }
     if (!found) m = s
     (m, found)
   }
+
+  def binarySearchForStart(
+      start: Int, length: Int,
+      keys: Int => InternalRow, candidate: InternalRow,
+      compare: (InternalRow, InternalRow) => Int): (Int, Boolean) = {
```
linhong-intel: eh.... Do you mind to think about how to combine there 2 functions?
I tried and failed. But I believe there should be an elegant way...
Or we can file an issue to reconsider the BTree search logic.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
adrian-wang: Not a must.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/UnsafeIndexTree.scala
adrian-wang: Only that would lead to a total rewrite of scanner builder...

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexUtils.scala
adrian-wang: Yea, currently in status of this PR, we would have 3 different binary search methods, each used once. I think generalizing would involve pass variables like initial methods, hit methods, boundary check etc.,... not sure the code would be elegant that way.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
adrian-wang: You're right.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexUtils.scala
```diff
@@ -114,19 +114,73 @@ private[oap] object IndexUtils {
       start: Int, length: Int,
       keys: Int => InternalRow, candidate: InternalRow,
       compare: (InternalRow, InternalRow) => Int): (Int, Boolean) = {
-    var s = 0
+    var s = start
     var e = length - 1
     var found = false
     var m = s
-    while (s <= e & !found) {
+    while (s <= e && !found) {
```
carsonwang: Oh, I think we need this as a single fix to 0.3.

adrian-wang: Just took a closer look, the difference is
`false && (throw new Exception())` => `false`
`false & (throw new Exception())` => Exception

adrian-wang:
@linhong-intel Thanks, I will fix this.

linhong-intel:
@adrian-wang 
About the 2nd problem.
please check the output of BPlusTreeScanner since Spark will filter out some results.

adrian-wang:
@linhong-intel 
This is because currently in this PR I didn't implement support for condition merge between pattern condition and others.
And with the interval you provided, it is empty actually, because no `b` is between `this3` and `this3`, for exact match.

linhong-intel:
@adrian-wang Please ignore my previous comment, here is the update.

Actually, the interval is:
```
start = {[3] + [this3]}
end = {[3] + [this3]}
startInclude = true
endInclude = true
isNull = false
isPrefixMatch = true <== It is TRUE, and we will use rowOrderingPattern, then there is an exception
```

> java.lang.Integer cannot be cast to org.apache.spark.unsafe.types.UTF8String

adrian-wang:
yes in `rowOrderingPattern` I only considered there is only one filter.

linhong-intel:
and below query will get EMPTY result. Will you fix or WA these two issues?
```
sql("create oindex index1 on oap_test_1 (b, a) using btree")  <== swapped (a,b) to (b,a)
sql("SELECT * FROM oap_test_1 WHERE a = 3 and b like 'this3%'").show()
```

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -89,8 +91,8 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
           if (flag == 1) {
             num += 1
           }
-          if (num > 0) {
-            availableIndexes.append((num - 1, indexMeta))
+          if (num>0) {
+            availableIndexes.append((num-1, indexMeta))
```
linhong-intel: nit: space

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordReader.scala
```diff
@@ -201,6 +218,40 @@ private[index] case class BTreeIndexRecordReader(
...
+      if (x.numFields > 1) {
+        val partialRes = partialOrdering.compare(x, y)
+        if (partialRes != 0) {
+          return partialRes
+        }
+      }
+      val xStr = x.getString(schema.length - 1)
+      val yStr = y.getString(schema.length - 1)
+      if (strMatching(xStr, yStr)) 0 else xStr.compare(yStr)
+    } else if (x.numFields < y.numFields) {
```
linhong-intel: How could `x.numFields < y.numFields` and `x.numFields > y.numFields` happen.
If `x` is fixed to be from `interval` and `interval. isPrefixMatch  == true`

adrian-wang: `x.numFields < y.numFields`: when index is like (a,b), while the filter is like `a like 'test%'`;
`x.numFields > y.numFields`: when index is like (a,b), while the filter is like `a=1 and b = 2 and c like 'test%'`

linhong-intel: schema in IndexScanner is the *USED* columns in index, not *ALL* columns in index.
So,
`a like 'test% => x.numFields = 1, y.numFields = 1`
`a=1 and b = 2 and c like 'test%' => x.numFields = 2, y.numFields = 2`


related code:
```
val fields = indexFields.slice(0, lastIdx + 1)
keySchema = StructType(fields)
```


I also tried you test cases, and the two ifs not happen:
"startswith using multi-dimension index"
"startswith using multi-dimension index - multi-filters"

linhong-intel:
good to me

adrian-wang:
merging this, thanks @linhong-intel @carsonwang

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapIndexQuerySuite.scala
```diff
@@ -126,4 +126,53 @@ class OapIndexQuerySuite extends QueryTest with SharedOapContext with BeforeAndA
 
     sql("drop oindex bmidx1 on oap_test_1")
   }
+
+  test("startswith using index") {
+    sqlConf.setConf(OapConf.OAP_EXECUTOR_INDEX_SELECTION_STATISTICS_POLICY, false)
+    val data: Seq[(Int, String)] =
```
zhixingheyi-tian: @adrian-wang 

Why disable OAP_EXECUTOR_INDEX_SELECTION_STATISTICS_POLICY?
If  enable OapConf.OAP_EXECUTOR_INDEX_SELECTION_STATISTICS_POLICY.  "startswith using index"  will work incorrectly?

PR #535, opened 2018-01-19
[Dont Merge]Try to use rwlock solve cache issue.

yueawang:

Try differentiate the miss-cache-load & in-cache-hit.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
@@ -19,6 +19,7 @@ package org.apache.spark.sql.execution.datasources.oap.filecache
 
 import java.util.concurrent.Callable
 import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.locks.{Lock, ReentrantReadWriteLock}
 
 import scala.collection.JavaConverters._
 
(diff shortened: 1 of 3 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -121,11 +122,32 @@ class GuavaOapCache(cacheMemory: Long, cacheGuardianMemory: Long) extends OapCac
...
+      rwLock.writeLock().unlock()
+      return fiberInCache
+    }
+    rwLock.writeLock().unlock()
+
+    rwLock.readLock().lock()
     val fiberCache = cache.get(fiber, cacheLoader(fiber, conf))
     // Avoid loading a fiber larger than MAX_WEIGHT / 4, 4 is concurrency number
-    assert(fiberCache.size() <= MAX_WEIGHT * KB / 4, "Can't cache fiber larger than MAX_WEIGHT / 4")
     fiberCache.occupy()
```
linhong-intel: I think I get your point.
You want all the cache hit case not block by cache load.
But the code has some problem.
In this readLock. All cache load can run parallel, so it means onRemoval and cacheLoad can happen at the same time. Then, right before `occupy` and after `cache.get`, the fiber maybe removed.
This is found by Hao and is our mainly reason to add synchronize.
Maybe you can swap the read/write lock, I'm not sure it's ok.
Also you can check a "double check locking", it's similar to this case.

yueawang: Yes, simple lock logic like this seems do not help here, we need some other method to separate occupy and get. let's discuss this later.

PR #538, opened 2018-01-22
[BRANCH-0.3] Use Global Lock For Each Fiber To Increase Concurrency

linhong-intel:

Use a global lock for each fiber.
fiber lock must be acquired before change this Fiber/FiberCache

## How was this patch tested?

Unit test

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -19,6 +19,7 @@ package org.apache.spark.sql.execution.datasources.oap.filecache
 
 import java.util.concurrent.{ConcurrentHashMap, LinkedBlockingQueue, TimeUnit}
 import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.locks.ReentrantReadWriteLock
 
 import com.google.common.cache._
 import org.apache.hadoop.conf.Configuration
@@ -60,14 +61,12 @@ private[filecache] class CacheGuardian(maxMemory: Long) extends Thread with Logg
     while (true) {
       val (fiber, fiberCache) = removalPendingQueue.take()
       logDebug(s"Removing fiber $fiberCache ...")
-      FiberLockManager.getFiberLock(fiber).synchronized {
-        // Block if fiber is in use.
-        while (!fiberCache.tryDispose(3000)) {
-          // Check memory usage every 3s while we are waiting fiber release.
-          if (_pendingFiberSize.get() > maxMemory) {
-            logWarning("Fibers pending on removal use too much memory, " +
-              s"current: ${_pendingFiberSize.get()}, max: $maxMemory")
-          }
+      // Block if fiber is in use.
+      while (!fiberCache.tryDispose(fiber, 3000)) {
+        // Check memory usage every 3s while we are waiting fiber release.
+        if (_pendingFiberSize.get() > maxMemory) {
+          logWarning("Fibers pending on removal use too much memory, " +
+            s"current: ${_pendingFiberSize.get()}, max: $maxMemory")
         }
       }
       // TODO: Make log more readable
@@ -273,14 +272,18 @@ private[oap] case class TestFiber(getData: () => FiberCache, name: String) exten
 }
 
 object FiberLockManager {
-  val lockMap = new ConcurrentHashMap[Fiber, Object]()
-  def getFiberLock(fiber: Fiber): Object = {
+  private val lockMap = new ConcurrentHashMap[Fiber, ReentrantReadWriteLock]()
+  def getFiberLock(fiber: Fiber): ReentrantReadWriteLock = {
     var lock = lockMap.get(fiber)
     if (lock == null) {
-      val newLock = new Object()
+      val newLock = new ReentrantReadWriteLock()
       val prevLock = lockMap.putIfAbsent(fiber, newLock)
       lock = if (prevLock == null) newLock else prevLock
     }
     lock
+  }
+
+  def removeFiberLock(fiber: Fiber): Unit = {
+    lockMap.remove(fiber)
   }
 }
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -17,7 +17,7 @@
 
 package org.apache.spark.sql.execution.datasources.oap.filecache
 
-import java.util.concurrent.{LinkedBlockingQueue, TimeUnit}
+import java.util.concurrent.{ConcurrentHashMap, LinkedBlockingQueue, TimeUnit}
 import java.util.concurrent.atomic.AtomicLong
 
 import com.google.common.cache._
@@ -42,13 +42,13 @@ private[filecache] class CacheGuardian(maxMemory: Long) extends Thread with Logg
 
   private val _pendingFiberSize: AtomicLong = new AtomicLong(0)
 
-  private val removalPendingQueue = new LinkedBlockingQueue[FiberCache]()
+  private val removalPendingQueue = new LinkedBlockingQueue[(Fiber, FiberCache)]()
 
   def pendingSize: Int = removalPendingQueue.size()
 
-  def addRemovalFiber(fiberCache: FiberCache): Unit = {
+  def addRemovalFiber(fiber: Fiber, fiberCache: FiberCache): Unit = {
     _pendingFiberSize.addAndGet(fiberCache.size())
-    removalPendingQueue.offer(fiberCache)
+    removalPendingQueue.offer(fiber, fiberCache)
     if (_pendingFiberSize.get() > maxMemory) {
       logWarning("Fibers pending on removal use too much memory, " +
           s"current: ${_pendingFiberSize.get()}, max: $maxMemory")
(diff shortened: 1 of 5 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -269,3 +271,16 @@ private[oap] case class TestFiber(getData: () => FiberCache, name: String) exten
...
 }
+
+object FiberLockManager {
+  val lockMap = new ConcurrentHashMap[Fiber, Object]()
+  def getFiberLock(fiber: Fiber): Object = {
+    var lock = lockMap.get(fiber)
+    if (lock == null) {
+      val newLock = new Object()
+      val prevLock = lockMap.putIfAbsent(fiber, newLock)
+      lock = if (prevLock == null) newLock else prevLock
```
chenghao-intel: is it possible to return a `null` `prevLock`? Should we put the `assert` instead of `if`

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -269,3 +271,16 @@ private[oap] case class TestFiber(getData: () => FiberCache, name: String) exten
     s"type: TestFiber name: $name"
   }
 }
+
+object FiberLockManager {
+  val lockMap = new ConcurrentHashMap[Fiber, Object]()
+  def getFiberLock(fiber: Fiber): Object = {
+    var lock = lockMap.get(fiber)
```
chenghao-intel: Maybe a double check is better.
```scala
var lock = lockMap.get(fiber)
if (lock == null) {
  synchronized(this) {
    lock = lockMap.get(fiber) {
      if (lock == null) {
         lock = lockMap.put(fiber, new ReentrantReadWriteLock())
      }
    }
  }
  lock
}
```

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -122,15 +122,16 @@ class GuavaOapCache(cacheMemory: Long, cacheGuardianMemory: Long) extends OapCac
...
-    val fiberCache = cache.get(fiber, cacheLoader(fiber, conf))
-    // Avoid loading a fiber larger than MAX_WEIGHT / 4, 4 is concurrency number
-    assert(fiberCache.size() <= MAX_WEIGHT * KB / 4,
-      s"Failed to cache fiber(${Utils.bytesToString(fiberCache.size())}) " +
-        s"with cache's MAX_WEIGHT(${Utils.bytesToString(MAX_WEIGHT.toLong * KB.toLong)}) / 4")
-    fiberCache.occupy()
-    fiberCache
-  }
+  override def get(fiber: Fiber, conf: Configuration): FiberCache =
+    FiberLockManager.getFiberLock(fiber).synchronized {
```
chenghao-intel:
```scala
val lock = ReentrantReadWriteLock.readLock()
try {
lock.lock()
...
} finally {
  lock.unlock()
}
```

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -269,3 +271,16 @@ private[oap] case class TestFiber(getData: () => FiberCache, name: String) exten
     s"type: TestFiber name: $name"
   }
 }
+
+object FiberLockManager {
+  val lockMap = new ConcurrentHashMap[Fiber, Object]()
+  def getFiberLock(fiber: Fiber): Object = {
```
chenghao-intel: `def getFiberLock(fiber: Fiber): ReentrantReadWriteLock`

linhong-intel:
A simple figure to show we are thread-safe **even if fiber is evicted before occupy**

screen shot 2018-01-22 at 1 40 39 pm

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
chenghao-intel: readLock will allow multiple thread to "READ" the data.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -58,14 +58,16 @@ private[filecache] class CacheGuardian(maxMemory: Long) extends Thread with Logg
...
     while (true) {
-      val fiberCache = removalPendingQueue.take()
+      val (fiber, fiberCache) = removalPendingQueue.take()
       logDebug(s"Removing fiber $fiberCache ...")
-      // Block if fiber is in use.
-      while (!fiberCache.tryDispose(3000)) {
-        // Check memory usage every 3s while we are waiting fiber release.
-        if (_pendingFiberSize.get() > maxMemory) {
-          logWarning("Fibers pending on removal use too much memory, " +
+      FiberLockManager.getFiberLock(fiber).synchronized {
```
chenghao-intel:
```scala
val lock = ReentrantReadWriteLock.writeLock()
try {
lock.tryLock(200, TimeUnit.Millisecond)
...
} finally {
  lock.unlock()
}
```
But instead of using lock here, actually I am suggesting to using the lock within method `tryDispose`.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: `putIfAbsent` will return the previous value.
So, if there is no value for key, putIfAbsent will put the value into map and return null

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
linhong-intel: I think below piece of code is equivalent to `ConcurrentHashMap.putIfAbsent`
```scala
synchronized(this) {
    lock = lockMap.get(fiber) {
      if (lock == null) {
         lock = lockMap.put(fiber, new ReentrantReadWriteLock())
      }
    }
```

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
linhong-intel: good idea.

linhong-intel:
@chenghao-intel 
updated to a lock version.
Didn't change the code logic in `FiberLockManager`. you can check my reply to your review. Then decide if we should use your code.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -123,13 +123,19 @@ class GuavaOapCache(cacheMemory: Long, cacheGuardianMemory: Long) extends OapCac
...
+    try {
+      val fiberCache = cache.get(fiber, cacheLoader(fiber, conf))
+      // Avoid loading a fiber larger than MAX_WEIGHT / 4, 4 is concurrency number
+      assert(fiberCache.size() <= MAX_WEIGHT * KB / 4,
+        s"Failed to cache fiber(${Utils.bytesToString(fiberCache.size())}) " +
+            s"with cache's MAX_WEIGHT(${Utils.bytesToString(MAX_WEIGHT.toLong * KB.toLong)}) / 4")
+      fiberCache.occupy()
+      fiberCache
+    } finally {
+      readLock.unlock()
```
yueawang: Can readlock replace refcount if we need a readlock here?? Just thought we don't need to check if refcount == 0, but we can check if all readlock are unlocked.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -52,8 +53,10 @@ trait FiberCache extends Logging {
...
   }
 
-  def tryDispose(timeout: Long): Boolean = synchronized {
+  // TODO: Couple Fiber and FiberCache. Pass fiber as a parameter is weired.
+  def tryDispose(fiber: Fiber, timeout: Long): Boolean = synchronized {
     val startTime = System.currentTimeMillis()
+    val writeLock = FiberLockManager.getFiberLock(fiber).writeLock()
     // Give caller a chance to deal with the long wait case.
     while (System.currentTimeMillis() - startTime <= timeout) {
       if (_refCount > 0) {
```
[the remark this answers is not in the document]
linhong-intel: I see your point. will change to lock area.
And if we want to remove `synchronized` then we need lock `release()` and `occupy` need be locked by caller.
Let me update the PR.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
linhong-intel: @yueawang 
about guava limitation, I think guava will not block other access during load a new fiber (means guava will release its lock during loading). so, concurrency should be ok. And, change concurrency level involves profiling I'm not sure we have time to do this. But I'm ok with changing the concurrency level.
about checking readlock unlock, it's not enough, we release readLock right after refCount+1, but upper layer caller will release the fiber and refCount-1 after read is done. We have to wait for this.

yueawang: No, I mean delay the readlock.unlock to the same place where refcount -1.

linhong-intel: Yes, I got your point.
This is one previous implementation. The problem is, put lock and unlock into different function is uncontrollable, I'm a little bit worry about this idea.
What's more, "multiple index union" feature has conflict with this idea. I can talk with you offline if you have interest

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -38,43 +38,51 @@ trait FiberCache extends Logging {
...
+            return true
+          }
+        } finally {
+          writeLock.unlock()
         }
-      } else {
-        realDispose()
-        return true
       }
+      Thread.sleep(100)
```
chenghao-intel: Can we remove this?

linhong-intel:
Thanks @xuyangit for the test result. paste some here. Full result is attached.
**The result can prove this PR FIXED the cache performance regression.**
```
Previous version:
+---------------------+------+-----+-----+---------+
|                   Q1| T1/ms|T2/ms|T3/ms|Median/ms|
+---------------------+------+-----+-----+---------+
|       oap-with-index|163948|77024|73777|    77024|
|    oap-without-index|132952|59206|28460|    59206|
|   parquet-with-index| 38235|34372|33119|    34372|
|parquet-without-index| 19925|14726|14823|    14823|
+---------------------+------+-----+-----+---------+
Current version:
+---------------------+-----+-----+-----+---------+
|                   Q1|T1/ms|T2/ms|T3/ms|Median/ms|
+---------------------+-----+-----+-----+---------+
|       oap-with-index|28101| 9902| 9049|     9902|
|    oap-without-index|27591|13915|11482|    13915|
|   parquet-with-index|37864|34535|35903|    35903|
|parquet-without-index|20873|14976|13816|    14976|
+---------------------+-----+-----+-----+---------+
```

From the log, parquet performance is identical, but oap performance increased with this PR.

linhong-intel:
I'll merge this by today since it's a blocker. @carsonwang @chenghao-intel

PR #539, opened 2018-01-22
[0.3.0 QA] Enable statistics for Bitmap scanner.

lee-lei:
Add the statistics before the bitmap footer.
And keep the bitmap footer size unchanged for backward compatibility.

## How was this patch tested?

mvn clean test package

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapIndexRecordWriter.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapIndexRecordWriter.scala
@@ -41,8 +41,8 @@ private[oap] object BitmapIndexSectionId {
   val entryListSection    : Int = 3 // bitmap entry list
   val entryNullSection    : Int = 4 // bitmap entry for null value rows
   val entryOffsetsSection : Int = 5 // bitmap entry offset list
-  val statisticsSection   : Int = 6 // keep the original statistics, not changed than before.
-  val footerSection       : Int = 7 // footer to save total key list size and length, total entry
+  val statsContentSection : Int = 6 // keep the original statistics, not changed than before.
+  val footerSection       : Int = 7 // footer to save the meta data for the above segments.
 }
 
 /* Below is the bitmap index general layout and sections.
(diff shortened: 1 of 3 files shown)


lee-lei:
Got it. Thanks.

lee-lei:
The failed test case is FiberCacheManagerSuite. It actually passed after running this single test case, since this patch doesn't touch FiberCacheManager. Thanks.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapIndexRecordWriter.scala
```diff
@@ -181,6 +186,10 @@ private[oap] class BitmapIndexRecordWriter(
     // The beginning of the footer are bitmap total size, key list size and offset total size.
     // Others keep back compatible and not changed than before.
     bmIndexEnd = bmEntryListOffset + bmEntryListTotalSize + bmNullEntrySize + bmOffsetListTotalSize
+    // The index end is also the starting position of statistics part.
+    val statisticsWriteManager = new StatisticsWriteManager
+    statisticsWriteManager.initialize(BitMapIndexType, keySchema, configuration)
+    val statSize = statisticsWriteManager.write(writer)
```
adrian-wang: This will write nothing but an empty statistics.

adrian-wang:
Can you enable those tests ignored in `StatisticsManagerSuite`

lee-lei:
Yes. I found nabled that yesterday.  And the tests passed.

yueawang:
Bitmap statistics may need a different threshold to use/skip index, please run a few rounds of benchmark to figure out it.

lee-lei:
Thanks for yue's remind. I checked the statistics manager. And there is only one case to use full scan. It's in part by value. Thus the probability of full scan is 0.25(1/4) which is great than 0.2 which is the current threshold. If the analysis result is below this threshold, then use index. If the result (0.25) is great than the threshold, then use full scan. Because, bitmap only supports equal query, it's very rare case to use full scan. IMHO, this threshold is applied for bitmap scanner as well. In other words, the result of statictics for bitmap scanner will be either using index or skipping index in most cases.

yueawang:
@lee-lei , Thanks! let's keep this 0.2 at 0.3 release, and maybe adjust it according to benchmark in future.

lee-lei:
Yes. Sure. Thanks.

carsonwang:
@adrian-wang @yueawang  Any further comments on this? We need get this in to start the benchmark.

adrian-wang:
LGTM except statistics data is still loaded separately from footer.

lee-lei:
Actually, in btree, although the stats content are put in the footer segment, it still loads twice during readFooter of BTreeIndexFileReader. The first is to get the footer length. The second is to load the footer itself. IMHO, this is reasonable, because we don't know the exact size of stats content beforehand. More importantly, once load or twice load are not the issue, because they just happened in the first time. After the first time, they are residing in the cache all the time. For the long-running applications which is just targeted by OAP, the initial load can be negligible from my point of view.

adrian-wang:
OK, I got wrong memory, since I implemented one-time reading in trie index...
As long as the index file format can be consistent for further potential improvement, this PR looks good to me.

Besides, @lee-lei Can you avoid using forced update when addressing comments? This is friendly for reviewer, since I have to re-read the whole changes every time you update.

lee-lei:
@adrian-wang Thanks for your suggestion. I will follow it next time. Thanks.

adrian-wang:

PR #542, opened 2018-01-24
Update README for memory configuration.

yueawang:
Update README

## How was this patch tested?
No need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,18 +13,17 @@ You should have [Apache Spark](http://spark.apache.org/) of version 2.1.0 instal
 2. Deploy `oap-<version>.jar` to master machine.
 3. Put below configurations to _$SPARK_HOME/conf/spark-defaults.conf_
 ```
-spark.files                      file:///path/to/oap-dir/oap-<version>.jar
-spark.executor.extraClassPath      ./oap-<version>.jar
-spark.driver.extraClassPath        /path/to/oap-dir/oap-<version>.jar
-spark.memory.offHeap.enabled       true
-spark.memory.offHeap.size          2g
-spark.driver.memory                2g
-spark.yarn.executor.memoryOverhead 2g
+spark.files                         file:///path/to/oap-dir/oap-<version>.jar
+spark.executor.extraClassPath       ./oap-<version>.jar
+spark.driver.extraClassPath         /path/to/oap-dir/oap-<version>.jar
+spark.memory.offHeap.enabled        true
+spark.memory.offHeap.size           20g
 ```
 4. Run spark by `bin/spark-sql`, `bin/spark-shell`, `sbin/start-thriftserver` or `bin/pyspark` and try our examples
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,10 +18,13 @@ spark.executor.extraClassPath      ./oap-<version>.jar
 spark.driver.extraClassPath        /path/to/oap-dir/oap-<version>.jar
 spark.memory.offHeap.enabled       true
 spark.memory.offHeap.size          2g
+spark.driver.memory                2g
+spark.yarn.executor.memoryOverhead 2g
```
adrian-wang: Since this is yarn only, we'd better not put it here. I think we can write this part and the following part in 2 sections, one for yarn and one for standalone, maybe another for mesos?

README.md
```diff
@@ -66,9 +69,9 @@ Parquet Support - Enable OAP support for parquet files
 * Default: true
 * Usage: `sqlContext.conf.setConfString(SQLConf.OAP_PARQUET_ENABLED.key, "false")`
 
-Fiber Cache Size - Total Memory size to cache Fiber. Unit: KB
-* Default: `spark.executor.memory * 0.3`
-* Usage: `sqlContext.conf.setConfString(SQLConf.OAP_FIBERCACHE_SIZE.key, s"{100 * 1024 * 1024}")`
+Fiber Cache Size - Total Memory size to cache Fiber, configured by 'spark.memory.offHeap.size'
+* Default: `spark.memory.offHeap.size * 0.35`
+* Usage: Fiber cache locates in off heap storage memory, basically this size is about spark.memory.offHeap.size * spark.memory.storageFraction (default 0.5) * 0.7
```
gczsjdy: - How to configure a proper `spark.memory.offHeap.size`? The user guide says `the larger the better`, it also says `spark.yarn.executor.memoryOverhead` should be around half of executor memory...
-  How to configure a proper cache size? Seems the user guide doesn't mention

yueawang: For 1. I think around half of the executor memory and as large as possible does not conflict :-), just set it as large as possible but no larger then 1/2 executor memory, as you need the other half to be the heap.
For 2. So far in 0.3.0, the cache size is calculated by spark.memory.offHeap.size * 0.35, you can adjust heap size and storageFraction to make a different cache, but you can not config it directly.

adrian-wang:
So driver memory and memory overhead for yarn has been removed at all?

yueawang:
Mentioned in few words in note 2, but give a link to user guide.
Do you think we need to put the recommanded config in both readme and user guide?

README.md
linhong-intel: I'm not sure about the cache size:
spark.memory.offHeap.size * 0.7 or
spark.memory.offHeap.size * 0.5 * 0.7?

yueawang: Oh, I didn't count the borrowing, I will change it back to 0.7 with an explanation.

carsonwang:
LGTM. Merged to master and 0.3

Issue #541, opened 2018-01-23
OAP IndexCache Compression


songzhan01:
In Baidu spark production environment，memory usage is very high，but cpu usage is relatively free.  IndexCache compression may be a good choice to improve cache hit ratio in the same size of memory.  At the same time，there exist a major negative effect that decompression tasks a bit of extra time when IndexCache is accessed. so let's discuss this, in the following google-doc is the result of IndexCache compression tests.
https://docs.google.com/document/d/1ZGbhAV7g2RssJ6V5GvbdipEFIyJ3-z2bhTigeefQtKU/edit?usp=sharing

linhong-intel:
Thanks @songzhan01 
2 questions:
1. during my test, uncompressed file usually is 2x larger than gzip compressed file. I'm not sure if the compressed fiber can also achieve this compression rate.

2. Can you perform the same test on a single task please? Then compare the running time and cache usage.
I'd like to avoid the affect like: locality, slower task, etc.

Anyway, the test result shows this is a useful feature. Thanks.

songzhan01:
Run a spark job which handles one task with different compression algorithm.  The test results as following google-doc.
https://docs.google.com/document/d/1lZtB4mQioxq8PN0wwtbtu5fovKmEqTvfBIHpjvknRBI/edit?usp=sharing

songzhan01:
Here is test result of DataCache with different compression algorithm using OapFileFormat. The details is as follows. 
https://docs.google.com/document/d/1-rZIL0utXGpcjBHNQM2aOOc8CAf3DNoouiRKz7QVPsc/edit?usp=sharing

Issue #509, opened 2018-01-15
Polish and improve the index selection file policy 


lee-lei:
In our index scanners, the index selection file policy is to compare if the index file size is greater than the ratio(default is 0.7) of oap data file size. If greater, then use full scan. If not, use index. However, the oap data file size here is the compressed size rather than the raw uncompressed file size. This comparison is unfair. The "SNAPPY" and "GZIP" have the good compression ratio. Thus the compressed oap data file size is much smaller. With this policy, it will use full scan rather than index. This is unexpected. So far, this file policy is disabled by default for all the unit tests.

The improvement will get the fair comparison between uncompressed index file size and uncompressed oap data file size. And it will be done in the next release(0.4.0).

chenghao-intel:
I don't think this is valid, as we need to check the filtered record ratio, other than the index file size.

lee-lei:
I am not sure what do you mean by filtered record ratio. What I meant is the code in lines 121 - 126 of IndexScanner.scala. IMHO, if it's invalid, it should be removed.

chenghao-intel:
@yueawang @carsonwang  can you please check the index selection strategy for BitMap? Sorry per my understanding, the index selection will reply on the filtered out record count (ratio), not the index file size itself. Anything not follow our design?

yueawang:
There are 2 policies (file size + statistics) which are applied to query together. The file size policy plays a IO cost role, however, we should have better prediction of file io cost.
So far I think roaring bitmap file size is impossible ~= db file size, so this policy should not take effect here.

lee-lei:
The key issue is that in this file size policy, the oap data file size is compressed with gizp format. The index file size is not compressed. It may lead to incorrect result (expected to use index may cause unexpected full scan) to compare the uncompressed index file size with the compressed oap data file size. This will impact both btree and bitmap index, from my point of view.

lee-lei:
By default the gzip appends the uncompressed size in the tail 4 byte of the file. However, the snappy and lz4 doesn't have it. For the solution of this issue, I plan to append the raw uncompressed size in the tail 4 byte of oap data file size. This will keep backward compatibility. Meanwhile it's convenient and fast to get the uncompressed file size. This issue doesn't impact the functionality of OAP. The plan is to fix it in 0.4.0 release.

chenghao-intel:
I think we need a unified standard to judge the index size, not only for BitMap Index, but also should apply to B+Tree Index.

[...]

PR #543, opened 2018-01-24
[0.3.0 QA] Polish the file size policy in index scanner.

lee-lei:

1. The solution is intended to compare the index file size and data file
   size in a fair approach. In other words, compare the umcompressed size.
2. Currently index file size is uncompressed. For oap data file, append the
   total uncompressed size while keep backward compatibility.
3. For parquet file, it may not be appropriate to change the file format to
   add the uncompressed size. Thus use the conservative expected compression
   ratio of snappy just for 0.3.0 release. It needs to figure out an elegant approach
   to address this issue in the next release.
## How was this patch tested?

mvn clean test package

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
@@ -114,7 +114,24 @@ private[oap] abstract class IndexScanner(idxMeta: IndexMeta)
 
       // Policy 1: index file size < data file size.
       val indexFileSize = indexPath.getFileSystem(conf).getContentSummary(indexPath).getLength
-      val dataFileSize = dataPath.getFileSystem(conf).getContentSummary(dataPath).getLength
+      val dataFileSize = if (dataPath.getName.endsWith(OapFileFormat.OAP_DATA_EXTENSION)) {
+        val fs = dataPath.getFileSystem(conf)
+        val fin = fs.open(dataPath)
+        val oapTotalSize = fs.getContentSummary(dataPath).getLength
+        val oapUncompressedSizeOffset = oapTotalSize - 12
+        fin.seek(oapUncompressedSizeOffset)
+        val oapUncompressedSize = fin.readLong()
+        fin.close()
+        oapUncompressedSize
+      } else {
+        // TODO: 1. Parquet file is compressed with snappy by default. It needs to figure out an
+        //          elegant approach to get the total uncompressed file size.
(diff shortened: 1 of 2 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
```diff
@@ -114,7 +114,24 @@ private[oap] abstract class IndexScanner(idxMeta: IndexMeta)
...
+        fin.close()
+        oapUncompressedSize
+      } else {
+        // TODO: 1. Parquet file is compressed with snappy by default. It needs to figure out an
+        //          elegant approach to get the total uncompressed file size.
+        // 2. In the future, the index file may be compressed as well. At that time, this policy
+        //    will get polished or refined.
+        // 3. Below is the conservative expected compression ratio of snappy. It may need to adjust
+        //    after extensive testing for typical use cases in the next release.
+        dataPath.getFileSystem(conf).getContentSummary(dataPath).getLength * 3
```
linhong-intel: We skip index when "read index file need more time than read data file".

So, I think all your code is based on 1 assumption:
time of reading uncompressed data
is equal to
time of reading same but compressed date + time of decompressing data

Then, why we choose to compress it?

lee-lei: Actually I don't have this kind of assumption. IMHO, this cost policy is to tentatively avoid the case that index time is greater than full scan time, because the huge index size may increase the total time due to the cost of high disk IO. The advantage of compression is the reduction of both disk IO and memory consumption in the cost of additional computing to do the decompression. It's worthy doing that. The issue #541 is one example.  Actually, if most of the uncompressed data is remote, IMHO, the time of reading these data is greater than the time of reading compressed and then decompression. Thanks.

linhong-intel: then why you compare index file with compressed file's uncompressed size?

lee-lei: Actually I described the motivation in the code. The index file size is uncompressed. The oap data file is compressed. The default comparison may be misleading to the cases which will use full scan although it may benefit from index. That's unfair from my point of view. You can get more background from the discussion in the issue #509.

lee-lei:
Close this pull request. In the next release(0.4.0), we will figure out an optimal solution. Thanks.

PR #549, opened 2018-01-29
[BRANCH-0.3][WA]Fix cache issue: LRU blocks memory release.

yueawang:
A simple WA to fix the cache release block issue i.e. the 1st item which is still in-use in pending queue blocks the guardian thread free the rest which have no reference.

## How was this patch tested?
mvn passed.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -18,7 +18,7 @@
 package org.apache.spark.sql.execution.datasources.oap.filecache
 
 import java.util.concurrent.{ConcurrentHashMap, LinkedBlockingQueue, TimeUnit}
-import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicLong}
 import java.util.concurrent.locks.ReentrantReadWriteLock
 
 import com.google.common.cache._
@@ -46,7 +46,12 @@ private[filecache] class CacheGuardian(maxMemory: Long) extends Thread with Logg
 
   private val removalPendingQueue = new LinkedBlockingQueue[(Fiber, FiberCache)]()
 
-  def pendingSize: Int = removalPendingQueue.size()
+  // Tell if guardian thread is trying to remove one Fiber.
+  val bRemoving: AtomicBoolean = new AtomicBoolean(false)
+
+  def pendingSize: Int =
+    if (bRemoving.get()) removalPendingQueue.size() + 1
+    else removalPendingQueue.size()
 
   def addRemovalFiber(fiber: Fiber, fiberCache: FiberCache): Unit = {
     _pendingFiberSize.addAndGet(fiberCache.size())
@@ -61,19 +66,23 @@ private[filecache] class CacheGuardian(maxMemory: Long) extends Thread with Logg
     // Loop forever, TODO: provide a release function
     while (true) {
       val (fiber, fiberCache) = removalPendingQueue.take()
+      bRemoving.set(true)
       logDebug(s"Removing fiber: $fiber")
       // Block if fiber is in use.
-      while (!fiberCache.tryDispose(fiber, 3000)) {
+      if (!fiberCache.tryDispose(fiber, 3000)) {
         // Check memory usage every 3s while we are waiting fiber release.
         logDebug(s"Waiting fiber to be released timeout. Fiber: $fiber")
+        removalPendingQueue.offer((fiber, fiberCache))
         if (_pendingFiberSize.get() > maxMemory) {
           logWarning("Fibers pending on removal use too much memory, " +
             s"current: ${_pendingFiberSize.get()}, max: $maxMemory")
         }
+      } else {
+        // TODO: Make log more readable
+        _pendingFiberSize.addAndGet(-fiberCache.size())
+        logDebug(s"Fiber removed successfully. Fiber: $fiber")
       }
-      // TODO: Make log more readable
-      _pendingFiberSize.addAndGet(-fiberCache.size())
-      logDebug(s"Fiber removed successfully. Fiber: $fiber")
+      bRemoving.set(false)
     }
   }
 }
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -47,7 +47,7 @@ private[filecache] class CacheGuardian(maxMemory: Long) extends Thread with Logg
   private val removalPendingQueue = new LinkedBlockingQueue[(Fiber, FiberCache)]()
 
   // Tell if guardian thread is trying to remove one Fiber.
-  @volatile var bRemoving: Boolean = false
+  @volatile private var bRemoving: Boolean = false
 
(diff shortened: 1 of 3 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,7 +46,12 @@ private[filecache] class CacheGuardian(maxMemory: Long) extends Thread with Logg
 
   private val removalPendingQueue = new LinkedBlockingQueue[(Fiber, FiberCache)]()
 
-  def pendingSize: Int = removalPendingQueue.size()
+  // Tell if guardian thread is trying to remove one Fiber.
+  val bRemoving: AtomicBoolean = new AtomicBoolean(false)
```
linhong-intel: no need to use an Atomic.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,7 +46,12 @@ private[filecache] class CacheGuardian(maxMemory: Long) extends Thread with Logg
...
+
+  def pendingSize: Int =
+    if (bRemoving.get()) removalPendingQueue.size() + 1
+    else removalPendingQueue.size()
```
linhong-intel: you have to use
```
if {
} else {
}
```
or something like
`removalPendingQueue.size() + bRemoving.get().toInt`

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -70,7 +70,7 @@ trait FiberCache extends Logging {
           writeLock.unlock()
         }
       }
-      logWarning(s"Fiber Cache Dispose waiting detected for ${this}")
+      logDebug(s"Fiber Cache Dispose waiting detected for ${this}")
```
linhong-intel: let's keep using warning.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManagerSuite.scala
```diff
@@ -260,4 +260,35 @@ class FiberCacheManagerSuite extends SharedOapContext {
     Thread.sleep(500)
     assert(fiberCache.isDisposed)
   }
+
+  test("LRU blocks memory free") {
+    spark.sparkContext.setLogLevel("OFF")
+    val memorySizeInMB = (MemoryManager.cacheMemory / mbSize).toInt
+    val dataInUse = generateData(mbSize)
+    val fiberInUse =
+      TestFiber(() => MemoryManager.putToDataFiberCache(dataInUse), s"test fiber #1.0")
```
linhong-intel: as I can remember, there is a fiber named "test fiber #1.0" in previous test.
If that fiber is in cache, there will be something wrong. so, please change another name

linhong-intel:
good to me now. How about go through another round of test to validate this PR?

yueawang:
Yes, I am going to re-run the test, see if we are able to catch a different log.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -46,7 +46,14 @@ private[filecache] class CacheGuardian(maxMemory: Long) extends Thread with Logg
 
   private val removalPendingQueue = new LinkedBlockingQueue[(Fiber, FiberCache)]()
 
-  def pendingSize: Int = removalPendingQueue.size()
+  // Tell if guardian thread is trying to remove one Fiber.
+  @volatile var bRemoving: Boolean = false
```
chenghao-intel: `private` ?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -61,19 +68,23 @@ private[filecache] class CacheGuardian(maxMemory: Long) extends Thread with Logg
     // Loop forever, TODO: provide a release function
     while (true) {
       val (fiber, fiberCache) = removalPendingQueue.take()
+      bRemoving = true
       logDebug(s"Removing fiber: $fiber")
       // Block if fiber is in use.
-      while (!fiberCache.tryDispose(fiber, 3000)) {
+      if (!fiberCache.tryDispose(fiber, 3000)) {
```
chenghao-intel: maybe `3000` is too big, and we can change that into smaller value or make it configurable. this may not be done in this PR, we can do that for next PR.

chenghao-intel: Another optimization is we can using the multiple threads instead of a single thread for the waiting. what do you think? And also we can leave it for next PR, if we do see many failed `tryDispose`s.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -60,18 +60,23 @@ trait FiberCache extends Logging {
...
     // Give caller a chance to deal with the long wait case.
     while (System.currentTimeMillis() - startTime <= timeout) {
-      if (writeLock.tryLock(200, TimeUnit.MILLISECONDS)) {
-        try {
-          if (refCount == 0) {
-            realDispose(fiber)
-            return true
+      if (refCount != 0) {
+        // LRU access done, but fiber was not released.
+        Thread.sleep(200)
```
chenghao-intel: I would suggest to not use the `thread.sleep(200)`, as the tryLock(200) has the same meaning, but can wake up earlier if the lock freed in another thread. Probably we'd better to revert it back? Or I am missing the original purpose for this?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
yueawang: Yes, it should be configurable. I think 3000ms is ok, however, sometimes it becomes polling instead of waiting, see my comments below. 
For multi-thread free, I think it's hard to control the concurrency level, especially when we see a large pending list, Does it matter?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
yueawang: The reason for this is the tryLock does not always wait 200ms, It happens that the lock successes on a fiber whose refcount != 0, so the next 3000ms waiting becomes 3000ms polling. And as I saw lots of log during the test this days, I think make this thread wait under such specific situation is better than polling. It's not a critical change actually.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
chenghao-intel: let's keep this optimization for next release.

chenghao-intel:
I just talked to  @yueawang offline, and we can leave some of the further optimization in 0.4 release. I don' have any further comments on this.  Thank you @yueawang for the good catch for bug fixing.

adrian-wang:
This pr has been merge to master and branch 0.3

PR #558, opened 2018-02-01
[BRANCH-0.3] OAP strategies do not use IsNotNull any more.

yueawang:
OAP strategies do not use IsNotNull any more.

It happens rarely that IsNotNull meet index scanner's requirement -- index hit range <= 20%.
In most case, null value is far less than valid value, therefore, we leave IsNotNull to full scan.
As a result of above description, OAP strategies do not use IsNotNull for an empty filter.

## How was this patch tested?
MVN test passed.


adrian-wang:

PR #573, opened 2018-02-11
follow up #561 release footer cache in bitmap scanner

adrian-wang:

if will not use bitmap index, bitmap index footer cache should also be released.
Thanks @songzhan01 for finding out!

## How was this patch tested?

ut


carsonwang:
LGTM. Merged to master and 0.3

PR #596, opened 2018-02-26
Disable specific indices by names

gczsjdy:

This PR allows to disable indices by index names, mainly for test purpose.

## How was this patch tested?

Test case in IndexSelectionSuite.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
@@ -55,14 +55,9 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
 
   private def selectAvailableIndex(
       intervalMap: mutable.HashMap[String, ArrayBuffer[RangeInterval]],
-      conf: RuntimeConfig): Unit = {
+      indexDisableListStr: String): Unit = {
     logDebug("Selecting Available Index:")
-    val indexDisableList = if (conf == null) {
-      Seq("")
-    } else {
-      conf.get(OapConf.OAP_INDEX_DISABLE_LIST.key, OapConf.OAP_INDEX_DISABLE_LIST.defaultValue.get)
-        .split(",").map(_.trim).toSeq
-    }
+    val indexDisableList = indexDisableListStr.split(",").map(_.trim).toSeq
 
     meta.indexMetas.filterNot(meta => indexDisableList.contains(meta.name)).foreach {
       indexMeta => indexMeta.indexType match {
@@ -73,7 +68,7 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
           }
         case BTreeIndex(entries) =>
           var num = 0 // the number of matched column
-        var flag = 0
+          var flag = 0
           // flag (terminated indication):
           // 0 -> Equivalence column; 1 -> Range column; 2 -> Absent column
           for (entry <- entries if flag == 0) {
@@ -97,7 +92,7 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
             num += 1
           }
           if (num > 0) {
-            availableIndexes.append((num-1, indexMeta))
+            availableIndexes.append((num - 1, indexMeta))
           }
         case BitMapIndex(entries) =>
           for (entry <- entries) {
@@ -199,8 +194,8 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
       intervalMap: mutable.HashMap[String, ArrayBuffer[RangeInterval]],
       options: Map[String, String] = Map.empty,
       maxChooseSize: Int = 1,
-      conf: RuntimeConfig = null): Unit = {
-    selectAvailableIndex(intervalMap, conf)
+      indexDisableList: String = ""): Unit = {
+    selectAvailableIndex(intervalMap, indexDisableList)
     val availableIndexers = getAvailableIndexers(intervalMap.size, maxChooseSize)
 
     //    intervalArray.sortWith(compare)
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
@@ -17,10 +17,10 @@
 
 package org.apache.spark.sql.execution.datasources.oap.index
 
-import org.apache.hadoop.conf.Configuration
-
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
+
+import org.apache.hadoop.conf.Configuration
 
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.{CatalystTypeConverters, InternalRow}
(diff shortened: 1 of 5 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -51,11 +54,19 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
     scanners = null
   }
 
-  private def selectAvailableIndex(intervalMap: mutable.HashMap[String, ArrayBuffer[RangeInterval]])
+  private def selectAvailableIndex(
+      intervalMap: mutable.HashMap[String, ArrayBuffer[RangeInterval]],
+      conf: Option[Configuration])
```
carsonwang: Why not use `Configuration` directly instead of an Option? Can you also move the below line to this line?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -51,11 +54,19 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
...
   : Unit = {
     logDebug("Selecting Available Index:")
     var idx = 0
-    while (idx < meta.indexMetas.length) {
+    val indexDisableList = conf match {
+      case None => Seq("")
+      case Some(configuration) => configuration.get(
+        OapConf.OAP_INDEX_DISABLE_LIST.key, OapConf.OAP_INDEX_DISABLE_LIST.defaultValue.get)
+          .toLowerCase().split(",").toSeq
```
carsonwang: Better to call `.trim` at last.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexSelectionSuite.scala
```diff
@@ -249,4 +251,21 @@ class IndexSelectionSuite extends SharedOapContext with BeforeAndAfterEach{
...
+    sql("create oindex idxb on oap_test(b)")
+
+    val oapMeta = DataSourceMeta.initialize(path, configuration)
+    val ic = new IndexContext(oapMeta)
+    val filters: Array[Filter] = Array(EqualTo("a", 8), EqualTo("b", 9))
+    ScannerBuilder.build(filters, ic, Map.empty, 2)
+    assert(ic.getScanners.get.scanners.length == 2)
+    ic.clear()
+
+    val conf = new Configuration()
```
carsonwang: can you use the existing `configuration` variable?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
carsonwang: The index name is always in lower case even if user uses upper case?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
gczsjdy: The rootest caller `build()` has been called many times without the last  `conf` parameter in `IndexSelectionSuite`. Because `conf` is not needed. If use `Configuration` directly here, the `conf` should be `null` here.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
gczsjdy: Good advice. Lower case doesn't make sense. I previously used index species to filter.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexSelectionSuite.scala
gczsjdy: I thought about this. But if using `configuration`, we should unset it in the end, which makes it not better than current way?

chenghao-intel:
@gczsjdy instead of disable a specific index via configuration, probably a better way is to provide the API and even the DDL to disable the index for CURRENT SESSION.

chenghao-intel:
maybe we can do that in the following up PR. @gczsjdy

gczsjdy:
@chenghao-intel OK, I will do that.

carsonwang:
can you fix the conf issue we discussed offline?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -51,11 +53,15 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
...
   }
 
-  private def selectAvailableIndex(intervalMap: mutable.HashMap[String, ArrayBuffer[RangeInterval]])
-  : Unit = {
+  private def selectAvailableIndex(
+      intervalMap: mutable.HashMap[String, ArrayBuffer[RangeInterval]]): Unit = {
     logDebug("Selecting Available Index:")
     var idx = 0
-    while (idx < meta.indexMetas.length) {
+    val indexDisableList = SparkSession.builder().getOrCreate().conf.get(
```
yueawang: A trivial suggestion, use getActiveSession rather than getOrCreate as I thought you do not want to create a session here.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -51,11 +53,15 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
...
+        OapConf.OAP_INDEX_DISABLE_LIST.key, OapConf.OAP_INDEX_DISABLE_LIST.defaultValue.get)
+          .split(",").map(_.trim).toSeq
+
+    while (idx < meta.indexMetas.length && !indexDisableList.contains(meta.indexMetas(idx).name)) {
```
yueawang: Can we use filter like below??
meta.indexMetas.filter( !indexDisableList.contains(_.name))).foreach{ 
_.indexType match{}
}

gczsjdy: Good advice:) I use `filterNot` instead :P

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -51,20 +53,27 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
...
         case BTreeIndex(entries) if entries.length == 1 =>
           val attribute = meta.schema(entries(0).ordinal).name
           if (intervalMap.contains(attribute)) {
-            availableIndexes.append((0, meta.indexMetas(idx)) )
+            availableIndexes.append((0, indexMeta) )
           }
         case BTreeIndex(entries) =>
           var num = 0 // the number of matched column
-          var flag = 0
+        var flag = 0
```
carsonwang: Why change this? The original style looks correct.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -88,18 +97,17 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
             num += 1
           }
           if (num > 0) {
-            availableIndexes.append((num - 1, meta.indexMetas(idx)) )
+            availableIndexes.append((num-1, indexMeta))
```
carsonwang: Why change ``num - 1`` to ``num-1``? The original style looks correct.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
```diff
@@ -19,11 +19,10 @@ package org.apache.spark.sql.execution.datasources.oap.index
 
 import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
-
```
carsonwang: why remove this line?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexScanner.scala
```diff
@@ -19,11 +19,10 @@ package org.apache.spark.sql.execution.datasources.oap.index
...
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
-
```
carsonwang: why remove this line?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
```diff
@@ -51,20 +53,27 @@ private[oap] class IndexContext(meta: DataSourceMeta) extends Logging {
     scanners = null
   }
 
-  private def selectAvailableIndex(intervalMap: mutable.HashMap[String, ArrayBuffer[RangeInterval]])
-  : Unit = {
+  private def selectAvailableIndex(
+      intervalMap: mutable.HashMap[String, ArrayBuffer[RangeInterval]],
+      conf: RuntimeConfig): Unit = {
```
carsonwang: Maybe passing `indexDisableList` as the parameter makes more sense. It looks a little strange to have a null conf.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexContext.scala
gczsjdy: Sorry, that's a merge error.

carsonwang:
LGTM. Merged.

PR #606, opened 2018-03-01
Fix a OapStrategy Issue.

yueawang:
OapSortLimitStrategy should support deserialized logicalPlan.
Normally we call sql like sql("").collect which uses plain logicalPlan, so that OapSortLimitStrategy is applicable.
But sometimes it can be written as: sql("").foreach{...} which will refer to a deserialized logicalPlan as the output of each rdd should be deserialized. OapSortLimitStrategy should also work fine with this case.

## How was this patch tested?
mvn test passed.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
@@ -82,6 +82,14 @@ trait OapStrategies extends Logging {
         case _ =>
           Nil
       }
+      case logical.Limit(IntegerLiteral(limit), logical.Sort(order, true, child)) =>
+        val childPlan = calcChildPlan(child, limit, order)
+        TakeOrderedAndProjectExec(limit, order, child.output, childPlan) :: Nil
+      case logical.Limit(
+          IntegerLiteral(limit),
+          logical.Project(projectList, logical.Sort(order, true, child))) =>
+        val childPlan = calcChildPlan(child, limit, order)
+        TakeOrderedAndProjectExec(limit, order, projectList, childPlan) :: Nil
       case _ => Nil
     }
 
@@ -347,9 +355,10 @@ trait OapStrategies extends Logging {
 
           val outputAttributes = readDataColumns ++ partitionColumns
 
+          val oapRelation = _fsRelation.copy(options = oapOption)(_fsRelation.sparkSession)
           val scan =
             new FileSourceScanExec(
-              _fsRelation,
+              oapRelation,
               outputAttributes,
               outputSchema,
               partitionKeyFilters.toSeq,
--- a/src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
+++ b/src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
@@ -25,7 +25,7 @@ import org.scalatest.BeforeAndAfterEach
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.TableIdentifier
 import org.apache.spark.sql.catalyst.catalog.CatalogTableType
-import org.apache.spark.sql.catalyst.encoders.{ExpressionEncoder, RowEncoder}
+import org.apache.spark.sql.catalyst.encoders.{encoderFor, ExpressionEncoder, RowEncoder}
 import org.apache.spark.sql.catalyst.plans.logical.CatalystSerde
 import org.apache.spark.sql.execution.QueryExecution
 import org.apache.spark.sql.execution.datasources.oap.utils.OapUtils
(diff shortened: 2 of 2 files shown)


src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
```diff
@@ -135,6 +138,27 @@ class OapPlannerSuite
...
+
+    val sqlText = "SELECT b FROM oap_sort_opt_table WHERE a >= 1 AND a <= 10 ORDER BY a LIMIT 4"
+    val logicPlan = spark.sessionState.sqlParser.parsePlan(sqlText)
+    val qe = new QueryExecution(spark, logicPlan)
+    assert(qe.sparkPlan.toString().contains("OapOrderLimitFileScanExec"))
+
+    implicit val exprEnc: ExpressionEncoder[Row] = encoderFor(RowEncoder(qe.analyzed.schema))
+    val deserializedPlan = CatalystSerde.deserialize[Row](logicPlan)
+    val qe1 = new QueryExecution(spark, deserializedPlan)
+    assert(qe1.sparkPlan.toString().contains("OapOrderLimitFileScanExec"))
```
gczsjdy: Can we test using
```
val df = sql("xxx")
df.foreach{println}
assert(df.queryExecution.sparkPlan.toString().contains("xxx"))
```
?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
```diff
@@ -347,9 +355,10 @@ trait OapStrategies extends Logging {
 
           val outputAttributes = readDataColumns ++ partitionColumns
 
+          val oapRelation = _fsRelation.copy(options = oapOption)(_fsRelation.sparkSession)
```
gczsjdy: Orinal code doesn't have `oapOption` information, was it a bug here?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
yueawang: df.queryExecution.sparkPlan does not know you called .foreach, it will give its logicalPlan which is a construct param of queryExecution.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
yueawang: Yes, this is another bug.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
carsonwang: I think we can use `df.queryExecution.executedPlan` for that purpose. @yueawang can you take a look?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
```diff
@@ -135,6 +138,27 @@ class OapPlannerSuite
     sql("drop oindex index2 on oap_sort_opt_table")
   }
 
+  test("SortPushDown should consider deserialized plan") {
+    val data = (1 to 300).map{ i => (i, s"this is test $i")}
```
carsonwang: nit: add a space between `map` and `{`

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
```diff
@@ -23,9 +23,13 @@ import org.apache.hadoop.fs.Path
 import org.scalatest.BeforeAndAfterEach
 
 import org.apache.spark.sql._
-import org.apache.spark.sql.catalyst.TableIdentifier
 import org.apache.spark.sql.catalyst.catalog.CatalogTableType
+import org.apache.spark.sql.catalyst.encoders._
+import org.apache.spark.sql.catalyst.encoders.{ExpressionEncoder, RowEncoder}
```
carsonwang: nit: One of the above two lines can be removed?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
```diff
@@ -23,9 +23,13 @@ import org.apache.hadoop.fs.Path
...
+import org.apache.spark.sql.catalyst.plans.logical.CatalystSerde
+import org.apache.spark.sql.catalyst.TableIdentifier
```
carsonwang: nit: The original order of this line is correct.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
```diff
@@ -23,9 +23,13 @@ import org.apache.hadoop.fs.Path
...
 import org.apache.spark.sql.execution.datasources.oap.utils.OapUtils
+import org.apache.spark.sql.execution.QueryExecution
```
carsonwang: nit: The import order is not correct. Move this one line above.

carsonwang:
Thanks @yueawang . Merging this.

PR #607, opened 2018-03-01
Add DDL support for indices disabling

gczsjdy:

To enable disabling indices using DDL.

Note that only 1 index can be disabled at a time. For example, to disable `indexA`, `indexB`:
Two statements: `Disable oindex indexA on table` & `Disable oindex indexB on table` are needed.
Due to supporting of multiple indices disabling at a time will cause other queries' parse error.

## How was this patch tested?

Test case in `IndexSelectionSuite`.

Out of date: <del>Test is currently not added, due to this actually depend on PR#596.</del>

### DIFF
--- a/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
+++ b/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
@@ -157,6 +157,7 @@ statement
     | DISABLE SINDEX (IF EXISTS)? IDENTIFIER ON tableIdentifier
         partitionSpec?                                                 #oapDisableIndex
     | SHOW SINDEX (FROM | IN) tableIdentifier                          #oapShowIndex
+    | SHOW DISABLED SINDICES                                           #oapShowDisabledIndices
     | CHECK SINDEX ON tableIdentifier partitionSpec?                   #oapCheckIndex
     | unsupportedHiveNativeCommands .*?                                #failNativeCommand
     ;
@@ -720,7 +721,7 @@ nonReserved
     | UNBOUNDED | WHEN
     | DATABASE | SELECT | FROM | WHERE | HAVING | TO | TABLE | WITH | NOT | CURRENT_DATE | CURRENT_TIMESTAMP
     | CHECK
-    | DISABLE
+    | DISABLE | DISABLED
     ;
 
 SELECT: 'SELECT';
@@ -831,6 +832,7 @@ COMMIT: 'COMMIT';
 ROLLBACK: 'ROLLBACK';
 MACRO: 'MACRO';
 DISABLE: 'DISABLE';
+DISABLED: 'DISABLED';
 
 IF: 'IF';
 
@@ -951,6 +953,7 @@ CURRENT_TIMESTAMP: 'CURRENT_TIMESTAMP';
 
 CHECK: 'CHECK';
 SINDEX: 'SINDEX' | 'OINDEX';
+SINDICES: 'SINDICES' | 'OINDICES';
 BTREE: 'BTREE';
 BLOOM: 'BLOOM';
 BITMAP: 'BITMAP';
--- a/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
+++ b/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
@@ -156,6 +156,7 @@ statement
         partitionSpec?                                                 #oapDropIndex
     | DISABLE SINDEX (IF EXISTS)? IDENTIFIER ON tableIdentifier
         partitionSpec?                                                 #oapDisableIndex
+    | ENABLE SINDEX IDENTIFIER                                         #oapEnableIndex
     | SHOW SINDEX (FROM | IN) tableIdentifier                          #oapShowIndex
     | SHOW DISABLED SINDICES                                           #oapShowDisabledIndices
     | CHECK SINDEX ON tableIdentifier partitionSpec?                   #oapCheckIndex
@@ -721,7 +722,7 @@ nonReserved
     | UNBOUNDED | WHEN
     | DATABASE | SELECT | FROM | WHERE | HAVING | TO | TABLE | WITH | NOT | CURRENT_DATE | CURRENT_TIMESTAMP
     | CHECK
-    | DISABLE | DISABLED
+    | DISABLE | DISABLED | ENABLE
     ;
 
 SELECT: 'SELECT';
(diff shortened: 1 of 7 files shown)


carsonwang:
can you add a test using the DDL?

carsonwang:
@gczsjdy , we also want to ENABLE the index again after DISABLE. LIST all disabled index is also necessary.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -649,3 +649,54 @@ case class OapCheckIndexCommand(
 
   }
 }
+
+/**
+ * Disable specific indices
+ */
+case class OapDisableIndexCommand(
+    indexNames: String,
```
carsonwang: nit: indexName

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -649,3 +649,54 @@ case class OapCheckIndexCommand(
...
+          // TODO get `fs` outside of foreach() to boost
+          val fs = parent.getFileSystem(sparkSession.sparkContext.hadoopConfiguration)
+          if (fs.exists(new Path(parent, OapFileFormat.OAP_META_FILE))) {
+            val metaBuilder = new DataSourceMetaBuilder()
+            val m = OapUtils.getMeta(sparkSession.sparkContext.hadoopConfiguration, parent)
+            assert(m.nonEmpty)
+            val oldMeta = m.get
+            val existsIndexNameList = oldMeta.indexMetas.map(_.name)
+            val existsData = oldMeta.fileMetas
+            val indexNameList = indexNames.split(",").map(_.trim)
```
carsonwang: Please update this if we only support disable one index

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -649,3 +649,54 @@ case class OapCheckIndexCommand(
...
+            val notExistsIndexNameList = indexNameList.filterNot(existsIndexNameList.contains(_))
+            val notExistsIndicesString = notExistsIndexNameList.mkString(", ")
+            if (!notExistsIndexNameList.isEmpty) {
+              if (!allowNotExists) {
+                throw new AnalysisException(
+                  s"""Index $notExistsIndicesString does not exist on
+                     | ${identifier.getOrElse(parent)}""".stripMargin)
+              } else {
+                logWarning(s"drop non-exists index $notExistsIndicesString")
```
carsonwang: nit: drop -> Disable

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -649,3 +649,54 @@ case class OapCheckIndexCommand(
...
+    partitionSpec: Option[TablePartitionSpec]) extends RunnableCommand {
+
+  override def run(sparkSession: SparkSession): Seq[Row] = {
+    val qe = sparkSession.sessionState.executePlan(UnresolvedRelation(table))
+    qe.assertAnalyzed()
+    val relation = qe.optimizedPlan
+
+    relation match {
+      case LogicalRelation(HadoopFsRelation(fileCatalog, _, _, _, format, _), _, identifier)
+        if format.isInstanceOf[OapFileFormat] || format.isInstanceOf[ParquetFileFormat] =>
```
carsonwang: nit: need two more spaces

src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexSelectionSuite.scala
```diff
@@ -20,12 +20,11 @@ package org.apache.spark.sql.execution.datasources.oap.index
 import java.io.File
 
 import scala.collection.mutable.ArrayBuffer
-
 import org.apache.hadoop.fs.Path
 import org.scalatest.BeforeAndAfterEach
-
```
carsonwang: Why removing these empty lines?

src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
```diff
@@ -156,7 +156,9 @@ statement
         partitionSpec?                                                 #oapDropIndex
     | DISABLE SINDEX (IF EXISTS)? IDENTIFIER ON tableIdentifier
         partitionSpec?                                                 #oapDisableIndex
+    | ENABLE SINDEX IDENTIFIER                                         #oapEnableIndex
     | SHOW SINDEX (FROM | IN) tableIdentifier                          #oapShowIndex
+    | SHOW DISABLED SINDICES                                           #oapShowDisabledIndices
```
yueawang: Why not just add a column (enable) to show index?? refer to breakpoint in gdb.

|Num|Type          |    Disp|**Enb**|Address                       |What                                          |
|-----|------------|--------|--------|------------------------|----------------------------------|
|1     |breakpoint |  keep  | **y**   |0x000000000040077a| in main() at hello_world.cpp:6 |
|2     |breakpoint |  keep  |    n      | PENDING                   | main + 5                                   |

Another thing is will we need "SHOW ENABLED SINDICES"?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
gczsjdy: Sorry, why?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
gczsjdy: Done, : )

src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4
gczsjdy: Sounds good, this seems more elegant than the current implementation to add a new grammar.
If a column is added, `show enabled/disabled sindices` is not needed I think. 
@carsonwang What do you think?

carsonwang: I like this idea.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
carsonwang: It is part of the `case` statement, right?

carsonwang: All right, just ignore this.

carsonwang:
LGTM. Merging.

PR #611, opened 2018-03-06
Fix indentation and new lines in scala style

carsonwang:
Fix indentation and new lines in scala. Some are not must but for consistency. 

## How was this patch tested?
Existing tests.


linhong-intel:
thanks @carsonwang . 
Can you leave a comment for each kind of code style fix. Then we can use your comment as an instruction. It will be helpful.
For example:
only one empty line between import and class
4 indents for function parameters break.

PR #616, opened 2018-03-09
Fix oap aggregation strategy performance issue.

yueawang:

The aggregation optimization in oapStrategies requires
group-ed items, bitmap index does not meet such requirement.
Now, it makes oap aggregation use btree index only.
Also, it uses another dedicated flag to control oap strategy.

## How was this patch tested?
mvn passed.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
@@ -172,6 +172,7 @@ class OapAggregationIterator(
         result.copy()
       }
 
+      numOutputRows += 1
       // If this is the last record, update the task's peak memory usage.
       // So far it has no map/sorter/spill memory.
       if (!hasNext) {
@@ -183,7 +184,6 @@ class OapAggregationIterator(
         spillSize += 0L
         metrics.incPeakExecutionMemory(maxMemory)
       }
-      numOutputRows += 1
       res
     } else {
       // no more result
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
@@ -290,15 +290,12 @@ trait OapStrategies extends Logging {
       oapOption: Map[String, String],
       indexHint: Seq[Expression],
       indexRequirements: Seq[IndexType]): Option[SparkPlan] = {
-    // If executor index selection (EIS) is enabled, oapStrategies are disabled.
     val conf = SparkSession.getActiveSession.get.sessionState.conf
-    if (!conf.getConf(OapConf.OAP_ENABLE_OPTIMIZATION_STRATEGIES)) {
+    if (!conf.getConf(OapConf.OAP_ENABLE_OPTIMIZATION_STRATEGIES) ||
+        conf.getConf(OapConf.OAP_ENABLE_EXECUTOR_INDEX_SELECTION)) {
+      // If executor index selection (spark.sql.oap.oindex.eis.enabled) is enabled,
+      // oapStrategies does not work because exeutor may skip the index scan.
       return None
-    } else {
-      if (conf.getConf(OapConf.OAP_ENABLE_EXECUTOR_INDEX_SELECTION)) {
-        logWarning("Please disable OAP_ENABLE_EXECUTOR_INDEX_SELECTION to make oapStrategies work.")
-        return None
-      }
     }
 
     // Filters on this relation fall into four categories based
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
@@ -41,10 +41,11 @@ trait OapStrategies extends Logging {
   def oapStrategies: Seq[Strategy] = {
     // BtreeIndex applicable strategies.
     OapSortLimitStrategy ::
+    OapGroupAggregateStrategy ::
     // BitMapIndex applicable strategies.
     OapSemiJoinStrategy ::
     // No requirement.
-    OapGroupAggregateStrategy :: Nil
+    Nil
   }
 
   /**
@@ -250,13 +251,21 @@ trait OapStrategies extends Logging {
           file @ HadoopFsRelation(_, _, _, _, _ : OapFileFormat, _), _, table)) =>
         val filterAttributes = AttributeSet(ExpressionSet(filters))
         val groupingAttributes = AttributeSet(groupExpressions.map(_.toAttribute))
+        val indexRequirement = filters.map(_ => BTreeIndex())
 
         if (groupingAttributes.size == 1 && filterAttributes == groupingAttributes) {
           val oapOption = new CaseInsensitiveMap(file.options +
             (OapFileFormat.OAP_INDEX_GROUP_BY_OPTION_KEY -> "true"))
 
           createOapFileScanPlan(
-            projectList, filters, relation, file, table, oapOption, filters, Nil) match {
+            projectList,
+            filters,
+            relation,
+            file,
+            table,
+            oapOption,
+            filters,
+            indexRequirement) match {
             case Some(fastScan) => OapAggregationFileScanExec(aggExpressions, projectList, fastScan)
             case _ => PlanLater(child)
           }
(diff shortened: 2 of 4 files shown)


src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
```diff
@@ -82,7 +82,13 @@ class OapPlannerSuite
     spark.sqlContext.dropTempTable("oap_fix_length_schema_table")
   }
 
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    spark.conf.set(OapConf.OAP_ENABLE_OPTIMIZATION_STRATEGIES.key, "true")
```
gczsjdy: `spark.conf.set(OapConf.OAP_ENABLE_OPTIMIZATION_STRATEGIES.key, true)`?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
```diff
@@ -283,8 +292,13 @@ trait OapStrategies extends Logging {
...
     // If executor index selection (EIS) is enabled, oapStrategies are disabled.
     val conf = SparkSession.getActiveSession.get.sessionState.conf
-    if (conf.getConf(OapConf.OAP_ENABLE_EXECUTOR_INDEX_SELECTION)) {
+    if (!conf.getConf(OapConf.OAP_ENABLE_OPTIMIZATION_STRATEGIES)) {
       return None
+    } else {
+      if (conf.getConf(OapConf.OAP_ENABLE_EXECUTOR_INDEX_SELECTION)) {
+        logWarning("Please disable OAP_ENABLE_EXECUTOR_INDEX_SELECTION to make oapStrategies work.")
+        return None
+      }
```
gczsjdy: 1. 
```
if (!conf.getConf(...)) {
  return None
} 
if (conf.getConf(...)) {
  logWarning("...") // Add some comments here?
  return None
}
``` 
Will this be more clear?
2. It seems at most 1 of statistics and OapStrategies can be enabled? Could we explain the reason in readme and maybe the conf docs?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
```diff
@@ -250,13 +251,21 @@ trait OapStrategies extends Logging {
           file @ HadoopFsRelation(_, _, _, _, _ : OapFileFormat, _), _, table)) =>
         val filterAttributes = AttributeSet(ExpressionSet(filters))
         val groupingAttributes = AttributeSet(groupExpressions.map(_.toAttribute))
+        val indexRequirement = filters.map(_ => BTreeIndex())
```
gczsjdy: I saw `OapStrategies enabled` degrades some of the benchmark, is this the main reason?

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
```diff
@@ -172,6 +172,7 @@ class OapAggregationIterator(
         result.copy()
       }
 
+      numOutputRows += 1
```
gczsjdy: Sorry why?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
yueawang: Yes, this suite should be run when .OAP_ENABLE_OPTIMIZATION_STRATEGIES is enabled, any concern??

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
yueawang: for 1, it seems no big diff, any specific reason for that?
for 2, yes, we should add some description in readme or wiki, while in running, user can find the correct setting by reading log.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
yueawang: So far it does not impact any benchmark other than those in oapStrategy suite.

src/main/scala/org/apache/spark/sql/execution/aggregate/OapAggregationIterator.scala
yueawang: Trivial change, and it is just because I want to log the correct row count in the if block next to it. :)

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
gczsjdy: 1 is a :nit, an `else` makes it not clear.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
gczsjdy: the second parameter can be boolean true?

yueawang: oh, overload may look better, although they are same, I'll change.

src/main/scala/org/apache/spark/sql/internal/oap/OapConf.scala
```diff
@@ -147,6 +147,13 @@ object OapConf {
       .booleanConf
       .createWithDefault(true)
 
+  val OAP_ENABLE_OPTIMIZATION_STRATEGIES =
+    SQLConfigBuilder("spark.sql.oap.strategies.enabled")
+      .internal()
+      .doc("To indicate if enable/disable oap strategies")
+      .booleanConf
+      .createWithDefault(false)
```
adrian-wang: why default false?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
```diff
@@ -82,7 +82,13 @@ class OapPlannerSuite
     spark.sqlContext.dropTempTable("oap_fix_length_schema_table")
   }
 
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    spark.conf.set(OapConf.OAP_ENABLE_OPTIMIZATION_STRATEGIES.key, true)
+  }
+
   override def afterAll(): Unit = {
+    spark.conf.set(OapConf.OAP_ENABLE_OPTIMIZATION_STRATEGIES.key, false)
```
adrian-wang: better use original value before this test.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapStrategies.scala
adrian-wang: I'm not sure, but using `BTreeIndex()` here is strange. I think if we don't care the indexed column, we should modify other interfaces to use `BTreeIndexType` instead.

src/main/scala/org/apache/spark/sql/internal/oap/OapConf.scala
yueawang: By default index selection is on, so oapstrategy can not work. Here I think default true value can be misleading.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
yueawang: OK, I can use defaultValue, as this flag is controlled only by this suite

yueawang:
Yes, I know your point, I can refine it in next PR.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/OapPlannerSuite.scala
```diff
@@ -88,7 +88,8 @@ class OapPlannerSuite
   }
 
   override def afterAll(): Unit = {
-    spark.conf.set(OapConf.OAP_ENABLE_OPTIMIZATION_STRATEGIES.key, false)
+    spark.conf.set(OapConf.OAP_ENABLE_OPTIMIZATION_STRATEGIES.key,
```
adrian-wang: break after `(` is prior to break after `,`

adrian-wang:
LGTM, merging, thanks!

PR #622, opened 2018-03-13
[WIP] Index File Compression Implementation

linhong-intel:

Partially compress BTree index and Bitmap Index.

## How was this patch tested?

current tests pass

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
@@ -25,6 +25,7 @@ import scala.collection.JavaConverters._
 import com.google.common.collect.ArrayListMultimap
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.mapreduce.{RecordWriter, TaskAttemptContext}
+import org.apache.parquet.format.CompressionCodec
 
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions._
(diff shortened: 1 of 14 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BTreeIndexRecordWriter.scala
```diff
@@ -44,6 +42,8 @@ private[index] case class BTreeIndexRecordWriter(
   @transient private lazy val genericProjector = FromUnsafeProjection(keySchema)
   private lazy val nnkw = new NonNullKeyWriter(keySchema)
 
+  private val codecFactory = new CodecFactory(configuration)
```
adrian-wang: No specific to this change, but maybe we need to add some transient properly in our code

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexUtils.scala
```diff
@@ -193,4 +194,34 @@ private[oap] object IndexUtils {
       (m, found)
     }
   }
+  def compressIndexData(
+      configuration: Configuration,
+      codecFactory: CodecFactory,
+      bytes: Array[Byte]): Array[Byte] = {
+    val codec = CompressionCodec.valueOf(
+        configuration.get(OapFileFormat.COMPRESSION, OapFileFormat.DEFAULT_COMPRESSION))
```
adrian-wang: how about just pass the codec or even codec string to this function? That would make this util easier for unit test.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexUtils.scala
```diff
@@ -193,4 +195,33 @@ private[oap] object IndexUtils {
...
+      val decompressor = codecFactory.getDecompressor(c)
+      val length = Platform.getInt(bytes, Platform.BYTE_ARRAY_OFFSET + 9)
+      val dec = decompressor.decompress(bytes.slice(13, bytes.length), length)
+      dec
+    } else {
+      bytes
+    }
+  }
+
+  def hash(bytes: Array[Byte]): String = {
```
yueawang: Is this used for security check? Which function will call it??

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexUtils.scala
yueawang: Should we use a OapCodec inherits CompressionCodec? so we can add specific codec on our own.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexUtils.scala
linhong-intel: just for debug if I decompress correctly. I'll remove it.

linhong-intel:
@adrian-wang 

> Not specific to this change, but maybe we need to add some transient properly in our code

Done in this PR. I'll fix other `codecFactory` seperately

> how about just pass the codec or even codec string to this function? That would make this util easier for unit test.

Each open file should have separated `CodecFactory/Compressor/Decompressor`. Since they are not thread-safe

@yueawang 
> Should we use a OapCodec inherits CompressionCodec? so we can add specific codec on our own.

Good idea.  It's also requested by Baidu.

linhong-intel:
Dup to #709 
close this.

PR #623, opened 2018-03-14
New OAP RPC module

gczsjdy:

This is the first subtask of new OAP RPC. Aiming to 
- Setup new RPC module framework
- Enable sending `DummyMessage` between Driver and Executor(two directions)
- Enable keeping Executors' DummyStatus on Driver

Refer to the design doc at: [design](https://drive.google.com/open?id=16-KkeIE27gODVkoFyp2A7eNlyuCB8jH4) 

Note that commit before `Basic` can be ignored.

## How was this patch tested?

`org.apache.spark.rpc.TestOapRpc`. 

Due to this is an RPC module. Local unit test doesn't work.
You can test on your cluster through: `bin/spark-submit --master spark://localhost:7077 --class org.apache.spark.rpc.TestOapRpc --executor-cores 2 --executor-memory 5g  /path/to/your/OAP/target/oap-0.4.0-SNAPSHOT-tests.jar`


src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedSchedulerBackend.scala
```diff
@@ -0,0 +1,613 @@
...
+      case KillTask(taskId, executorId, interruptThread) =>
+        executorDataMap.get(executorId) match {
+          case Some(executorInfo) =>
+            executorInfo.executorEndpoint.send(KillTask(taskId, executorId, interruptThread))
+          case None =>
+            // Ignoring the task kill since the executor is not registered.
+            logWarning(s"Attempted to kill task $taskId for unknown executor $executorId.")
+        }
+
+      case message: OapMessage => OapRpcManagerMaster.handleOapMessage(message)
```
gczsjdy: For this class, only this line added.

src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedSchedulerBackend.scala
```diff
@@ -0,0 +1,613 @@
...
+  private val maxRegisteredWaitingTimeMs =
+    conf.getTimeAsMs("spark.scheduler.maxRegisteredResourcesWaitingTime", "30s")
+  private val createTime = System.currentTimeMillis()
+
+  // Accessing `executorDataMap` in `DriverEndpoint.receive/receiveAndReply` doesn't need any
+  // protection. But accessing `executorDataMap` out of `DriverEndpoint.receive/receiveAndReply`
+  // must be protected by `CoarseGrainedSchedulerBackend.this`. Besides, `executorDataMap` should
+  // only be modified in `DriverEndpoint.receive/receiveAndReply` with protection by
+  // `CoarseGrainedSchedulerBackend.this`.
+  private[spark] val executorDataMap = new HashMap[String, ExecutorData]
```
linhong-intel: Since we can get `executorDataMap` through reflection. So don't change the modifier.

src/main/scala/org/apache/spark/executor/CoarseGrainedExecutorBackend.scala
```diff
@@ -125,8 +124,16 @@ private[spark] class CoarseGrainedExecutorBackend(
...
       }.start()
 
-    case message: OapMessage =>
-      OapMessageUtils.handleOapMessage(message)
+    case message: OapMessage => OapRpcManagerSlave.handleOapMessage(message)
+  }
+
+  // Before sending RPC message to Driver using OapRpcManagerSlave, please register the
+  // DriverEndpoint first
+  private def registerDriverEndpointToRpcManager = {
```
linhong-intel: I didn't see where this function is used.
Can you check `RpcUtils.makeDriverRef`and `NettyRpcEnv.asyncSetupEndpointRefByURI(driver_url)` see if we can avoid using this function?

src/main/scala/org/apache/spark/rpc/OapMessage.scala
```diff
@@ -0,0 +1,32 @@
...
+
+private[spark] sealed trait OapMessage extends Serializable
+
+private[spark] sealed trait OapDummyMessage extends OapMessage
+
+private[spark] sealed trait OapCacheMessage extends OapMessage
+private[spark] sealed trait OapIndexMessage extends OapMessage
+private[spark] sealed trait OapMetricsMessage extends OapMessage
+
+private[spark] object OapMessages {
```
linhong-intel: We need some super class can indicate the message direction. i.e. `ExecutorToDriverMessage`, `DriverToExecutorMessage`

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -216,9 +217,11 @@ case class DropIndexCommand(
     qe.assertAnalyzed()
     val relation = qe.optimizedPlan
 
-    sparkSession.sparkContext.schedulerBackend match {
+    val scheduler = sparkSession.sparkContext.schedulerBackend
+    scheduler match {
       case scheduler: CoarseGrainedSchedulerBackend =>
-          OapMessageUtils.sendMessageToExecutors(scheduler, CacheDrop(indexName))
+          OapRpcManagerMaster.registerScheduler(scheduler)
```
linhong-intel: `register` usually is used during `initialization` and expected only be called once.
if every time you send a message need to do the `register`, then `sendOapMessage(scheduler, message)` is better.

src/test/scala/org/apache/spark/rpc/TestOapRpc.scala
```diff
@@ -0,0 +1,66 @@
...
+object TestOapRpc {
+
+  val spark = SparkSession.builder().master("spark://localhost:7077").getOrCreate()
+
+  def testSendMessageDriverToExecutor: Unit = {
+    OapRpcManagerMaster.registerScheduler(spark.sparkContext.schedulerBackend)
+    OapRpcManagerMaster.sendOapMessage(DummyMessage("He is a wanderer"))
+  }
+
+  def testSendMessageExecutorToDriver: Unit = {
```
linhong-intel: Have you tried the test on cluster mode? I don't think it will be work.
BTW, it better to describe how to use your test case. I will be helpful for others to understand you design.

src/main/scala/org/apache/spark/rpc/OapRpcManagerBase.scala
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.spark.rpc
+
+import org.apache.spark.internal.Logging
+
+private[spark] trait OapRpcManagerBase extends Logging {
+
+  def handleOapMessage(message: OapMessage): Unit
```
yueawang: Just named by handle & send like Endpoint??
def handle/send(message: OapMessage)

src/main/scala/org/apache/spark/rpc/OapRpcManagerMaster.scala
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  private def sendMessageToExecutors(message: OapMessage): Unit = {
+    _scheduler match {
+      case Some(scheduler: CoarseGrainedSchedulerBackend) =>
+        val executorDataMap = scheduler.executorDataMap
+        for ((_, executorData) <- executorDataMap) {
+          executorData.executorEndpoint.send(message)
+        }
+      case Some(_) => throw new IllegalArgumentException("Not CoarseGrainedSchedulerBackend")
```
yueawang: Does this mean local mode throws exception??

src/main/scala/org/apache/spark/rpc/OapRpcManagerMaster.scala
```diff
@@ -0,0 +1,78 @@
...
+      logWarning(s"Dummy message from $executorId received on Driver: $someContent")
+      statusKeeper.dummyStatusMap += executorId -> someContent
+  }
+
+  private def handleCacheMessage(message: OapCacheMessage): Unit = message match {
+    case _ => ;
+  }
+
+  override def handleOapMessage(message: OapMessage): Unit = message match {
+    case dummyMessage: OapDummyMessage => handleDummyMessage(dummyMessage)
```
yueawang: Nit: can message have its own handler? so this becomes message.handler.handle()

src/main/scala/org/apache/spark/rpc/OapRpcManagerMaster.scala
```diff
@@ -0,0 +1,78 @@
...
+    case cacheMessage: OapCacheMessage => handleCacheMessage(cacheMessage)
+  }
+
+  private def sendDummyMessage(message: OapDummyMessage): Unit = message match {
+    case dummyMessage: DummyMessage => sendMessageToExecutors(dummyMessage)
+  }
+
+  override def sendOapMessage(message: OapMessage): Unit = message match {
+    case dummyMessage: OapDummyMessage => sendDummyMessage(dummyMessage)
+    case cacheMessage: OapCacheMessage => handleCacheMessage(cacheMessage)
```
yueawang: Why does send call handleCacheMessage which is also called by handleOapMessage? And handle sounds like a function called when we receive something.

src/main/scala/org/apache/spark/rpc/OapRpcManagerSlave.scala
```diff
@@ -0,0 +1,54 @@
...
+package org.apache.spark.rpc
+
+import org.apache.spark.rpc.OapMessages._
+
+private[spark] object OapRpcManagerSlave extends OapRpcManagerBase {
+
+  private var _driverEndpoint: Option[RpcEndpointRef] = None
+
+  def registerDriverEndpoint(driverEndpoint: RpcEndpointRef): Unit = _driverEndpoint match {
+    case None => _driverEndpoint = Some(driverEndpoint)
```
yueawang: Add a warning for case Some?

src/main/scala/org/apache/spark/rpc/OapRpcManagerSlave.scala
```diff
@@ -0,0 +1,54 @@
...
+    case DummyMessage(someContent) => logWarning(
+      s"Dummy message received on Executor: $someContent")
+  }
+
+  private def handleCacheMessage(message: OapCacheMessage): Unit = message match {
+    case _ => ;
+  }
+
+  override def handleOapMessage(message: OapMessage): Unit = message match {
+    case dummyMessage: OapDummyMessage => handleDummyMessage(dummyMessage)
```
yueawang: Nit, just suggest considering the self handle-able message.

src/main/scala/org/apache/spark/rpc/OapRpcManagerBase.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  def sendOapMessage(message: OapMessage): Unit
```
yueawang: Send always successes? Should we consider the return value?

src/test/scala/org/apache/spark/rpc/TestOapRpc.scala
```diff
@@ -0,0 +1,66 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.spark.rpc
+
+import org.apache.spark.rpc.OapMessages.{DummyMessage, DummyMessageWithId}
+import org.apache.spark.scheduler.cluster.CoarseGrainedSchedulerBackend
+import org.apache.spark.sql.SparkSession
+
+object TestOapRpc {
```
yueawang: Is it a test suite?? Please follow name convention if possible.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
yueawang: +1, should have a better place to register scheduler, maybe right after createEnv?

src/main/scala/org/apache/spark/rpc/OapRpcManagerSlave.scala
```diff
@@ -0,0 +1,54 @@
...
+    case _ =>
+  }
+
+  private def handleDummyMessage(message: OapDummyMessage): Unit = message match {
+    case DummyMessage(someContent) => logWarning(
+      s"Dummy message received on Executor: $someContent")
+  }
+
+  private def handleCacheMessage(message: OapCacheMessage): Unit = message match {
+    case _ => ;
```
yueawang: No implementation for cache message? Or you have not implemented it yet?

src/main/scala/org/apache/spark/scheduler/cluster/CoarseGrainedSchedulerBackend.scala
gczsjdy: Since we have to add this class in OAP, changing this to `private[spark]` is more elegant than getting `executorDataMap` through reflection?

src/main/scala/org/apache/spark/executor/CoarseGrainedExecutorBackend.scala
gczsjdy: I originally keep this function for later use. 
Now I move `OapRpcManagerSlave.registerDriverEndpoint(ref)` this inside `onStart`, this is better.

src/main/scala/org/apache/spark/rpc/OapMessage.scala
gczsjdy: Good advice : )

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
gczsjdy: Good point! The registeration should be only once.
I have moved this to `DriverEndpoint`'s onStart, then this registration is done automatically.
Also similar for the executor side.

src/test/scala/org/apache/spark/rpc/TestOapRpc.scala
gczsjdy: Yes, I tested on a Standalone cluster. It works.
OK, I will update that.

src/test/scala/org/apache/spark/rpc/TestOapRpc.scala
gczsjdy: No, due to this should be tested on cluster.

src/main/scala/org/apache/spark/rpc/OapRpcManagerSlave.scala
gczsjdy: No.
Yes, this is the 1st subtask for dealing only with `DummyMessage`.

src/main/scala/org/apache/spark/rpc/OapRpcManagerMaster.scala
gczsjdy: Now the register function(with parameter `CoarsegrainedSchedulerBakcend`) guarantees that it won't be local mode. Because local mode needs not RPC.

src/main/scala/org/apache/spark/rpc/OapRpcManagerMaster.scala
gczsjdy: Sorry, typo here. :P

src/main/scala/org/apache/spark/rpc/OapRpcManagerBase.scala
gczsjdy: You are right. We are going to consider this in the following PR since this only deals with simple one-way send requirement.
And because `Driver2ExecutorMessage` and opposite `Executor2DriverMessage` should be different trait so this `Base` class has to be removed. So the trait design is not needed.

src/main/scala/org/apache/spark/rpc/OapRpcManagerMaster.scala
gczsjdy: Can you explain it in detail?

yueawang: Just a thought, I remember I have seen somewhere that message handles itself. Anyway, you don't need to change this.

src/test/scala/org/apache/spark/rpc/TestOapRpc.scala
```diff
@@ -0,0 +1,55 @@
...
+  def testStatusKeeper: Unit = {
+    val ans = OapRpcManagerMaster.statusKeeper.getDummyStatus()
+    assert(ans.size == 1)
+    ans.foreach { item =>
+      assert(idAndContent == item)
+    }
+    logWarning("Message check passed!")
+  }
+
+  def main(args: Array[String]): Unit = {
```
carsonwang: This is not a unit test. It seems there is no need include this. We can add some unit tests later to cover the test.

src/main/scala/org/apache/spark/rpc/OapRpcManagerSlave.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.spark.internal.Logging
+import org.apache.spark.rpc.OapMessages._
+
+private[spark] object OapRpcManagerSlave extends Logging {
+
+  private var _driverEndpoint: Option[RpcEndpointRef] = None
+
+  private[spark] def registerDriverEndpoint(driverEndpoint: RpcEndpointRef): Unit = {
+    if (_driverEndpoint.isEmpty) {
+      _driverEndpoint = Some(driverEndpoint)
```
carsonwang: why check empty? we should support re-register.

src/main/scala/org/apache/spark/rpc/OapRpcManagerMaster.scala
```diff
@@ -0,0 +1,67 @@
...
+
+private[spark] object OapRpcManagerMaster extends Logging {
+
+  private var _scheduler: Option[CoarseGrainedSchedulerBackend] = None
+
+  private[rpc] val statusKeeper = RpcRelatedStatusKeeper
+
+  private[spark] def registerScheduler(schedulerBackend: CoarseGrainedSchedulerBackend): Unit = {
+    if (_scheduler.isEmpty) {
+      _scheduler = Some(schedulerBackend)
```
carsonwang: why check empty? we should support re-register.

gczsjdy:
@chenghao-intel @linhong-intel New implementation avoiding changes in `CoarseGrainedSchedulerBackend` & `CoarseGrainedExecutorBackend`, while make changes in `SparkEnv`, probably some Spark code changes can be moved to this class in the future.

src/main/scala/org/apache/spark/rpc/OapMessages.scala
```diff
@@ -0,0 +1,42 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.rpc
+
+private[spark] object OapMessages {
```
chenghao-intel: In order to follow the same pattern of Spark, i don't see Spark itself will divide the message with different trait says "ToRPCMaster" or "ToRPCSlave", and i think we should keep our code the simple the better.

chenghao-intel: And put the `OapMessage` into the `OapRpcManager`, so people will not change 2 files if they want to add a new message.

src/main/scala/org/apache/spark/rpc/OapRpcManagerSlave.scala
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.spark.internal.Logging
+import org.apache.spark.rpc.OapMessages._
+import org.apache.spark.sql.execution.datasources.oap.filecache.FiberCacheManager
+import org.apache.spark.sql.internal.oap.OapConf
+import org.apache.spark.util.{ThreadUtils, Utils}
+
+/**
+ * Similar OapRpcManager class with [[OapRpcManagerMaster]], however running on Executor
+ */
+private[spark] class OapRpcManagerSlave(
```
chenghao-intel: the package should be `oap` instead? can we?

src/main/scala/org/apache/spark/rpc/OapMessages.scala
gczsjdy: Please see `BlockManagerMessages`, `ToMaster` & `ToSlave` are seperated, this makes it more clear I think. 

In Spark, `BlockManagerMessages` are put in different files with `BlockManagerMaster/Slave` class. While `MapOutputTrackerMessages` are  put in the same. I think the former one is better, because it's cheap to switch between files, but it's worthy to be more clear.

src/main/scala/org/apache/spark/rpc/OapRpcManagerMaster.scala
```diff
@@ -0,0 +1,79 @@
...
+    true
+  }
+
+  private def handleDummyMessage(dummyMessage: MyDummyMessage) = dummyMessage match {
+    case MyDummyMessage(id, someContent) =>
+      logWarning(s"Dummy message received on Driver with id: $id, content: $someContent")
+  }
+
+  private def handleHeartBeat(heartBeat: HeartBeat) = heartBeat match {
+    case CacheMetrics(some: Int) => logWarning(s"Cache metrics received $some")
```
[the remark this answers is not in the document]
gczsjdy: This is for using method hints, I will remove this after the solution in this PR is understood and accepted, of course, before it's merged.

src/main/scala/org/apache/spark/rpc/OapRpcManagerSlave.scala
gczsjdy: Same opinion here, but `org.apache.spark.sql.execution.datasources.oap` has already existed, is it good if I create a `org.apache.spark.sql.oap.rpc` package?

chenghao-intel:
BTW, `SparkEnv.scala` seems no change at all, but also committed, probably you need to revert this. Still i'd like to see how the code is called from the Spark initialization from both master & slave side, sorry if i missed something.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -216,9 +217,10 @@ case class DropIndexCommand(
     qe.assertAnalyzed()
     val relation = qe.optimizedPlan
 
-    sparkSession.sparkContext.schedulerBackend match {
+    val scheduler = sparkSession.sparkContext.schedulerBackend
+    scheduler match {
       case scheduler: CoarseGrainedSchedulerBackend =>
-          OapMessageUtils.sendMessageToExecutors(scheduler, CacheDrop(indexName))
+        SparkEnv.get.oapRpcManager.send(CacheDrop(indexName))
```
chenghao-intel: will that work? I didn't see any code change from `SparkEnv`?

chenghao-intel: ok, most like the `SparkEnv` changes has been merged into master already, never mind.

src/main/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSlave.scala
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  override private[spark] def send(message: OapMessage): Unit = driverEndpoint.send(message)
+
+  private def reportHeartBeat(getMaterials: Seq[() => HeartBeat]): Unit = {
+    val materials = getMaterials.map(_.apply())
+    materials.foreach(send)
+  }
+
+  private[spark] def sendHearBeat(getMaterials: Seq[() => HeartBeat]): Unit = {
```
linhong-intel: Not a `startHeartBeat` during initialize?
I thought heart beat should be a periodic message and only need be started and leave it alone after that.

src/main/scala/org/apache/spark/SparkEnv.scala
```diff
@@ -0,0 +1,465 @@
...
+      serializerManager,
+      mapOutputTracker,
+      shuffleManager,
+      broadcastManager,
+      blockManager,
+      securityManager,
+      metricsSystem,
+      memoryManager,
+      outputCommitCoordinator,
+      oapRpcManager,
```
linhong-intel: we can change this to `oapManager` and let `oapRpcManager` be part of it if necessary in the future. right?

src/main/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSlave.scala
```diff
@@ -0,0 +1,85 @@
...
+    val MS = 1000
+    val intervalMs = MS * conf.getInt(
+          OapConf.OAP_HEARTBEAT_INTERVAL.key, OapConf.OAP_HEARTBEAT_INTERVAL.defaultValue.get)
```
carsonwang: nit: indent

src/main/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSlave.scala
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.spark.sql.execution.datasources.oap.filecache.FiberCacheManager
+import org.apache.spark.sql.internal.oap.OapConf
+import org.apache.spark.sql.oap.rpc.OapMessages._
+import org.apache.spark.util.{ThreadUtils, Utils}
+
+/**
+ * Similar OapRpcManager class with [[OapRpcManagerMaster]], however running on Executor
+ */
+private[spark] class OapRpcManagerSlave(
+  rpcEnv: RpcEnv, val driverEndpoint: RpcEndpointRef, executorId: String, conf: SparkConf)
```
carsonwang: nit: indent

src/main/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSlave.scala
```diff
@@ -0,0 +1,85 @@
...
+    extends OapRpcManager {
```
carsonwang: nit: indent

src/main/scala/org/apache/spark/sql/internal/oap/OapConf.scala
```diff
@@ -183,6 +183,13 @@ object OapConf {
     .stringConf
     .createWithDefault("")
 
+  val OAP_HEARTBEAT_INTERVAL =
+    SQLConfigBuilder("spark.sql.oap.rpc.heartbeat.interval")
+    .internal()
+    .doc("To Configure the OAP status update interval, for example OAP metrics")
+    .intConf
+    .createWithDefault(2)
```
carsonwang: To be consistent to "spark.executor.heartbeatInterval" (default 10s), we can make it a string conf and use `conf.getTimeAsMs` to get the value.

src/main/scala/org/apache/spark/SparkEnv.scala
gczsjdy: Yes. : )

src/main/scala/org/apache/spark/SparkEnv.scala
```diff
@@ -0,0 +1,465 @@
...
+
+    val oapRpcManagerMasterEndpoint = new OapRpcManagerMasterEndpoint(rpcEnv)
+    val oapRpcDriverEndpoint = registerOrLookupEndpoint(
+      OapRpcManagerMaster.DRIVER_ENDPOINT_NAME, oapRpcManagerMasterEndpoint)
+
+    val oapRpcManager = if (isDriver) {
+      new OapRpcManagerMaster(oapRpcManagerMasterEndpoint)
+    } else {
+      new OapRpcManagerSlave(rpcEnv, oapRpcDriverEndpoint, executorId, conf)
+    }
```
gczsjdy: @chenghao-intel These are lines are added. And the initialization are also here and inside(for example OapRpcManagerSlave's initialize)

src/main/scala/org/apache/spark/sql/oap/rpc/OapMessages.scala
```diff
@@ -15,10 +15,27 @@
...
-private[spark] sealed trait OapMessage extends Serializable
+import org.apache.spark.rpc.RpcEndpointRef
 
 private[spark] object OapMessages {
-  case class CacheDrop(indexName: String) extends OapMessage
+
+  sealed trait OapMessage extends Serializable
+
+  sealed trait ToOapRpcManagerSlave extends OapMessage
+  sealed trait ToOapRpcManagerMaster extends OapMessage
```
chenghao-intel: As i didn't see the same definition in Spark, probably we'd better follow the same approach. what do you think? otherwise, the code changes should happen in Spark upstream also.

src/main/scala/org/apache/spark/sql/oap/rpc/OapRpcManager.scala
```diff
@@ -0,0 +1,38 @@
...
+ * A base trait of OapRpcManagerMaster/Slave, use this class running on Driver/Executors to send
+ * messages to Executors/Driver.
+ * Note that not all functions need to be implemented in OapRpcManagerMaster/Slave
+ */
+
+trait OapRpcManager extends Logging {
+
+  private[spark] def send(message: OapMessage): Unit
+
+  // Different types of sending corresponding to interfaces in RpcEndpointRef, currently not used
```
chenghao-intel: Remove the used code and comments.

src/main/scala/org/apache/spark/sql/oap/rpc/OapMessages.scala
```diff
@@ -15,10 +15,27 @@
  * limitations under the License.
  */
 
-package org.apache.spark.sql.execution.datasources.oap
+package org.apache.spark.sql.oap.rpc
 
-private[spark] sealed trait OapMessage extends Serializable
+import org.apache.spark.rpc.RpcEndpointRef
 
 private[spark] object OapMessages {
```
chenghao-intel: can we use the package as '[oap]`?

src/main/scala/org/apache/spark/rpc/OapRpcManagerSlave.scala
chenghao-intel: why not just `o.a.s.s.oap.rpc`?

src/main/scala/org/apache/spark/rpc/OapMessages.scala
chenghao-intel: I mean we should follow the same pattern with Spark. and `ToMaster` or `ToSlave` will not give any constraints for the other developers even if they abuse it, otherwise, you should provide the explicit `assert` before we used it in the framework. what do you think? either way you can take.

src/main/scala/org/apache/spark/rpc/OapRpcManagerMaster.scala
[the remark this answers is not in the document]
chenghao-intel: As we talked offline, I think now it's the time, as we'd like to know how much effort we will take if we  used a separated endpoint for OAP service.

chenghao-intel:
Still don't commit the unchanged file `SparkEnv.scala`.

gczsjdy:
@chenghao-intel 
`SparkEnv.scala` has been changed, please see my at.

src/test/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSuite.scala
```diff
@@ -0,0 +1,122 @@
...
+    super.beforeEach()
+    val conf = new SparkConf()
+        .setMaster("local[2]")
+        .setAppName("test")
+        .set("spark.memory.offHeap.enabled", "true")
+        .set("spark.memory.offHeap.size", "1g")
+
+    sc = new SparkContext(conf)
+    rpcEnv = sc.env.rpcEnv
+    rpcManagerMasterEndpoint = spy(new OapRpcManagerMasterEndpoint(rpcEnv))
```
linhong-intel: Will you include this `spy` in your introduction? It seems interesting.

src/test/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSuite.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  test("Send non-heartbeat message from Executor to Driver") {
+    val rpcManagerSlave1 = addRpcManagerSlave(executorId1)
+
+    rpcManagerSlave1.send(message)
+    Thread.sleep(2000)
+    verify(rpcManagerMasterEndpoint).invokePrivate(_handleNormalOapMessage(message))
+  }
+
+  test("Send heartbeat message from Executor to Driver") {
```
linhong-intel: It's still not clear to me about heartbeat mechanism.
What's the plan of migrate spark's heartbeat to out own heartbeat?
I just want to see the real usage

src/test/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSuite.scala
```diff
@@ -0,0 +1,122 @@
...
+  private var rpcManagerMaster: OapRpcManagerMaster = _
+  private var rpcDriverEndpoint: RpcEndpointRef = _
+
+  // Helper private method accessors for OapRpcManagerMaster
+  private val _handleNormalOapMessage = PrivateMethod[Unit]('handleNormalOapMessage)
+  private val _handleHeartbeat = PrivateMethod[Unit]('handleHeartbeat)
+  // Helper private method accessors for OapRpcManagerSlave
+  private val _handleOapMessage = PrivateMethod[Unit]('handleOapMessage)
+
+  private val idAndContent = ("0", "He is a wanderer")
```
linhong-intel: should we assert this content in test case to ensure message is correct?

src/test/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSuite.scala
gczsjdy: Yep.

src/test/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSuite.scala
gczsjdy: Yes, actually the `verify(xxx)` asserts the certain function is called with certain parameters, which means the message will be checked.

gczsjdy:
@linhong-intel Mock is for unit test : )
Ok.

src/main/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSlave.scala
```diff
@@ -0,0 +1,103 @@
...
+  private val oapHeartbeatMaterials = mutable.HashSet.empty[() => Heartbeat]
+
+  private def getHeartbeatMaterials = oapHeartbeatMaterials
+
+  private val slaveEndpoint = rpcEnv.setupEndpoint(
+    s"OapRpcManagerSlave_$executorId", new OapRpcManagerSlaveEndpoint(rpcEnv))
+
+  initialize()
+
+  startOapHeartbeater()
```
linhong-intel: why don't you put this into `initialize`?

gczsjdy: Seems `startHeartbeater` doesn't belong to initialization in semantics?

gczsjdy:
@linhong-intel Thanks.

PR #628, opened 2018-03-21
[TBB] use openHashMap instead of multi hash map in index building

adrian-wang:

Use openHashMap since it should be faster.

## How was this patch tested?

ut passed, will benchmark this.


lee-lei:
It's great to know the benchmark result about how much reduction of the index building time. Thanks.

lee-lei:
IMHO, the role of the map(hashmap derivatives or other replacement) will span the whole lifecyle of the index writer. In other words, even if the replacement can spill to disk, somewhere we may still need to load back into memory from disk since we may use it later. That may cause the data moving back and forth between memory and disk in an extreme case. That may be another huge issue, compared with the existing GC issue in index building phase. One good approach is to do some performance comparison before deciding to use whatever approach. Thanks.

gczsjdy:
Sorry, but what is TBB?

adrian-wang:
to be benchmark

linhong-intel:
any update on this? Is benchmark still needed.

PR #633, opened 2018-03-23
Import class ApiRootResource

TigerSong:

* Import src/main/scala/org/apache/spark/status/api/v1/ApiRootResource.scala from `spark v2.1.0` tag,  which is the root resource class of RESTful api.
* To minimize modifications, only import class ApiRootResource.

## How was this patch tested?

no difference between the original class


carsonwang:
Thanks @TigerSong . Merging this.

PR #637, opened 2018-03-28
Bitmap utils to directly traverse row IDs in fiber caches

lee-lei:

This patch provides the lightweight functionality to directly
traverse the row IDs in fiber caches which are compatible with
the roaring bitmap format.
It doens't need to create roaring bitmap object on heap.

Meanwhile this patch provides two test cases to test the functionality
of traversing the row IDs in single fiber cache or multi fiber caches
for two typical equal query cases with bitmap index.
One is single equal value case. The other is multi equal value case.

1. select * from t where a = apple
2. select * from t where a in {"apple", "banana", ...}

## How was this patch tested?
mvn clean test package and all tests passed locally

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
@@ -20,46 +20,6 @@ package org.apache.spark.sql.execution.datasources.oap.utils
 import scala.collection.mutable.ArrayBuffer
 import scala.util.control.Breaks._
 
-/* The chunks inside of one single fiber cache are physically consecutive,
- * so it doesn't require to set chunk offset. This class is to directly get
- * the row ID list recorded in all of the chunks of this single fiber cache
- * in ascending order.
- */
-private[oap] case class ChunksInSingleFiberCacheIterator(wfc: OapBitmapWrappedFiberCache)
-  extends ChunksIterator {
-
-  override def init(): Iterator[Int] = {
-    totalLength = wfc.getTotalChunkLength
-    if (idx < totalLength) {
-      iteratorForChunk = wfc.getIteratorForChunk(idx)
-      val cks = wfc.getChunkKeys
-      highPart = (cks(idx) & 0xFFFF) << 16
-    }
-    return this
-  }
-}
-
-/* The chunks inside of different multi fiber cache are not physically consecutive, so it
- * requires to set chunk offset for different fibers. This class is to directly get the row ID
- * list recorded in all of the chunks in all of multi fiber caches in ascending order.
- */
-private[oap] case class ChunksInMultiFiberCachesIterator(
-    chunksInFc: ArrayBuffer[OapBitmapChunkInFiberCache])
-  extends ChunksIterator {
-
-  override def init(): Iterator[Int] = {
-    totalLength = chunksInFc.length
-    if (idx < totalLength) {
-      val wfc = chunksInFc(idx).wfc
-      val chunkIdx = chunksInFc(idx).chunkIdx
-      wfc.setOffset(chunkIdx)
-      iteratorForChunk = wfc.getIteratorForChunk(chunkIdx)
-      highPart = (chunksInFc(idx).getChunkKey & 0xFFFF) << 16
-    }
-    return this
-  }
-}
-
 private[oap] object BitmapUtils {
 
   // Below constants are used by OapBitmapWrappedFiberCache class.
@@ -69,10 +29,66 @@ private[oap] object BitmapUtils {
   val DEFAULT_MAX_SIZE: Int = 4096
   val BITMAP_MAX_CAPACITY: Int = 1 << 16
 
-  /* Below method will virtually link all the chunks in multi fiber caches in ascending order
+  def iterator(wfcSeq: Seq[OapBitmapWrappedFiberCache]): Iterator[Int] = {
+    // For only one fiber cache, use ChunksInSingleFiberCacheIterator for best performance.
+    if (wfcSeq.size == 1) {
+      val wfc = wfcSeq.head
+      wfc.init
+      ChunksInSingleFiberCacheIterator(wfc).prepare
+    } else {
+      ChunksInMultiFiberCachesIterator(wfcSeq).prepare
+    }
+  }
+}
+
+/**
+ * The chunks inside of one single fiber cache are physically consecutive,
+ * so it doesn't require to set chunk offset. This class is to directly get
+ * the row ID list recorded in all of the chunks of this single fiber cache
+ * in ascending order.
+ */
+private[oap] case class ChunksInSingleFiberCacheIterator(wfc: OapBitmapWrappedFiberCache)
+  extends ChunksIterator {
+
+  override def prepare(): Iterator[Int] = {
+    totalLength = wfc.getTotalChunkLength
+    if (idx < totalLength) {
+      iteratorForChunk = wfc.getIteratorForChunk(idx)
+      val cks = wfc.getChunkKeys
+      highPart = (cks(idx) & 0xFFFF) << 16
+    }
+    this
+  }
+}
+
+/**
+ * The chunks inside of different multi fiber cache are not physically consecutive, so it
+ * requires to set chunk offset for different fibers. This class is to directly get the row ID
+ * list recorded in all of the chunks in all of multi fiber caches in ascending order.
+ */
+private[oap] case class ChunksInMultiFiberCachesIterator(
+    wfcSeq: Seq[OapBitmapWrappedFiberCache])
+  extends ChunksIterator {
+
+  private val chunksInFc: ArrayBuffer[OapBitmapChunkInFiberCache] = or
+
+  override def prepare(): Iterator[Int] = {
+    totalLength = chunksInFc.length
+    if (idx < totalLength) {
+      val wfc = chunksInFc(idx).wfc
+      val chunkIdx = chunksInFc(idx).chunkIdx
+      wfc.setOffset(chunkIdx)
+      iteratorForChunk = wfc.getIteratorForChunk(chunkIdx)
+      highPart = (chunksInFc(idx).getChunkKey & 0xFFFF) << 16
+    }
+    this
+  }
+
+  /**
+   * Below method will virtually link all the chunks in multi fiber caches in ascending order
    * of chunk key. It will provide the input for the above ChunksInMultiFiberCachesIterator.
    */
-  def or(wfcSeq: Seq[OapBitmapWrappedFiberCache]): ArrayBuffer[OapBitmapChunkInFiberCache] = {
+  private def or(): ArrayBuffer[OapBitmapChunkInFiberCache] = {
     val firstWfc = wfcSeq(0)
     firstWfc.init
     val initialChunkLength = firstWfc.getTotalChunkLength
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
@@ -20,7 +20,11 @@ package org.apache.spark.sql.execution.datasources.oap.utils
 import scala.collection.mutable.ArrayBuffer
 import scala.util.control.Breaks._
 
-// The chunks are physically consecutive, so it doesn't require to set chunk offset.
+/* The chunks inside of one single fiber cache are physically consecutive,
+ * so it doesn't require to set chunk offset. This class is to directly get
+ * the row ID list recorded in all of the chunks of this single fiber cache
+ * in ascending order.
+ */
 private[oap] case class ChunksInSingleFiberCacheIterator(wfc: OapBitmapWrappedFiberCache)
   extends ChunksIterator {
 
@@ -35,7 +39,10 @@ private[oap] case class ChunksInSingleFiberCacheIterator(wfc: OapBitmapWrappedFi
   }
 }
 
-// The chunks are not physically consecutive, so it requires to set chunk offset.
+/* The chunks inside of different multi fiber cache are not physically consecutive, so it
+ * requires to set chunk offset for different fibers. This class is to directly get the row ID
+ * list recorded in all of the chunks in all of multi fiber caches in ascending order.
+ */
 private[oap] case class ChunksInMultiFiberCachesIterator(
     chunksInFc: ArrayBuffer[OapBitmapChunkInFiberCache])
   extends ChunksIterator {
@@ -53,8 +60,9 @@ private[oap] case class ChunksInMultiFiberCachesIterator(
   }
 }
 
-// This will virtually link all the chunks in multi fiber caches in ascending order of chunk key.
-// It will provide the input for the above ChunksInMultiFiberCachesIterator class.
+/* This method will virtually link all the chunks in multi fiber caches in ascending
+ * order of chunk key. It will provide the input for the above ChunksInMultiFiberCachesIterator.
+ */
 private[oap] object BitmapUtils {
 
   // Just get array of the chunks across multi fiber caches in acending order of key.
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
@@ -1,0 +1,113 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.execution.datasources.oap.utils
+
+import scala.collection.mutable.ArrayBuffer
+import scala.util.control.Breaks._
+
+// The chunks are physically consecutive, so it doesn't require to set chunk offset.
+private[oap] case class ChunksInSingleFiberCacheIterator(wfc: OapBitmapWrappedFiberCache)
+  extends ChunksIterator {
+
+  override def init(): Iterator[Int] = {
+    totalLength = wfc.getTotalChunkLength
+    if (idx < totalLength) {
+      iteratorForChunk = wfc.getIteratorForChunk(idx)
+      val cks = wfc.getChunkKeys
+      highPart = (cks(idx) & 0xFFFF) << 16
+    }
+    return this
+  }
+}
+
+// The chunks are not physically consecutive, so it requires to set chunk offset.
+private[oap] case class ChunksInMultiFiberCachesIterator(
+    chunksInFc: ArrayBuffer[OapBitmapChunkInFiberCache])
+  extends ChunksIterator {
+
+  override def init(): Iterator[Int] = {
+    totalLength = chunksInFc.length
+    if (idx < totalLength) {
+      val wfc = chunksInFc(idx).wfc
+      val chunkIdx = chunksInFc(idx).chunkIdx
+      wfc.setOffset(chunkIdx)
+      iteratorForChunk = wfc.getIteratorForChunk(chunkIdx)
+      highPart = (chunksInFc(idx).getChunkKey & 0xFFFF) << 16
+    }
+    return this
+  }
+}
+
+private[oap] object OapBitmapFastAggregation {
+
+  // Just get array of the chunks across multi fiber caches in acending order of key.
+  def or(wfcSeq: Seq[OapBitmapWrappedFiberCache]): ArrayBuffer[OapBitmapChunkInFiberCache] = {
+    val firstWfc = wfcSeq(0)
+    firstWfc.init
+    val initialChunkLength = firstWfc.getTotalChunkLength
+    val finalChunkArray = new ArrayBuffer[OapBitmapChunkInFiberCache]()
+    var initialIdx = 0
+    var nextIdx = 0
+    (0 until initialChunkLength).map(idx => {
+      finalChunkArray += OapBitmapChunkInFiberCache(firstWfc, idx)
+    })
+    (1 until wfcSeq.length).foreach(idx => {
+      initialIdx = 0
+      var initialKey = finalChunkArray(initialIdx).getChunkKey
+      val nextWfc = wfcSeq(idx)
+      nextWfc.init
+      nextIdx = 0
+      val nextChunkLength = nextWfc.getTotalChunkLength
+      val nextChunkKeys = nextWfc.getChunkKeys
+      var nextKey = nextChunkKeys(nextIdx)
+      breakable {
+        while (true) {
+          val result = initialKey - nextKey
+          result match {
+            case res if res < 0 =>
+              initialIdx += 1
+              if (initialIdx == finalChunkArray.length) break
+              initialKey = finalChunkArray(initialIdx).getChunkKey
+            case res if res == 0 =>
+              // Just link the next chunk to be adjacent for traversing.
+              finalChunkArray.insert(initialIdx, OapBitmapChunkInFiberCache(nextWfc, nextIdx))
+              // Bypass the two adjacent chunks with equal keys.
+              initialIdx += 2
+              nextIdx += 1
+              if (initialIdx == finalChunkArray.length || nextIdx == nextChunkLength) break
+              initialKey = finalChunkArray(initialIdx).getChunkKey
+              nextKey = nextChunkKeys(nextIdx)
+            case res if res > 0 =>
+              // Insert the next chunk with nextIdx from the next fiber cache.
+              finalChunkArray.insert(initialIdx, OapBitmapChunkInFiberCache(nextWfc, nextIdx))
+              initialIdx += 1
+              nextIdx += 1
+              if (nextIdx == nextChunkLength) break
+              nextKey = nextChunkKeys(nextIdx)
+          }
+        }
+      }
+      if (initialIdx == finalChunkArray.length && nextIdx < nextChunkLength) {
+        // Append the remaining chunks from the above next fiber cache.
+        (nextIdx until nextChunkLength).foreach(idx =>
+          finalChunkArray += OapBitmapChunkInFiberCache(nextWfc, idx))
+      }
+    })
+    finalChunkArray
+  }
+}
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
@@ -53,7 +53,9 @@ private[oap] case class ChunksInMultiFiberCachesIterator(
   }
 }
 
-private[oap] object OapBitmapFastAggregation {
+// This will virtually link all the chunks in multi fiber caches in ascending order of chunk key.
+// It will provide the input for the above ChunksInMultiFiberCachesIterator class.
+private[oap] object BitmapUtils {
 
   // Just get array of the chunks across multi fiber caches in acending order of key.
   def or(wfcSeq: Seq[OapBitmapWrappedFiberCache]): ArrayBuffer[OapBitmapChunkInFiberCache] = {
@@ -78,12 +80,11 @@ private[oap] object OapBitmapFastAggregation {
       breakable {
         while (true) {
           val result = initialKey - nextKey
-          result match {
-            case res if res < 0 =>
+            if (result < 0) {
               initialIdx += 1
               if (initialIdx == finalChunkArray.length) break
               initialKey = finalChunkArray(initialIdx).getChunkKey
-            case res if res == 0 =>
+            } else if (result == 0) {
               // Just link the next chunk to be adjacent for traversing.
               finalChunkArray.insert(initialIdx, OapBitmapChunkInFiberCache(nextWfc, nextIdx))
               // Bypass the two adjacent chunks with equal keys.
@@ -92,7 +93,7 @@ private[oap] object OapBitmapFastAggregation {
               if (initialIdx == finalChunkArray.length || nextIdx == nextChunkLength) break
               initialKey = finalChunkArray(initialIdx).getChunkKey
               nextKey = nextChunkKeys(nextIdx)
-            case res if res > 0 =>
+            } else if (result > 0) {
               // Insert the next chunk with nextIdx from the next fiber cache.
               finalChunkArray.insert(initialIdx, OapBitmapChunkInFiberCache(nextWfc, nextIdx))
               initialIdx += 1
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
@@ -60,12 +60,18 @@ private[oap] case class ChunksInMultiFiberCachesIterator(
   }
 }
 
-/* This method will virtually link all the chunks in multi fiber caches in ascending
- * order of chunk key. It will provide the input for the above ChunksInMultiFiberCachesIterator.
- */
 private[oap] object BitmapUtils {
 
-  // Just get array of the chunks across multi fiber caches in acending order of key.
+  // Below constants are used by OapBitmapWrappedFiberCache class.
+  val SERIAL_COOKIE: Int = 12347
+  val SERIAL_COOKIE_NO_RUNCONTAINER: Int = 12346
+  val NO_OFFSET_THRESHOLD: Int = 4
+  val DEFAULT_MAX_SIZE: Int = 4096
+  val BITMAP_MAX_CAPACITY: Int = 1 << 16
+
+  /* Below method will virtually link all the chunks in multi fiber caches in ascending order
+   * of chunk key. It will provide the input for the above ChunksInMultiFiberCachesIterator.
+   */
   def or(wfcSeq: Seq[OapBitmapWrappedFiberCache]): ArrayBuffer[OapBitmapChunkInFiberCache] = {
     val firstWfc = wfcSeq(0)
     firstWfc.init
@@ -108,7 +114,7 @@ private[oap] object BitmapUtils {
               nextIdx += 1
               if (nextIdx == nextChunkLength) break
               nextKey = nextChunkKeys(nextIdx)
-          }
+            }
         }
       }
       if (initialIdx == finalChunkArray.length && nextIdx < nextChunkLength) {
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCache.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCache.scala
@@ -28,12 +28,6 @@ import org.apache.spark.sql.execution.datasources.oap.filecache.{FiberCache, Wra
 // The spec link is https://github.com/RoaringBitmap/RoaringFormatSpec/
 private[oap] class OapBitmapWrappedFiberCache(fc: FiberCache) extends WrappedFiberCache(fc) {
 
-  private val SERIAL_COOKIE: Int = 12347
-  private val SERIAL_COOKIE_NO_RUNCONTAINER: Int = 12346
-  private val NO_OFFSET_THRESHOLD: Int = 4
-  private val DEFAULT_MAX_SIZE: Int = 4096
-  private val BITMAP_MAX_CAPACITY: Int = 1 << 16
-
   private var chunkLength: Int = 0
   // It indicates no this section.
   private var chunkOffsetListOffset: Int = -1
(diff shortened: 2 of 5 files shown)


src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapMicroBenchmarkSuite.scala
```diff
@@ -267,4 +287,133 @@ class BitmapMicroBenchmarkSuite extends QueryTest with SharedOapContext with Bef
...
+            wfc.release
+          })
+          // The row id is starting from 0.
+          maxRowId += maxRowIdInPartition + 1
+          bmFooterCache.release
+          bmOffsetListCache.release
+          fin.close
+        }
+      })
+      assert(sql("select * from t").count == maxRowId)
```
gczsjdy: I think comparing the count is not enough, the result set has to be exactly the same.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapMicroBenchmarkSuite.scala
```diff
@@ -20,12 +20,18 @@ package org.apache.spark.sql.execution.datasources.oap.index
 import java.io._
```
gczsjdy: Actually not benchmark suite.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/OapBitmapWrappedFiberCache.scala
```diff
@@ -0,0 +1,282 @@
...
+
+import org.apache.spark.sql.execution.datasources.OapException
+import org.apache.spark.sql.execution.datasources.oap.filecache.{FiberCache, WrappedFiberCache}
+
+// Oap Bitmap Scanner will use below class to directly traverse the row IDs
+// in fiber caches while keeping compatible with Roaring Bitmap format.
+// It doesn't need to create the roaring bitmap object on heap any more.
+// The spec link is https://github.com/RoaringBitmap/RoaringFormatSpec/
+private[oap] class OapBitmapWrappedFiberCache(
+    fc: FiberCache) extends WrappedFiberCache(fc) {
```
gczsjdy: Can we extract a class named `OapBitmap`? It doesn't contain `FiberCache` related logic, for example it can hold an argument of offheap address.

Then, the correctness test ensures this whole class work right. I think this is a better way.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/OapBitmapWrappedFiberCache.scala
```diff
@@ -0,0 +1,282 @@
...
+      (fc.getByte(curPos) & 0xFFL)
+  }
+
+  def getTotalChunkLength(): Int = chunkLength
+
+  def getChunkKeys(): Array[Short] = chunkKeys
+
+  def getChunkCardinality(idx: Int): Short = chunkCardinalities(idx)
+
+  def getBitmapCapacity(): Int = BITMAP_MAX_CAPACITY
```
gczsjdy: Are there any functions can be made private?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapMicroBenchmarkSuite.scala
```diff
@@ -267,4 +287,128 @@ class BitmapMicroBenchmarkSuite extends QueryTest with SharedOapContext with Bef
...
+    val bmOffsetListOffset = bmEntryListOffset + bmEntryListTotalSize + bmNullEntrySize
+    val bmOffsetListFiber = BitmapFiber(
+      () => loadBmOffsetList(fin, bmOffsetListOffset, bmOffsetListTotalSize),
+      idxPath.toString, BitmapIndexSectionId.entryOffsetsSection, 0)
+    val bmOffsetListCache = WrappedFiberCache(FiberCacheManager.get(bmOffsetListFiber, conf))
+    (keyCount, bmOffsetListCache, bmFooterCache)
+  }
+
+  test("test the functionality of directly traversing single fiber cache without roaring bitmap") {
+    dataArray.foreach(dataIdx => {
```
gczsjdy: Test of different elements of `dataArray` can be split to different unit tests.
And also, there are some common parts that can be extracted, making the function shorter and more readable.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/OapBitmapWrappedFiberCache.scala
lee-lei: All of the functions are made to be private.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/OapBitmapWrappedFiberCache.scala
lee-lei: The fiber cache itself abstracts the offheap address. And OAP data files and index files are directly loaded into fiber caches which are managed by fiber cache managers. In order to avoid data copy, we just need to be based on fiber cache instead of accessing the detailed address, since it's encapsulated by fiber cache already. Thanks.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapMicroBenchmarkSuite.scala
lee-lei: IMHO, the common parts are extracted already. It is just to extract the common part to use dataArray. Thanks.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,186 @@
...
+            val entrySize = getIdxOffset(bmOffsetListCache.fc, 0L, idx + 1) - curIdxOffset
+            val entryFiber = BitmapFiber(
+              () => loadBmEntry(fin, curIdxOffset, entrySize), idxPath.toString,
+            BitmapIndexSectionId.entryListSection, idx)
+            val wfc = new OapBitmapWrappedFiberCache(FiberCacheManager.get(entryFiber, conf))
+            wfc.init
+            val rowIdIterator = ChunksInSingleFiberCacheIterator(wfc).init
+            // The row id is starting from 0.
+            val rowIdSeq = Seq(0 to rowIdIterator.max)
+            rowIdIterator.foreach(rowId => assert(rowIdSeq.contains(rowId) == true))
```
gczsjdy: Each of set A is contained in B is not enough? The other direction?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapMicroBenchmarkSuite.scala
gczsjdy: Actually, the unit test can be more `unit`, different elements are actually different units. Like testing `run-length encoding`, testing `Short Array Encoding`.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/OapBitmapWrappedFiberCache.scala
gczsjdy: You mean we can't extract the `RoaringBitMap` related code to another class right?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,186 @@
...
+      sql("insert overwrite table oap_test select * from t")
+      sql("create oindex index_bm on oap_test (a) USING BITMAP")
+      dir.listFiles.foreach(fileName => {
+        if (fileName.toString.endsWith(OapFileFormat.OAP_INDEX_EXTENSION)) {
+          val idxPath = new Path(fileName.toString)
+          val conf = new Configuration()
+          val fin = idxPath.getFileSystem(conf).open(idxPath)
+          val (keyCount, bmOffsetListCache, bmFooterCache) =
+            getMetaDataAndFiberCaches(fin, idxPath, conf)
+          val wfcSeq = (0 until keyCount).map(idx => {
```
gczsjdy: Please use full name.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
```diff
@@ -0,0 +1,113 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.execution.datasources.oap.utils
+
+import scala.collection.mutable.ArrayBuffer
+import scala.util.control.Breaks._
+
+// The chunks are physically consecutive, so it doesn't require to set chunk offset.
+private[oap] case class ChunksInSingleFiberCacheIterator(wfc: OapBitmapWrappedFiberCache)
```
gczsjdy: Good to add comments for the classes added. : )
But first it should address the functionality of this class?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
```diff
@@ -0,0 +1,113 @@
...
+  }
+}
+
+// The chunks are not physically consecutive, so it requires to set chunk offset.
+private[oap] case class ChunksInMultiFiberCachesIterator(
+    chunksInFc: ArrayBuffer[OapBitmapChunkInFiberCache])
+  extends ChunksIterator {
+
+  override def init(): Iterator[Int] = {
+    totalLength = chunksInFc.length
```
gczsjdy: :nit Maybe we don't save letters here.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
```diff
@@ -0,0 +1,113 @@
...
+      iteratorForChunk = wfc.getIteratorForChunk(idx)
+      val cks = wfc.getChunkKeys
+      highPart = (cks(idx) & 0xFFFF) << 16
+    }
+    return this
+  }
+}
+
+// The chunks are not physically consecutive, so it requires to set chunk offset.
+private[oap] case class ChunksInMultiFiberCachesIterator(
```
gczsjdy: Compared to the former class with similar names, this one takes use of `...FastAggregation`? Can we explain this in comments?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
```diff
@@ -0,0 +1,113 @@
...
+    (1 until wfcSeq.length).foreach(idx => {
+      initialIdx = 0
+      var initialKey = finalChunkArray(initialIdx).getChunkKey
+      val nextWfc = wfcSeq(idx)
+      nextWfc.init
+      nextIdx = 0
+      val nextChunkLength = nextWfc.getTotalChunkLength
+      val nextChunkKeys = nextWfc.getChunkKeys
+      var nextKey = nextChunkKeys(nextIdx)
+      breakable {
```
gczsjdy: Spark code seldom uses this, it's just not 'scala', also the `while(true)` in the next line.

Can we try to avoid this?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
```diff
@@ -0,0 +1,113 @@
...
+        while (true) {
+          val result = initialKey - nextKey
+          result match {
+            case res if res < 0 =>
```
gczsjdy: If we only check the value, doesn't need to use pattern match.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
```diff
@@ -0,0 +1,113 @@
...
+      highPart = (chunksInFc(idx).getChunkKey & 0xFFFF) << 16
+    }
+    return this
+  }
+}
+
+private[oap] object OapBitmapFastAggregation {
+
+  // Just get array of the chunks across multi fiber caches in acending order of key.
+  def or(wfcSeq: Seq[OapBitmapWrappedFiberCache]): ArrayBuffer[OapBitmapChunkInFiberCache] = {
```
gczsjdy: Do we need to support `and`? For example `select * from table where a = 'banana' and b = 'tomato'`

lee-lei: They are two things. For your case, a and b are two columns. Thanks.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
lee-lei: They are different things. I will add some comments. Thanks.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
lee-lei: I followed the logic in the implementation of roaring bitmap. To be honest, I tried to replace it with scala common logic. But I didn't figure out the effective approach. Actually it's convenient to identify the problem/issue to use the same logic as roaring bitmap.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
gczsjdy: This is weird, go through 0 to max doesn't make sense.

We should make sure the 2 sets are exactly the same, which means each of A is contained in B, and vice versa.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
```diff
@@ -0,0 +1,112 @@
...
+      val chunkIdx = chunksInFc(idx).chunkIdx
+      wfc.setOffset(chunkIdx)
+      iteratorForChunk = wfc.getIteratorForChunk(chunkIdx)
+      highPart = (chunksInFc(idx).getChunkKey & 0xFFFF) << 16
+    }
+    return this
+  }
+}
+
+private[oap] object OapBitmapFastAggregation {
```
gczsjdy: Can we also provide an interface here to use functionality of `OapBitmapFastAggregation.or`? then this is a full `BitMapUtils`

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
lee-lei: This is a little bit time consuming for this test cases with larger data set. I will add another test cases to use small data sets to validate it. Thanks.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
gczsjdy: If it's just for test, it should exist in test package.
Or if not, we can explain it.

lee-lei: It's not for test. It's used by the bitmap scanner. Thanks.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
lee-lei: It's not appropriate to use Set to just match if they are equal, because it can't avoid the possibility that the actual row ID list may contain some repeated same elements.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,264 @@
...
+      val fin = rbPath.getFileSystem(conf).open(rbPath)
+      val rbFileSize = rbPath.getFileSystem(conf).getFileStatus(rbPath).getLen
+      val rbFiber = BitmapFiber(() => loadBmSection(fin, 0, rbFileSize.toInt), rbPath.toString, 0, 0)
+      fin.close
+      new OapBitmapWrappedFiberCache(FiberCacheManager.get(rbFiber, conf))
+    })
+    var actualRowIdSeq = Seq.empty[Int]
+    val chunkList = BitmapUtils.or(wrapperFiberCacheSeq)
+    ChunksInMultiFiberCachesIterator(chunkList).init.foreach(rowId => actualRowIdSeq :+= rowId)
+    actualRowIdSeq.foreach(rowId => assert(seqCombination.contains(rowId) == true))
```
gczsjdy: Just `assert(xx)`, `=true` not needed.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,264 @@
...
+      rb.serialize(rbDos)
+      rbBos.writeTo(rbFos)
+      rbBos.close()
+      rbDos.close()
+      rbFos.close()
+      val rbPath = new Path(rbFile)
+      val conf = new Configuration()
+      val fin = rbPath.getFileSystem(conf).open(rbPath)
+      val rbFileSize = rbPath.getFileSystem(conf).getFileStatus(rbPath).getLen
+      val rbFiber = BitmapFiber(() => loadBmSection(fin, 0, rbFileSize.toInt), rbPath.toString, 0, 0)
```
gczsjdy: We should extract some common code. Now too many duplicated.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCacheSuite.scala
```diff
@@ -0,0 +1,87 @@
...
+      rbBos.writeTo(rbFos)
+      rbBos.close()
+      rbDos.close()
+      rbFos.close()
+      val rbPath = new Path(rbFile.toString)
+      val conf = new Configuration()
+      val fin = rbPath.getFileSystem(conf).open(rbPath)
+      val rbFileSize = rbPath.getFileSystem(conf).getFileStatus(rbPath).getLen
+      val rbFiber = BitmapFiber(() => loadRbFile(fin, 0, rbFileSize.toInt), rbPath.toString, 0, 0)
+      val rbWfc = new OapBitmapWrappedFiberCache(FiberCacheManager.get(rbFiber, conf))
```
gczsjdy: Can you change all the abbrevations in this PR? 
They are actually hard to read.
For example this can be `wrapped`.
`rbDos` can be `outputStream`. Since all in this class are related to `RoaringBitmap` & `data`

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  test("test ChunksInSingleFiberCacheIterator to get the expected row IDs") {
+    val seqArray =
+      Array(seqForRunChunk, seqForArrayChunk, seqForBitmapChunk, seqCombination)
+    seqArray.foreach(seqIdx => {
+      val subDir = Utils.createTempDir()
+      val rb = new RoaringBitmap()
+      seqIdx.foreach(rb.add)
+      val rbFile = subDir.getAbsolutePath + "rb.bin"
```
gczsjdy: `rbFile`, `subDir` are both only used once. In this case they don't need a `val`.

In a word, maybe more refactor is needed to make the code more clean and less procedural.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,264 @@
...
+            | USING oap
+            | OPTIONS (path '$path')""".stripMargin)
+  }
+
+  override def afterEach(): Unit = {
+    sqlContext.dropTempTable("oap_test")
+    dir.delete()
+  }
+
+  test("test ChunksInSingleFiberCacheIterator to get the expected row IDs") {
```
gczsjdy: :nit No `test`

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,264 @@
...
+    val seqArray =
+      Array(seqForRunChunk, seqForArrayChunk, seqForBitmapChunk, seqCombination)
```
gczsjdy: I strongly think this should be split into 4 unit tests, instead of combine them into on array. @chenghao-intel Your opinion?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,264 @@
...
+            wrappedFiberCache.release
+          })
+          // The row Id is starting from 0.
+          accumulatorRowId += maxRowIdInPartition + 1
+          footerCache.release
+          offsetListCache.release
+          fin.close
+        }
+      })
+      // It's not appropriate to use Set to just match if they are equal, because it can't avoid
```
gczsjdy: As we have synced, it's appropriate to use `Set`, because rowId in one partition won't be duplicated.

Anyway, this comment doesn't have to exist here.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,264 @@
...
+      // the possibility that the actual row ID seq may contain some repeated same elements.
+      actualRowIdSeq.foreach(rowId => assert(expectedRowIdSeq.contains(rowId) == true))
+      expectedRowIdSeq.foreach(rowId => assert(actualRowIdSeq.contains(rowId) == true))
+      sql("drop oindex index_bm on oap_test")
+    })
+  }
+
+  test("test how to directly get the row Id list after bitwise OR among multi fiber caches") {
```
gczsjdy: In this test suite, we probably don't need to test the first 2 unit tests for `ChunksInMultiFiberCachesIterator`, etc.
Just the last 2 will be enough. They will cover the first 2. Right?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  test("test how to directly get the row ID list from single fiber cache without roaring bitmap") {
+    dataArray.foreach(dataIdx => {
+      dataIdx.toDF("key", "value").createOrReplaceTempView("t")
+      sql("insert overwrite table oap_test select * from t")
+      sql("create oindex index_bm on oap_test (a) USING BITMAP")
+      // For dataCombination, the total rows is 3 * 20000 = 60000.
+      val expectedRowIdSeq = if (dataIdx != dataCombination) (0 until 20000).toSeq
+        else (0 until 60000).toSeq
```
gczsjdy: Don't use this hardcode, get `expected` from `dataIdx` is appropriate.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,264 @@
...
+            wrappedFiberCache.init
+            ChunksInSingleFiberCacheIterator(wrappedFiberCache).init.foreach(rowId => {
+              val realRowId = rowId + accumulatorRowId
+              actualRowIdSeq :+= realRowId
+              if (maxRowIdInPartition < rowId) maxRowIdInPartition = rowId
+            })
+            wrappedFiberCache.release
+          })
+          // The row Id is starting from 0.
+          accumulatorRowId += maxRowIdInPartition + 1
```
gczsjdy: Not used?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,264 @@
...
+  private val BITMAP_FOOTER_SIZE = 6 * 8
+  private var dir: File = _
+  private val dataForRunChunk: Seq[(Int, String)] =
+    (1 to 20000).map{i => (i / 100, s"this is test $i")}
+  private val dataForArrayChunk: Seq[(Int, String)] =
+    (1 to 20000).map{i => (i, s"this is test $i")}
+  private val dataForBitmapChunk: Seq[(Int, String)] =
+    (1 to 20000).map{i => (i % 2, s"this is test $i")}
+  private val dataCombination =
+    dataForBitmapChunk ++ dataForArrayChunk ++ dataForRunChunk
```
gczsjdy: To cover more full, we can split the `dataCombination` to 4 cases: run+array, array+bitmap, run+bitmap, all three.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
```diff
@@ -0,0 +1,122 @@
...
+              initialKey = finalChunkArray(initialIdx).getChunkKey
+              nextKey = nextChunkKeys(nextIdx)
+            } else if (result > 0) {
+              // Insert the next chunk with nextIdx from the next fiber cache.
+              finalChunkArray.insert(initialIdx, OapBitmapChunkInFiberCache(nextWfc, nextIdx))
+              initialIdx += 1
+              nextIdx += 1
+              if (nextIdx == nextChunkLength) break
+              nextKey = nextChunkKeys(nextIdx)
+          }
```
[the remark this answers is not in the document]
lee-lei: Thanks for the good catch!

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCache.scala
```diff
@@ -0,0 +1,275 @@
...
+import org.apache.spark.sql.execution.datasources.OapException
+import org.apache.spark.sql.execution.datasources.oap.filecache.{FiberCache, WrappedFiberCache}
+
+// Oap Bitmap Scanner will use below class to directly traverse the row IDs
+// in fiber caches while keeping compatible with Roaring Bitmap format.
+// It doesn't need to create the roaring bitmap object on heap any more.
+// The spec link is https://github.com/RoaringBitmap/RoaringFormatSpec/
+private[oap] class OapBitmapWrappedFiberCache(fc: FiberCache) extends WrappedFiberCache(fc) {
+
+  private val SERIAL_COOKIE: Int = 12347
```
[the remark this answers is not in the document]
lee-lei: Sure.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCache.scala
```diff
@@ -0,0 +1,275 @@
...
+  private var hasRun: Boolean = false
+  private var bitmapOfRunChunks: Array[Byte] = _
+
+  private var chunkKeys: Array[Short] = _
+  private var chunkCardinalities: Array[Short] = _
+
+  // The reading byte order is little endian to keep consistent with roaring bitmap writing order.
+  // Read from the specific offset.
+  private def getIntNoMoving(offset: Int): Int = {
+    val curPos = offset
```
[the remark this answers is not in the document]
lee-lei: Sure.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
lee-lei: accumulatorRowId is required for multi partitions. Thanks.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCacheSuite.scala
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.spark.util.Utils
+
+// Below are used to test the functionality of OapBitmapWrappedFiberCache class.
+class OapBitmapWrappedFiberCacheSuite
+  extends QueryTest with SharedOapContext {
+
+  private def loadRbFile(fin: FSDataInputStream, offset: Int, size: Int): FiberCache =
+    MemoryManager.putToIndexFiberCache(fin, offset, size)
+
+  test("test the public methods of OapBitmapWrappedFiberCache class") {
```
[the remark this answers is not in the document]
lee-lei: I think that the separate test for three different chunk is validated in BitmapUtils class. In this class three different data sources defined for run chunk, array chunk, and bitmap chunk are just to achieve this goal. Thanks.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
lee-lei: Sure.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
lee-lei: Thanks for the suggestion.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
lee-lei: Sure.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCache.scala
```diff
@@ -0,0 +1,268 @@
...
+        return ArrayChunkIterator(this, idx)
+      case idx if (isBitmapChunk(idx)) =>
+        return BitmapChunkIterator(this)
+      case _ =>
+        throw new OapException("It's illegal chunk in bitmap index fiber caches.\n")
+    }
+  }
+
+  def init(): Unit = {
+    val cookie = getInt
```
linhong-intel: function declaration and usage should be same. I mean both with `()` or without `()`
So you either use:
```
def getInt: Int
val cookie = getInt
```
or
```
def getInt(): Int
val cookie = getInt()
```

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCache.scala
```diff
@@ -0,0 +1,268 @@
...
+      idx += 1
+      init
+    }
+    value
+  }
+}
+
+private[oap] case class RunChunkIterator(wfc: OapBitmapWrappedFiberCache) extends Iterator[Int] {
+
+  private var totalRuns: Int = wfc.getShort & 0xFFFF
```
linhong-intel: var is unused.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCache.scala
```diff
@@ -0,0 +1,268 @@
...
+      ((fc.getByte(offset + 1) & 0xFF) << 8) |
+      (fc.getByte(offset) & 0xFF)
+  }
+
+  private def getChunkSize(chunkIdx: Int): Int = {
+    // NO run chunks in this case.
+    chunkIdx match {
+      case idx if (isArrayChunk(idx)) =>
+        (chunkCardinalities(idx) & 0xFFFF) * 2
+      case idx if (isBitmapChunk(idx)) =>
```
linhong-intel: remove unnecessary parentheses

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCache.scala
```diff
@@ -0,0 +1,268 @@
...
+      curOffset += accumulOffset
+    } else {
+      curOffset = getIntNoMoving(chunkOffsetListOffset + chunkIdx * 4)
+    }
+  }
+
+  def getIteratorForChunk(chunkIdx: Int): Iterator[Int] = {
+    chunkIdx match {
+      case idx if (isRunChunk(idx)) =>
+        return RunChunkIterator(this)
```
linhong-intel: `return` is unnecessary

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCache.scala
```diff
@@ -0,0 +1,268 @@
...
+        accumulOffset += getChunkSize(idx))
+      curOffset += accumulOffset
+    } else {
+      curOffset = getIntNoMoving(chunkOffsetListOffset + chunkIdx * 4)
+    }
+  }
+
+  def getIteratorForChunk(chunkIdx: Int): Iterator[Int] = {
+    chunkIdx match {
+      case idx if (isRunChunk(idx)) =>
```
linhong-intel: parentheses is unnecessary

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCache.scala
```diff
@@ -0,0 +1,268 @@
...
+  private var chunkOffsetListOffset: Int = -1
+
+  private var curOffset: Int = 0
+
+  // No run chunks by default.
+  private var hasRun: Boolean = false
+  private var bitmapOfRunChunks: Array[Byte] = _
+
+  private var chunkKeys: Array[Short] = _
+  private var chunkCardinalities: Array[Short] = _
```
linhong-intel: typo?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,204 @@
...
+    dataForBitmapChunk ++ dataForRunChunk
+  private val dataCombination3 =
+    dataForArrayChunk ++ dataForRunChunk
+  private val dataSourceArray =
+    Array((dataForRunChunk, expectedRowIdSeq), (dataForArrayChunk, expectedRowIdSeq),
+      (dataForBitmapChunk, expectedRowIdSeq), (dataCombination1, expectedRowIdSeqCombination),
+      (dataCombination2, expectedRowIdSeqCombination),
+      (dataCombination3, expectedRowIdSeqCombination),
+      (dataCombinationTotal, expectedRowIdSeqCombinationTotal))
+*/
```
linhong-intel: If you don't need this. just remove it.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
```diff
@@ -0,0 +1,128 @@
...
+    }
+    return this
+  }
+}
+
+/* The chunks inside of different multi fiber cache are not physically consecutive, so it
+ * requires to set chunk offset for different fibers. This class is to directly get the row ID
+ * list recorded in all of the chunks in all of multi fiber caches in ascending order.
+ */
+private[oap] case class ChunksInMultiFiberCachesIterator(
```
linhong-intel: Please combine `ChunksInSingleFiberCacheIterator ` and `ChunksInMultiFiberCachesIterator `
single should be covered by multiple

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
```diff
@@ -0,0 +1,204 @@
...
+          })
+          // The row Id is starting from 0.
+          accumulatorRowId += maxRowIdInPartition + 1
+          footerCache.release
+          offsetListCache.release
+          fin.close
+        }
+      })
+      actualRowIdSeq.foreach(rowId => assert(dataSourceElement._2.contains(rowId)))
+      dataSourceElement._2.foreach(rowId => assert(actualRowIdSeq.contains(rowId)))
```
linhong-intel: use `sameElements` to assert 2 Seqs are same
In your assert, you should make sure the result sequence are same, too.

linhong-intel:
About your implementation, here is my suggestion:
Current usage is:
```
val fiberCache: FiberCache = FiberCacheManager.get(fiber)
val obwfc: OapBitmapWrappedFiberCache = OapBitmapWrappedFiberCache(fiberCache)
obwfc.init()
val cisfci: ChunksInSingleFiberCacheIterator = ChunksInSingleFiberCacheIterator(obwfc)
cisfci.init()
val iterator = cisfci
```
I suggested usage:
```
val fiberCache: FiberCache = FiberCacheManager.get(fiber)
val obwfc: OapBitmapWrappedFiberCache = OapBitmapWrappedFiberCache(fiberCache)
val iterator = obwfc.iterator()
```

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtils.scala
```diff
@@ -0,0 +1,128 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.sql.execution.datasources.oap.utils
+
+import scala.collection.mutable.ArrayBuffer
+import scala.util.control.Breaks._
+
+/* The chunks inside of one single fiber cache are physically consecutive,
```
linhong-intel: multi-line comments for class and function should use:
```
  /**
   * xxxx
   */
```

src/test/scala/org/apache/spark/sql/execution/datasources/oap/utils/BitmapUtilsSuite.scala
lee-lei: This case finds the all of row IDs for key1, key2, ... So the returned row IDs are out-of-order. The BitMapIndexSuite test case does the exact check for specific key by using "select * where key = value" sql clause.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCache.scala
lee-lei: cardinality is an english word.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/OapBitmapWrappedFiberCache.scala
lee-lei: It seems ok for this case:). Please see below link: https://stackoverflow.com/questions/6939908/scala-methods-with-no-arguments.

PR #638, opened 2018-03-28
Bitmap index micro benchmarks for equal query cases

lee-lei:

This patch provides two micro performance test cases to test
the performance comparison w/ and w/o roaring bitmap for below
two bitmap index equal query cases.

1. select * from t where a = apple
2. select * from t where a in {"apple", "banana", ...}

## How was this patch tested?

mvn clean test package and all of unit tests passed

### DIFF
--- a/src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapMicroBenchmarkSuite.scala
+++ b/src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapMicroBenchmarkSuite.scala
@@ -20,12 +20,18 @@ package org.apache.spark.sql.execution.datasources.oap.index
 import java.io._
 import java.util
 
+import scala.collection.JavaConverters._
 import scala.collection.mutable
 
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{FSDataInputStream, Path}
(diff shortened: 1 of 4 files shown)


src/test/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapMicroBenchmarkSuite.scala
```diff
@@ -20,12 +20,18 @@ package org.apache.spark.sql.execution.datasources.oap.index
 import java.io._
```
gczsjdy: Agree with @chenghao-intel , probably this Micro-benchmark PR shouldn't be merged to the code chunk.
We can add a [DO NOT MERGE] to the title to indicate this is just for review perpose.

gczsjdy: And the benchmark result can be shown in the #637

lee-lei: I will close this PR and keep the micro benchmark locally. Thanks.

lee-lei:
Close this PR. Thanks.

PR #666, opened 2018-04-03
Add SparkEnv.scala

gczsjdy:

Add SparkEnv.scala.

This class is added to collect more Spark code changes to 1 class, so as to decouple OAP with Spark as much as possible. SparkEnv seems a good entrance, it runs on both Driver and Executors, and it's easy to get its reference. An `OapManager` may be later inserted in the parameter list, and it holds various `OapxxManager` references.

In the future, we could migrate some of(not all) the Spark code in current OAP to this class.

## How was this patch tested?

Current UTs.


gczsjdy:
@linhong-intel

linhong-intel:
just original spark code, no changes. right?

PR #672, opened 2018-04-08
Migrate current Heartbeat to new RPC module

gczsjdy:

This is to migrate the current Heartbeat's sending Cache/Index/Metrics implementation(by changing Spark code) to new RPC module.

## How was this patch tested?

`FiberSensorSuite`


gczsjdy:
@carsonwang The new commit can solve #678

carsonwang:
Can you please make a separated PR to resolve #678?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberSensorSuite.scala
```diff
@@ -89,6 +88,8 @@ class FiberSensorSuite extends QueryTest with SharedOapContext
       }
     }
 
+    FiberCacheManager.registerHeartbeat()
```
carsonwang: Should this be registered already? Can we avoid this?

src/main/scala/org/apache/spark/SparkEnv.scala
```diff
@@ -379,7 +379,7 @@ object SparkEnv extends Logging {
     val oapRpcDriverEndpoint = registerOrLookupEndpoint(
       OapRpcManagerMaster.DRIVER_ENDPOINT_NAME, oapRpcManagerMasterEndpoint)
 
-    val oapRpcManager = if (isDriver) {
+    val oapRpcManager = if (isDriver && !isLocal) {
```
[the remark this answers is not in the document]
carsonwang: If it is local, only a slave will be created and no heartbeat will occur? It seems the master should always be created.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberSensorSuite.scala
carsonwang: any clue on this? This is not expected to be registered here.

linhong-intel:
@gczsjdy @carsonwang 
any update on this PR?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/OapDataReaderWriter.scala
```diff
@@ -200,6 +196,18 @@ private[oap] class OapDataReader(
 
   import org.apache.spark.sql.execution.datasources.oap.INDEX_STAT._
 
+  registerHeartbeat()
```
carsonwang: This is not the right place to registerHeartbeat - Each reader is going to register.

carsonwang:
Merging this. The only concern is the place to start heart beat. Ideally it should be started inside the OapRpcManagerSlave. Will have followup to see if it can be addressed.

gczsjdy:
It's indeed ideal to start it automatically, while the problem is that some state are not ready at that time.

Probably we can make this better if we introduce something like `OapContext`, we know that all Spark related class(like `SparkContext`, etc) has been ready.

PR #688, opened 2018-04-11
Resolve warn: postfix operator should be explicitly enabled

### LINKED ISSUE #685: Compile Warning in OapRpcManagerSlave
@gczsjdy can you check this issue?
```
[WARNING] /Users/linhongl/IdeaProjects/OAP/src/main/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSlave.scala:85: postfix operator += should be enabled
by making the implicit value scala.language.postfixOps visible.
Warning:(85, 48) postfix operator += should be enabled
by making the implicit value scala.language.postfixOps visible.
This can be achieved by adding the import clause 'import scala.language.postfixOps'
or by setting the compiler option -language:postfixOps.
See the Scaladoc for value scala.language.postfixOps for a discussion
why the feature should be explicitly enabled.
    getMaterials.foreach(oapHeartbeatMaterials +=)
```

gczsjdy:

To resolve #685 .

## How was this patch tested?

Existing UTs.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSlave.scala
+++ b/src/main/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSlave.scala
@@ -82,6 +82,7 @@ private[spark] class OapRpcManagerSlave(
   }
 
   private[spark] def registerHearbeat(getMaterials: Seq[() => Heartbeat]): Unit = {
+    import scala.language.postfixOps
     getMaterials.foreach(oapHeartbeatMaterials +=)
   }
 }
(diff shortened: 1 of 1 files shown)


src/main/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSlave.scala
```diff
@@ -82,6 +82,7 @@ private[spark] class OapRpcManagerSlave(
   }
 
   private[spark] def registerHearbeat(getMaterials: Seq[() => Heartbeat]): Unit = {
+    import scala.language.postfixOps
     getMaterials.foreach(oapHeartbeatMaterials +=)
```
adrian-wang: I think we can just use
```
 getMaterials.foreach(oapHeartbeatMaterials += _)
```

gczsjdy: Both ways are graceful, : )

gczsjdy: Ok, I like your suggestion more.

linhong-intel:
LGTM.
checked the test log. no warning any more.

gczsjdy:
Thanks @linhong-intel @adrian-wang

PR #698, opened 2018-04-17
Throw Exception instead of assert in non-test code

gczsjdy:

- `Assert` should be used less in non-test code.
- Docs. 

## How was this patch tested?

Existing tests.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
@@ -123,7 +123,9 @@ case class CreateIndexCommand(
           metaBuilder.addIndexMeta(new IndexMeta(indexName, time, BTreeIndex(entries)))
         case BitMapIndexType =>
           // Currently OAP index type supports the column with one single field.
-          assert(indexColumns.length == 1, "BitMapIndexType only supports one single column")
+          if (indexColumns.length != 1) {
+            throw new OapException("BitMapIndexType only supports one single column")
+          }
           val entries = indexColumns.map(col =>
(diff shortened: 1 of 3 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -123,7 +123,9 @@ case class CreateIndexCommand(
           metaBuilder.addIndexMeta(new IndexMeta(indexName, time, BTreeIndex(entries)))
         case BitMapIndexType =>
           // Currently OAP index type supports the column with one single field.
-          assert(indexColumns.length == 1, "BitMapIndexType only supports one single column")
+          if (indexColumns.length != 1) {
```
chenghao-intel: that's a great catch!

src/main/scala/org/apache/spark/sql/execution/datasources/oap/package.scala
```diff
@@ -23,8 +23,11 @@ package object oap {
   type Key = InternalRow
 }
 
-class OapException(message: String, cause: Throwable)
-  extends Exception(message, cause) {
+/**
+ * To express OAP specific exceptions, including but not limited to indicate unsupported operations,
+ * for example: BitMapIndexType only supports one single column
+ */
+class OapException(message: String, cause: Throwable) extends Exception(message, cause) {
```
chenghao-intel: I am wondering if we can have more concrete OapException classes, this is too general, @gczsjdy can you please create an issue, I think we need couple of categories like:
`OAPCacheInsufficientException`, `OapIndexColumnException`, `UnsupportedOapDataTypeException` etc.

gczsjdy:
Thanks @chenghao-intel

PR #701, opened 2018-04-17
[Spark code change remove]remove json protocol

adrian-wang:

remove json protocol file from repo.
we only add two match error for `logEvent`, but in `SparkListener` we never called `logEvent`. I think we can find some ways to log the events, not via the `sparkEventToJson` method. we can create a `oapEventToJson` method elsewhere.
```
case customInfoUpdate: SparkListenerCustomInfoUpdate =>
  throw new MatchError(customInfoUpdate)  // TODO to implement this
case customInfoUpdate: SparkListenerOapIndexInfoUpdate =>
  throw new MatchError(customInfoUpdate)  // TODO to implement this
```

## How was this patch tested?

ut


carsonwang:
`logEvent` is only used to write the event log for the history server. There is no need to write the OAP events, right? By the way, is there a SPARK API change issue related to this?

adrian-wang:
No related issue :(

carsonwang:
LGTM

PR #710, opened 2018-04-23
[WIP]Return iterator directly if there is no partition columns.

yueawang:
It somehow saves a few genericUnsafeProjection overhead.

## How was this patch tested?
mvn test.


linhong-intel:
@yueawang 
Any update on this PR?

yueawang:
Did not see enough potential improvement of this change, will close it.
And I think OAP vectorization (if it comes in future) is able to implement similar logic.

PR #713, opened 2018-04-23
Remove derpecated code and tests

gczsjdy:

Remove deprecated code, this will cause
```
Warning:(47, 32) method readRowIdList in class BTreeIndexFileReaderV1 is deprecated: no need to read the whole row id list
    val rowIdListRead = reader.readRowIdList().toArray
```

## How was this patch tested?

Existing tests.


gczsjdy:
Seems this has been removed in latest master...

PR #717, opened 2018-04-24
Fix for endless `RpcEnv already stopped`

gczsjdy:

Before this PR, endless `WARN: RpcEnv already stopped` is shown after `OapRpcManagerSuite` is run,
due to my mistake of forgetting to close the manually started threadPool.

We should manually stop it here due to it's manually constructed, so not managed by Spark.

## How was this patch tested?

Current tests.


adrian-wang:
LGTM thanks!

PR #723, opened 2018-04-26
Start oap heartbeater automatically in OapRpcManagerSlave

carsonwang:
A followup to #672 Migrate current Heartbeat to new RPC module
Start oap heartbeater automatically in OapRpcManagerSlave.
@gczsjdy , please take a look.

## How was this patch tested?
Existing test


src/test/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSuite.scala
```diff
@@ -122,10 +121,8 @@ class OapRpcManagerSuite extends SparkFunSuite with BeforeAndAfterEach with Priv
 
   // This doesn't need to be spied due to it's used to send messages
   private def addRpcManagerSlave(executorId: String): OapRpcManagerSlave = {
-    class TestDummyHeartbeatMaterials extends OapHeartbeatMaterialsInterface{
-      override def get: HashSet[() => Heartbeat] = HashSet[() => Heartbeat] (() => heartbeat)
+    new OapRpcManagerSlave(rpcEnv, rpcDriverEndpoint, executorId, null, sc.conf) {
+      override def heartbeatMessages: Array[() => Heartbeat] = { Array(() => heartbeat) }
```
gczsjdy: This is beautiful. : )

src/main/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSlave.scala
```diff
@@ -38,53 +37,52 @@ private[spark] class OapRpcManagerSlave(
...
+  startOapHeartbeater()
+
+  protected def heartbeatMessages: Array[() => Heartbeat] = {
+    Array(
+      () => FiberCacheHeartbeat(
+        executorId, blockManager.blockManagerId, FiberCacheManager.status()),
+      () => IndexHeartbeat(executorId, blockManager.blockManagerId, OapIndexInfo.status),
+      () => FiberCacheMetricsHeartbeat(executorId, blockManager.blockManagerId,
+        CacheStats.status(FiberCacheManager.cacheStats, conf)))
+  }
```
gczsjdy: I think it's better to move this to another class, due to the `Heartbeat` related things is not technically bounded to `RpcManager` running on slave. It's the implementation that makes them together.

In a word, that will bring better decoupling.

carsonwang: It is not really necessary to wrap these messages in a class. It just likes RpcManager uses message A, B and C. If you wraps A, B, C to class D, then the RpcManager uses D. Similar to the master that need handling these messages, it is OK to directly use them.

src/test/scala/org/apache/spark/sql/oap/rpc/OapRpcManagerSuite.scala
```diff
@@ -96,14 +96,13 @@ class OapRpcManagerSuite extends SparkFunSuite with BeforeAndAfterEach with Priv
     rpcManagerSlave1.send(message)
     Thread.sleep(2000)
     verify(rpcManagerMasterEndpoint).invokePrivate(_handleNormalOapMessage(message))
+    rpcManagerSlave1.stop()
```
carsonwang: We also need stop it here. Otherwise a DummyMessage will always be sent.

carsonwang:
The test passed cleanly. I didn't see any warning. @gczsjdy , please let me know if you see any warning.

gczsjdy:
Thanks @carsonwang , I will figure this out and update this today.

gczsjdy:
@carsonwang The potential problem doesn't occur in this travis test. But it exists.

The `threadPool` started in `OapRpcManagerSlave` will wait for 1~2 * `heartbeatInterval`. This happens in the middle of `SparkContext` initilization.
If `SparkContext` finished construction during this time, then no exception will occur. If it is not, then `FiberCacheManager` will throw a `Oap can't run without SparkContext`, etc.
But can we make sure that, and this also depends on `heartbeatInterval`? Depending on this is not very reliable ?

gczsjdy:
Thanks @carsonwang , LGTM

PR #725, opened 2018-04-27
[DON'T MERGE] Exsort test

adrian-wang:
try to test travis


adrian-wang:
Even 1*1 failed the test with container, I think it is bug of container.

PR #726, opened 2018-04-27
LocalCluster test should require 2 * 1 as Spark do

gczsjdy:

From [travis-doc](https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System), only 2 cores are granted. In this condition, it's not necessary to use 2 * 2, `LocalCluster` mode aims to test under multi-executors, the parallelism in each doesn't matter.

## How was this patch tested?

Existing tests.

### DIFF
--- a/src/test/scala/org/apache/spark/sql/test/TestOapSession.scala
+++ b/src/test/scala/org/apache/spark/sql/test/TestOapSession.scala
@@ -43,7 +43,7 @@ private[sql] class TestOapLocalClusterSession(sc: SparkContext) extends TestSpar
   self =>
   def this(sparkConf: SparkConf) {
     this(new SparkContext(
-      s"local-cluster[2, 2, 1024]",
+      s"local-cluster[2, 1, 1024]",
       "test-oap-local-cluster-context",
       sparkConf.set("spark.sql.testkey", "true")
         .set("spark.hadoop.fs.file.impl", classOf[DebugFilesystem].getName)))


adrian-wang:
still not able to pass my case

adrian-wang:
while even 2*2 can pass my local vm with 2 cores.

gczsjdy:
@adrian-wang One time pass maybe cannot ensure consistent pass, it's good to follow travis guide to use <= 2 cores, even after we explicit indicate the environment type in `travis.yml`.
On the other hand, `LocalClusterMode` tests for cases using multiple executors, it dosn't matter if the parallel task number in 1 executor is 1 or 2.
So in either way, it's reasonable to change this.

adrian-wang:
Can you find one single run failed caused by this suite? If there is, I'll merge this.

gczsjdy:
[build](https://travis-ci.org/Intel-bigdata/OAP/builds/370857473?utm_source=github_status&utm_medium=notification) failed because of it.
Even if no suites failed because of it, this changing should be done, because
- travis granted ~2 cores
- There is no need to use 2 * 2, 2 * 1 is enough, `local-cluster` in Spark use the latter

adrian-wang:
The build you provided failed just because the vm setting... you can see it in the log, it is using ec2 container.
```
Worker information
hostname: 6868455d-e261-4f45-aab8-ad5f84d357e8@1.i-0ba8b54-production-2-worker-org-ec2.travisci.net
version: v3.6.0 https://github.com/travis-ci/worker/tree/170b2a0bb43234479fd1911ba9e4dbcc36dadfad
instance: a5fb9fd travisci/ci-garnet:packer-1512502276-986baf0 (via amqp)
startup: 1.854581411s
mode of ‘/usr/local/clang-5.0.0/bin’ changed from 0777 (rwxrwxrwx) to 0775 (rwxrwxr-x)
```

adrian-wang:
Also, this pr will never fix that build.

adrian-wang:
The true reason for the failure is ec2 container.
See https://forums.aws.amazon.com/thread.jspa?threadID=86751

gczsjdy:
Thanks @adrian-wang much. Now it gets much clearer.
Then my thought will be: we run on `local-cluster` mode we want to test in multi-executors' condition, the parallelism in one executor doesn't matter, so keep it as less as possible is better.

Anyway, I am okay with closing this, this is minor.

PR #735, opened 2018-05-07
Removed WrappedFiberCache class for BitmapUtils class and tests

lee-lei:

Since WrappedFiberCache class is reprecated, removed the usage of this class from BitmapUtils class and corresponding test cases.

## How was this patch tested?

mvn clean test package and all tests passed locally.


linhong-intel:
good to me

PR #738, opened 2018-05-09
[Draft] full scan cache profiling

linhong-intel:

draft code to improve oap full scan

## How was this patch tested?

no need test


linhong-intel:
### benchmark result

#### Test environment
```
Executor: 1
Core: 1
On-heap memory: 100G
Off-heap memory: 100G
Data File: 1
```

#### Create table
```
import org.apache.spark.sql.{QueryTest, Row}
import org.apache.spark.sql.Row
val rowCount = 1024 / 8 * 1024 * 1024 * 5
val rowRdd = sc.parallelize(0 until rowCount, 1).map(i => Seq(i.toLong)).map(Row.fromSeq)
val df = spark.createDataFrame(rowRdd, schema)
df.createOrReplaceTempView("t")
spark.sql("CREATE TABLE oap_test (a Long) Using OAP OPTIONS (path 'hdfs:///temp')")
sql("insert overwrite table oap_test select * from t")
```

#### Query
```
import org.apache.spark.sql.internal.oap.OapConf
def time[T](code: => T, action: String): Unit = {
  val t0 = System.nanoTime
  code
  val t1 = System.nanoTime
  println(action + ((t1 - t0) / 1000000) + "ms")
}
spark.conf.set(OapConf.OAP_DEBUG_FLAG.key, value = true)
time( spark.sql("select * from oap_test").foreach { _ => }, "no cache: ")
time( spark.sql("select * from oap_test").foreach { _ => }, "after oap cache#1: ")
time( spark.sql("select * from oap_test").foreach { _ => }, "after oap cache#2: ")
time( spark.sql("select * from oap_test").foreach { _ => }, "after oap cache#3: ")
time( spark.sql("cache table oap_test"), "cache table cost: ")
time( spark.sql("select * from oap_test").foreach { _ => }, "after spark cache#1: ")
time( spark.sql("select * from oap_test").foreach { _ => }, "after spark cache#2: ")
time( spark.sql("select * from oap_test").foreach { _ => }, "after spark cache#3: ")
```

#### Result
```
no cache: 48078ms
after oap cache#1: 26083ms
after oap cache#2: 33868ms
after oap cache#3: 30765ms
cache table cost: 142437ms
after spark cache#1: 41534ms
after spark cache#2: 41619ms
after spark cache#3: 40820ms
```

src/main/scala/org/apache/spark/sql/execution/datasources/oap/ColumnValues.scala
```diff
@@ -17,15 +17,76 @@
...
 import org.apache.spark.sql.catalyst.util.{ArrayData, MapData}
+import org.apache.spark.sql.execution.columnar.MutableUnsafeRow
 import org.apache.spark.sql.execution.datasources.oap.filecache.FiberCache
-import org.apache.spark.sql.types._
+import org.apache.spark.sql.types.{BooleanType, ByteType, DoubleType, FloatType, IntegerType, LongType, ShortType, StringType, _}
 import org.apache.spark.unsafe.types.{CalendarInterval, UTF8String}
 import org.apache.spark.util.collection.BitSet
 
-class ColumnValues(defaultSize: Int, dataType: DataType, val buffer: FiberCache) {
+class BooleanColumnValues(defaultSize: Int, buffer: FiberCache)
```
yueawang: This is a good idea, and I think OAP must adopt it. 
However, I don't know if we can directly use those ColumnType (Spark's column value used in InMemoryScan), which supports ByteBuffer, if we can wrap our fiberCache to ByteBuffer, maybe that is a smoother optimization.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/ColumnValues.scala
```diff
@@ -37,6 +98,27 @@ class ColumnValues(defaultSize: Int, dataType: DataType, val buffer: FiberCache)
     (buffer.getLong(idx >> 6 << 3) & bitmask) == 0  // div by 64 and mask
   }
 
+  def extractTo(row: InternalRow, idx: Int, column: Int): Unit = dataType match {
```
yueawang: Here we still have switch-cases, I think we can leverage the way Spark does, i.e. use type to differentiate the code path. 
Anyway, If this is a compile(code gen) time running code, that's no problem.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/OapDataFile.scala
```diff
@@ -179,6 +190,18 @@ private[oap] case class OapDataFile(
     buildIterator(configuration, requiredIds, Some(rowIds))
   }
 
+  def iteratorNG(requiredIds: Array[Int]): Iterator[InternalRow] = {
+
+    val columnTypes = requiredIds.map(i => schema(i).dataType)
+    val iterator = GenerateOapColumnAccessor.generate(columnTypes)
```
yueawang: Do we need completion check on this iterator??

linhong-intel: yes. we need. This can add in to code gen to save some nested map.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/ColumnValues.scala
linhong-intel: this is override by sub-class. never used actually.

yueawang: cool, then remove this function??

linhong-intel:
close

PR #742, opened 2018-05-11
skip stats building if no stats is enabled

adrian-wang:

If no stats is enabled, we don't have to keep things in memory.

## How was this patch tested?

N/A


adrian-wang:
I'll merge this as a simple improvement.

PR #745, opened 2018-05-14
reduced full scan code path

linhong-intel:

reduced full scan code path

## How was this patch tested?

null value test fails

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFullScanRDD.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFullScanRDD.scala
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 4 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/OapFullScanRDD.scala
```diff
@@ -0,0 +1,77 @@
...
+
+    val conf = new Configuration()
+    files.flatMap { file =>
+      InputFileNameHolder.setInputFileName(file.filePath)
+      val metaPath = new Path(file.filePath).getParent
+      val meta = DataSourceMeta.initialize(new Path(metaPath, OapFileFormat.OAP_META_FILE), conf)
+      val fileScanner = DataFile(file.filePath, meta.schema, meta.dataReaderClassName, conf)
+      val s = meta.schema
+      val ids = schema.map { field => s.indexOf(field)}
+      fileScanner.asInstanceOf[OapDataFile].iterator(ids.toArray)
```
ConeyLiu: Memory leak, the returned `iterator` need be closed after used. @linhong-intel @yangwanyuan

PR #755, opened 2018-05-17
Add mvn enforcer plugin to make sure mvn is compatible.

yueawang:

Add mvn enforcer plugin to make sure mvn is compatible.

## How was this patch tested?
No need.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -314,7 +314,28 @@
 
   <build>
     <plugins>
-
+     <plugin>
+       <inherited>true</inherited>
+       <groupId>org.apache.maven.plugins</groupId>
+       <artifactId>maven-enforcer-plugin</artifactId>
+       <version>1.3.1</version>
+         <executions>
+           <execution>
+             <id>enforce-maven-3</id>
+             <goals>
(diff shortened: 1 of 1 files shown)


gczsjdy:
What will happen if non-3.3 maven is used?

pom.xml
```diff
@@ -314,7 +314,28 @@
...
     <plugins>
-
+     <plugin>
+       <inherited>true</inherited>
+       <groupId>org.apache.maven.plugins</groupId>
+       <artifactId>maven-enforcer-plugin</artifactId>
+       <version>1.3.1</version>
+         <executions>
+           <execution>
+             <id>enforce-maven-3</id>
```
adrian-wang: add <phase> into <execution> to trigger auto run?

adrian-wang: add `<phase>` into `<execution>`

yueawang: enforcer:enforce
Attributes:
    Binds by default to the lifecycle phase: validate

No need to change actually for functionality, but can be done for readability :)

adrian-wang: ok, LGTM then

PR #763, opened 2018-05-22
Add func in CacheGuardian, refactor 2FiberData -> cache

gczsjdy:

- create a `releaseFiberCache` in `CacheGuardian`
- `2FiberData` / `fiberToData` -> `cache`

## How was this patch tested?

Existing tests.


gczsjdy:
@yueawang 
- 1 is a `TODO`
-  it's more clear this way, take `fiber2Data` as an example, it applies to a `Fiber`, so `fiber` is redundant, `data` is ambiguous. More important, what we actually do is putting data/index to OAP cache and get the memory reference(`FiberCahe`), so it's pretty nature to name it in the new way, it now reads like a sentence. : )

gczsjdy:
@carsonwang I got an offline LGTM from @linhong-intel ...

PR #768, opened 2018-05-25
remove kryo settings in OapEnv

adrian-wang:

This is what we should remove when upgrade to 2.1.0.
See SPARK-17350 .

## How was this patch tested?

N/A


linhong-intel:
ok to merge

PR #772, opened 2018-05-29
Add FiberLockManager and DataFileMetaCacheManager to OapRuntime

adrian-wang:

Add FiberLockManager and DataFileMetaCacheManager to OapRuntime

## How was this patch tested?

ut

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -155,7 +155,7 @@ private[sql] class FiberCacheManager(
     // The final bit set is: 010010001
     val statusRawData = dataFibers.groupBy(_.file).map {
       case (dataFile, fiberSet) =>
-        val fileMeta: DataFileMeta = OapRuntime.getOrCreate.dataFileMetaCacheManager(dataFile)
+        val fileMeta: DataFileMeta = OapRuntime.getOrCreate.dataFileMetaCacheManager.get(dataFile)
         val fiberBitSet = new BitSet(fileMeta.getGroupCount * fileMeta.getFieldCount)
         fiberSet.foreach(fiber =>
           fiberBitSet.set(fiber.columnIndex + fileMeta.getFieldCount * fiber.rowGroupId))
@@ -211,7 +211,7 @@ private[sql] class DataFileMetaCacheManager extends Logging {
         }
       })
 
-  def apply(fiberCache: DataFile): DataFileMeta = {
+  def get(fiberCache: DataFile): DataFileMeta = {
     cache.get(fiberCache)
   }
 
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
@@ -29,6 +29,7 @@ import org.apache.spark.internal.Logging
 import org.apache.spark.sql.execution.datasources.OapException
 import org.apache.spark.sql.execution.datasources.oap.io._
 import org.apache.spark.sql.execution.datasources.oap.utils.CacheStatusSerDe
+import org.apache.spark.sql.oap.OapRuntime
 import org.apache.spark.util.Utils
 import org.apache.spark.util.collection.BitSet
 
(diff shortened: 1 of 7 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -177,7 +178,7 @@ private[sql] class FiberCacheManager(
   }
 }
 
-private[oap] object DataFileMetaCacheManager extends Logging {
```
gczsjdy: after changing it to `class`, it's weird to keep `apply`, change to sth like `getMeta`?

src/main/scala/org/apache/spark/sql/oap/OapRuntime.scala
```diff
@@ -21,47 +21,44 @@ import org.apache.spark.{SparkConf, SparkContext, SparkEnv}
...
-private[oap] abstract class OapRuntime(sparkEnv: SparkEnv) extends Logging {
-  var memoryManager: MemoryManager = _
-  var fiberCacheManager: FiberCacheManager = _
-  var fiberSensor: FiberSensor = _
-  var oapRpcManager: OapRpcManager = _
-  var oapMetricsManager: OapMetricsManager = _
+private[oap] trait OapRuntime extends Logging {
+  val memoryManager: MemoryManager
+  val fiberCacheManager: FiberCacheManager
+  val fiberSensor: FiberSensor
```
gczsjdy: These 2 should be put in base trait? Seems more reasonable putting to 2 derived classes?

src/main/scala/org/apache/spark/sql/oap/OapRuntime.scala
```diff
@@ -21,47 +21,44 @@ import org.apache.spark.{SparkConf, SparkContext, SparkEnv}
...
+  val oapRpcManager: OapRpcManager
+  val oapMetricsManager: OapMetricsManager
+  // TODO are the following needed for driver?
+  val dataFileMetaCacheManager = new DataFileMetaCacheManager
+  val fiberLockManager = new FiberLockManager
```
gczsjdy: probably not needed : ) Same as last comment

src/main/scala/org/apache/spark/sql/oap/OapRuntime.scala
adrian-wang: If not in base class, we have to judge whether we are in the driver or executor. while I want to keep that logic inside of this.

gczsjdy: Like if in executor called `fiberSensor`, throw an exception?

adrian-wang: it will never happen

src/main/scala/org/apache/spark/sql/oap/OapRuntime.scala
```diff
@@ -21,47 +21,44 @@ import org.apache.spark.{SparkConf, SparkContext, SparkEnv}
...
-  var oapRpcManager: OapRpcManager = _
-  var oapMetricsManager: OapMetricsManager = _
+private[oap] trait OapRuntime extends Logging {
+  val memoryManager: MemoryManager
+  val fiberCacheManager: FiberCacheManager
+  val fiberSensor: FiberSensor
+  val oapRpcManager: OapRpcManager
+  val oapMetricsManager: OapMetricsManager
+  // TODO are the following needed for driver?
+  val dataFileMetaCacheManager = new DataFileMetaCacheManager
```
gczsjdy: New this in `ExecutorRuntime`?

src/main/scala/org/apache/spark/sql/oap/OapRuntime.scala
```diff
@@ -21,47 +21,44 @@ import org.apache.spark.{SparkConf, SparkContext, SparkEnv}
...
+ * Initializing [[FiberCacheManager]], [[MemoryManager]], [[FiberLockManager]]
+ * [[FiberSensor]], [[OapMetricsManager]], [[OapRpcManager]], [[DataFileMetaCacheManager]]
  */
-private[oap] abstract class OapRuntime(sparkEnv: SparkEnv) extends Logging {
-  var memoryManager: MemoryManager = _
-  var fiberCacheManager: FiberCacheManager = _
-  var fiberSensor: FiberSensor = _
-  var oapRpcManager: OapRpcManager = _
-  var oapMetricsManager: OapMetricsManager = _
+private[oap] trait OapRuntime extends Logging {
```
gczsjdy: Not understant why changing it to `trait`, actually it shouldn't be `with`ed by any class, I suggested: 
```
abstract class OapRuntime {
val memoryManager: MemoryManager = _
...
}
```
, then for `Driver/ExecutorRuntime`, override the object that exists in it.

linhong-intel:
@gczsjdy 
Could you please review this PR again?

gczsjdy:
LGTM , thanks @adrian-wang

PR #773, opened 2018-05-29
Get or cache a fiber doesn't need conf

gczsjdy:
## What changes were proposed in this pull request?

Modified corresponding `FiberCacheManager.get` and `Fiber.cache` functions' signature, removed `conf`.

## How was this patch tested?

Existing tests.

PR #777, opened 2018-06-05
Use cache loader instead of callable in GuavaOapCache

gczsjdy:

Follow up PR for #773 , `CacheLoader` is a better approach than `Callable`, see https://stackoverflow.com/questions/32283166/guava-cache-cacheloader-vs-getk-callable 

## How was this patch tested?

Existing tests.


linhong-intel:
LGTM, thanks @gczsjdy 
since we removed `configuration`, use `CacheLoader` is absolutely right

gczsjdy:
Thanks @linhong-intel

PR #780, opened 2018-06-06
Couple Fiber & FiberCache

gczsjdy:

Couple `Fiber` and `FiberCache`, while disposing the latter, doesn't need a `Fiber`

## How was this patch tested?

Existing tests


linhong-intel:
@gczsjdy Let's discuss some name convention before merge this.

PR #782, opened 2018-06-06
Reducing the locking scope to improve the efficiency of concurrently getting the same fiber in the cache manager

lee-lei:

Adding another lock is a little bit heavy when getting the desired fiber cache every time.

Guava cache manager provides the synchronized add/write if the entry is missed and
concurrent read mechanism if the entry is hit. Thus the existing locking is unnecessary.

## How was this patch tested?

mvn clean test package and all the tests passed.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
@@ -154,20 +154,13 @@ class GuavaOapCache(cacheMemory: Long, cacheGuardianMemory: Long) extends OapCac
     .build[Fiber, FiberCache]()
 
   override def get(fiber: Fiber, conf: Configuration): FiberCache = {
-    val readLock = FiberLockManager.getFiberLock(fiber).readLock()
-    readLock.lock()
-    try {
-      val fiberCache = cache.get(fiber, cacheLoader(fiber, conf))
-      // Avoid loading a fiber larger than MAX_WEIGHT / CONCURRENCY_LEVEL
-      assert(fiberCache.size() <= MAX_WEIGHT * KB / CONCURRENCY_LEVEL,
-        s"Failed to cache fiber(${Utils.bytesToString(fiberCache.size())}) " +
-          s"with cache's MAX_WEIGHT" +
-          s"(${Utils.bytesToString(MAX_WEIGHT.toLong * KB.toLong)}) / $CONCURRENCY_LEVEL")
-      fiberCache.occupy()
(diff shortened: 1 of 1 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -154,20 +154,13 @@ class GuavaOapCache(cacheMemory: Long, cacheGuardianMemory: Long) extends OapCac
...
-    } finally {
-      readLock.unlock()
-    }
+    /* Guava cache manager provides the synchronized add/write if the entry is missed and
+     * concurrent read if the entry is hit. In addition, Guava Cache will automatically
+     * first evict the entry if the size is above MAX_WEIGHT * KB / CONCURRENCY_LEVEL.
+     */
+    val fiberCache = cache.get(fiber, cacheLoader(fiber, conf))
+    fiberCache.occupy()
+    fiberCache
```
linhong-intel: @lee-lei 
screen shot 2018-06-07 at 10 35 08 am

lee-lei: @linhong-intel @gczsjdy  The guava cache manager meets this requirement. 
1. In our cache manager implementation, the cache content is still residing in memory as long as the refCount is not reduced to zero.
2. The cache eviction is done in guava cache manager. 

In other words, if the two threads concurrently get cacheA and cacheB, there are below four cases:
1. cacheA and cacheB will both be cached by guava cache manager.
2. cacheA is not cached. cacheB is cached. Thus the cacheA will get NULL by cache.get method.
3. cacheA is cached. cacheB is not cached. Thus the cacheB will get NULL by cache.get method.
4. cacheA and cacheB are both not cached. Thus both cacheA and cacheB will get NULL.

In your case, thread A gets cacheA successfully. Thread B tries to get cacheB. The guava cache manager will evict cacheA and then return cacheB. The cacheA is put in the queue. The memory of cacheA is stil there before the refCount is atomically reduced to zero.

refCount is atomic. No need another lock to guard it.

lee-lei:
For the existing code, the OAP cache manager can't support that several or many user threads concurrently read the same fiber cache. In other words, they can only serially access the same fiber if it's in the cache manager already. That indicates that the existing lock scope is too large. My plan is to refactor this PR and resolve the concurrency issue after the PR#796 is merged. Thanks.

lee-lei:
Close this PR, since the problem resolved by this PR is very rare in the production environment, and the #830 proposed another solution. Thanks.

PR #784, opened 2018-06-08
Backward compatible framework for OAP data file

gczsjdy:

Backward compatibility support for OAP file format, and only the newest version is used while creating new data files.

## How was this patch tested?

OapDataReaderSuite


src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/OapDataReaderWriter.scala
```diff
@@ -187,33 +192,57 @@ private[sql] object OapIndexInfo extends Logging {
   }
 }
 
-private[oap] class OapDataReader(
+private[oap] class OapDataReaderV1(
     pathStr: String,
-    meta: DataSourceMeta,
+    m: DataSourceMeta,
```
linhong-intel: why we change `meta` to `m`

gczsjdy:
Thanks @linhong-intel

PR #787, opened 2018-06-18
Reduce 1 nested loop, unecessary CompletionIterator

gczsjdy:

`inUseFiberCache` array is released whenever necessary: 
- In `update`
- In `close` of `OapIterator`

Accordingly, no need an extra nested CompletionIterator

## How was this patch tested?

N/A

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/DataFile.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/DataFile.scala
@@ -52,8 +52,17 @@ abstract class DataFile {
   }
 }
 
+// An Iterator automatically calling close() after iterations of all items
 private[oap] class OapIterator[T](inner: Iterator[T]) extends Iterator[T] with Closeable {
-  override def hasNext: Boolean = inner.hasNext
+  private[this] var completed = false
+  override def hasNext: Boolean = {
+    val r = inner.hasNext
+    if (!r && !completed) {
+      completed = true
+      close()
+    }
+    r
+  }
   override def next(): T = inner.next()
   override def close(): Unit = {}
 }
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/DataFile.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/DataFile.scala
@@ -40,9 +40,10 @@ abstract class DataFile {
 
   def getDataFileMeta(): DataFileMeta
   def cache(groupId: Int, fiberId: Int): FiberCache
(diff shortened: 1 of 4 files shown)


carsonwang:
The `CompletionIterator` is useful to release the fibers earlier and there should be no overhead. Considering the last iter, there is no `update` method to release the fibers. If you remove the `CompletionIterator`, you will wait until the task completion for the release. Otherwise it can be released earlier just after the iter is consumed.

gczsjdy:
@carsonwang I agree with you except 
> should be no overhead
One more nested loop is added.
I just merge the `CompletionIterator`'s functionality into OapIterator, it will solve the later relase issue.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/DataFile.scala
```diff
@@ -52,8 +52,17 @@ abstract class DataFile {
...
 
+// An Iterator automatically calling close() after iterations of all items
 private[oap] class OapIterator[T](inner: Iterator[T]) extends Iterator[T] with Closeable {
-  override def hasNext: Boolean = inner.hasNext
+  private[this] var completed = false
+  override def hasNext: Boolean = {
+    val r = inner.hasNext
+    if (!r && !completed) {
+      completed = true
+      close()
```
carsonwang: With this change, `close` will be called twice.

gczsjdy: Right, `close` should only be called in task completion listener. I have modified.

carsonwang:
This should work now.  One little drawback is that user need specify both `completionFunction` and `close` which are actually doing the similar thing.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/io/DataFile.scala
```diff
@@ -52,8 +52,20 @@ abstract class DataFile {
   }
 }
 
-private[oap] class OapIterator[T](inner: Iterator[T]) extends Iterator[T] with Closeable {
-  override def hasNext: Boolean = inner.hasNext
+// An OAP wrapped iterator calling completionFunction() automatically after iterations of all items
+// Also, it contains a close interface to do cleaning even if tasks fail
+private[oap] class OapIterator[T](inner: Iterator[T], completionFunction: => Unit = {}) extends
```
carsonwang: Do not use a default value for `completionFunction` to avoid leak? Maybe also rename `OapIterator` to something like `OapCompletionIterator`.

gczsjdy:
@carsonwang Yes, they are just slightly different. I have addressed the difference in comments : )

carsonwang:
LGTM. Thanks @gczsjdy .

PR #796, opened 2018-06-26
Improving CacheGuardian to free evicted fibers as soon as possible

lee-lei:

This PR is improving the cache guardian thread in order to:
1. ensure the pending fibers in the evicted queue to be freed as soon as possible
   after it's released by the last user.
2. ensure the queue to be empty as long as all the pending fibers are released by the last users.
With the above assurances, the memory pressure will be significantly mitigated.

We are based on the premise that the guava cache manager has a good eviction mechanism.
That indicates that the evicted fibers are relatively old and not accessed for a while and
won't be accessed in the future with the high possibility. Otherwise, we will upload another PR to address the inefficient eviction algorithm.

## How was this patch tested?

mvn clean test package and all the unit tests passed.

For the total 13 existing fiber cache manager test cases, the peak memory footprint of five are reduced significantly. The maximum memory size of cache manager is 7.34MB.

The detailed result is below:
The peak memory of "remove a fiber is in use" is reduced from 66.06MB to under 7.34MB.
The peak memory of "wait for other thread to release the fiber" is reduced from 31.46MB to under 7.34MB.
The peak memory of "add a very large fiber" is reduced from 39.32MB to under 7.34MB.
The peak memory of "cache guardian remove pending fibers" is reduced from 76.55MB to 69.21MB.
The peak memory of "get different fiber simultaneously" is reduced from 33.55MB to 12.58MB.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
@@ -28,8 +28,7 @@ import org.apache.spark.unsafe.Platform
 import org.apache.spark.unsafe.memory.MemoryBlock
 import org.apache.spark.unsafe.types.UTF8String
 
-// TODO: make it an alias of MemoryBlock
-case class FiberCache(var fiberId: FiberId, protected val fiberData: MemoryBlock) extends Logging {
+case class FiberCache(val fiberId: FiberId, protected val fiberData: MemoryBlock) extends Logging {
 
   // We use readLock to lock occupy. _refCount need be atomic to make sure thread-safe
   protected val _refCount = new AtomicLong(0)
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
@@ -51,45 +51,11 @@ case class FiberCache(protected val fiberData: MemoryBlock) extends Logging {
     _refCount.decrementAndGet()
   }
 
-  def tryDispose(): Boolean = {
-    require(fiberId != null, "FiberId shouldn't be null for this FiberCache")
-    val startTime = System.currentTimeMillis()
-    val writeLockOp = OapRuntime.get.map(_.fiberLockManager.getFiberLock(fiberId).writeLock())
-    writeLockOp match {
-      case None => return true // already stopped OapRuntime
-      case Some(writeLock) =>
-        // Give caller a chance to deal with the long wait case.
-        while (System.currentTimeMillis() - startTime <= DISPOSE_TIMEOUT) {
-          if (refCount != 0) {
-            // LRU access (get and occupy) done, but fiber was still occupied by at least one
-            // reader, so it needs to sleep some time to see if the reader done.
-            // Otherwise, it becomes a polling loop.
-            // TODO: use lock/sync-obj to leverage the concurrency APIs instead of explicit sleep.
-            Thread.sleep(100)
-          } else {
-            if (writeLock.tryLock(200, TimeUnit.MILLISECONDS)) {
-              try {
-                if (refCount == 0) {
-                  realDispose()
-                  return true
-                }
-              } finally {
-                writeLock.unlock()
-              }
-            }
-          }
-        }
-    }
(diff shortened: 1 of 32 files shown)


lee-lei:
For this case, the guava cache manager will directly free the evicted fiber if the refCount is zero already before putting in the queue. The code is in addRemovalFiber method in the first patch. Thanks.

lee-lei:
@carsonwang Please help review if the third commit resolves your concern when you are available. Thanks so much for your comments!

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
```diff
@@ -44,49 +51,19 @@ case class FiberCache(protected val fiberData: MemoryBlock) extends Logging {
...
     }
-    logWarning(s"Fiber Cache Dispose waiting detected for $fiber")
-    false
   }
 
-  protected var disposed = false
-  def isDisposed: Boolean = disposed
-  protected[filecache] def realDispose(fiber: Fiber): Unit = {
+  private var disposed: Boolean = false
+  def isDisposed(): Boolean = disposed
```
carsonwang: It is ok without () here because there is no side effect for this method.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
```diff
@@ -28,7 +29,7 @@ import org.apache.spark.unsafe.memory.MemoryBlock
 import org.apache.spark.unsafe.types.UTF8String
 
 // TODO: make it an alias of MemoryBlock
-case class FiberCache(protected val fiberData: MemoryBlock) extends Logging {
+case class FiberCache(var fiber: Fiber, protected val fiberData: MemoryBlock) extends Logging {
```
gczsjdy: For the changes of main interfaces/classes, please indicate your design and the necessity

gczsjdy:
You probably should indicate the direct evidence of why current cache guardian thread doesn't release fiber cache well(and as you say, cause memory pressure)

For your test part, some cases of `FiberCacheManagerSuite`'s producing `fibers pending on removal use too much memory` can't prove the inefficiency of `CacheGuardian`. So you can't prove your PR works by 

>   the test cases exceeding off-heap memory usage threshold are reduced from 6 to 2

Because they are made to overuse the memory in a sudden to let Guava perform evict, so it's likely the pending queue is temporarily full. And all of that happen in a rather short time.

In short, please
- Prove the evidence of memory pressure, maybe real cases and statistics (and it's caused by CacheGuardian)
- Use PR on the cases indicated above and give new results

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
lee-lei: For the last commit, there are primarily two changes. 
The first change is to add the fiber information to the corresponding fiber cache. The logic is that
we want to let the last user of fiber cache to release the fiber and free the memory together. Thus the fiber cache needs the information that it belongs to which fiber.
The second change is to refactor the test cases to use one unified apply function of FiberCache object. In this case, it's more clean.

lee-lei:
@gczsjdy For the unit test result, I updated in this PR. Thanks.

gczsjdy:
@lee-lei That's in extreme conditions of UT we created(like, probably occurs in 0.01% cases), however you still hasn't proved the memory pressure in non-extreme cases.

gczsjdy:
A complete logic chain proving the memory pressure + the result after resolving are needed

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
carsonwang: `fiber` should be a val ideally. Any plan to update this?

lee-lei:
@carsonwang please help review the last two commits when you are available. I am totally ok to split the last commit as a standalone PR after your review. Thanks.

gczsjdy:
@lee-lei maybe it's better to split this first, for the convenience of reviewers.
Also, we need confirmation from @linhong-intel

linhong-intel:
@gczsjdy this pr is ok to me

lee-lei:
Close this PR, since the problem resolved by this PR is very rare in the production environment, and the #830 proposed another solution. Thanks.

PR #803, opened 2018-07-02
Use ExternalSorter for Statistics while buiding index

gczsjdy:

To resolve #785 , in local test building B+ tree index for 3 * 10^7 rows, original takes 2.1min with 1.5min GC, with this PR it takes 1.8min with 58s GC. 

Current implementation has 1 not so graceful point: because Part-by-value & sample both need this sortedKeys, and one Iterator can only be traversed once, a duplicate ExternalSorter is created. 

Still trying to avoid creating 2...Another consideration is after the current sqlite version of sample PR is merged, we remove the original Part-by-value & sample code path, than this problem will resolve.

## How was this patch tested?

SampleBasedStatisticsSuite.scala, StatisticsManagerSuite.scala

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapIndexRecordWriter.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/BitmapIndexRecordWriter.scala
@@ -17,10 +17,9 @@
 
 package org.apache.spark.sql.execution.datasources.oap.index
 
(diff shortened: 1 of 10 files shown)


gczsjdy:
As synced offline, we will change to 1 sorter and merge this after #794 is merged

gczsjdy:
This may currently not be supported, due to the potential issue in PR description. : (

PR #818, opened 2018-07-25
Sample Statistics based on Sqlite Implementation

linhong-intel:

This is another sample statistics implementation. In light of Sqlite.
Different with current `SampleBasedStatistics`, this one contains more information with
less size. It looks like a combination of `SampleBasedStatistics` and `PartByValueStatistics`.

For now this statistics is not used yet.

## How was this patch tested?
current UT and one new UT passed

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/SqliteSampleStatistics.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/SqliteSampleStatistics.scala
@@ -37,9 +37,9 @@ import org.apache.spark.sql.types.StructType
  * nDlt: the number of distinct rows less than the sample row
  * Sample size is 1% of the total rows.
  * 1/3 of samples is selected from the rows evenly. called periodic sample
- * 2/3 of samples is the ones with greatest nEq. called "better" sample
+ * 2/3 of samples is the ones with greatest nEq. called normal sample
  */
-private[oap] case class Sample(key: Key, nEq: Int, nLt: Int, nDlt: Int, isPeriodic: Boolean)
+private[oap] case class Sample(key: Key, nEq: Int, nLt: Int, nDLt: Int, isPeriodic: Boolean)
 
 private[oap] class SampleStatisticsReader(schema: StructType) extends StatisticsReader(schema) {
   override val id: Int = StatisticsType.TYPE_SQLITE_SAMPLE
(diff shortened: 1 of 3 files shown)


linhong-intel:
Benchmark result
[Book1.xlsx](https://github.com/Intel-bigdata/OAP/files/2239376/Book1.xlsx)

linhong-intel:
Statistics size compare:
MinMax: 52 Bytes
Sample: 2752188 Bytes
Bloom: 131084 Bytes
Part: 104 Bytes
Sqlite: 1376092 Bytes

gczsjdy:
Thanks @linhong-intel , could we also add the key formulas(which is the core of sqlite's algorithm) to comments?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/SqliteSampleStatistics.scala
```diff
@@ -0,0 +1,241 @@
...
+        estimateKey(interval.start, isStart = true)._2
+    } else {
+      // Range
+      if (!intervalIsValid(interval)) {
+        0
+      } else {
+        val (iLowerLt, iLowerEq) = estimateKey(interval.start, isStart = true)
+        val (iUpperLt, iUpperEq) = estimateKey(interval.end, isStart = false)
+
+        val iLower = if (0 < iLowerLt + iLowerEq) iLowerLt + iLowerEq else 0
```
gczsjdy: just `val iLowerOrEqual = iLowerLt + iLowerEq`?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/statistics/SqliteSampleStatistics.scala
```diff
@@ -0,0 +1,241 @@
...
+        val iUpper = if (rowCount > iUpperLt + iLowerEq) iUpperLt + iUpperEq else rowCount
```
gczsjdy: Same with `val iGreaterOrEqual = UpperLt + iUpperEq`?

gczsjdy:
Thanks @linhong-intel for the great work! Some considerations and questions:
- Perhaps we can make it easier to maintain this by more comments for like `compareKeyWithInterval`, explaining the arguments, etc.
- About the `DUMMY_START/END`(Most headache part, yeah : )), currently we lack(commented?) the dealing code?
- Is multi-column index dealt well?
- In the implementation we sampled 1% of the data, why do we choose this percentage?

gczsjdy:
@linhong-intel Are we not going to support this?

PR #834, opened 2018-08-14
Add sparkSession to OapRuntime for easy control

gczsjdy:

Add sparkSession to OapRuntime for easy control. To easily get things like `sparkSession.conf`, don't need to pass conf everywhere.

Since we create `OapSession` in `OapEnv`, this PR gets `SparkSession`(base class of `OapSession`) from `OapEnv`.

## How was this patch tested?

OapRuntimeSuite

Manual test by: spark.sql("set whatever=666")
OapRuntime.getOrCreate.sparkSession.conf.get("whatever")
res0: String = 666


src/main/scala/org/apache/spark/sql/hive/thriftserver/OapEnv.scala
```diff
@@ -21,48 +21,56 @@ import java.io.PrintStream
...
 
-private[hive] object OapEnv extends Logging {
+private[sql] object OapEnv extends Logging {
   logDebug("Initializing Oap Env")
 
   var sqlContext: SQLContext = _
   var sparkContext: SparkContext = _
+  var sparkSession: SparkSession = _
+
+  var initialized = false
```
JkSelf: @gczsjdy what is the purpose of var initialized ?

src/main/scala/org/apache/spark/sql/hive/thriftserver/OapEnv.scala
```diff
@@ -21,48 +21,56 @@ import java.io.PrintStream
...
   def init() {
-    if (sqlContext == null) {
-      val sparkConf = new SparkConf(loadDefaults = true)
-      // If user doesn't specify the appName, we want to get [SparkSQL::localHostName] instead of
-      // the default appName [SparkSQLCLIDriver] in cli or beeline.
-      val maybeAppName = sparkConf
-        .getOption("spark.app.name")
-        .filterNot(_ == classOf[SparkSQLCLIDriver].getName)
+    // For unit tests, SparkSession is created in the testing framework
+    if (!initialized && !Utils.isTesting) {
```
JkSelf: @gczsjdy , here whether need synchronized this code to avoid multi thread call init method simultaneously ?

src/main/scala/org/apache/spark/sql/oap/OapRuntime.scala
```diff
@@ -44,12 +46,18 @@ private[oap] trait OapRuntime extends Logging {
 /**
  * Initializing [[FiberSensor]] and executor managers if local
  */
-private[oap] class OapDriverRuntime(sparkEnv: SparkEnv) extends OapRuntime {
+private[sql] class OapDriverRuntime(sparkEnv: SparkEnv) extends OapRuntime {
+  private var _sparkSession = { OapEnv.init(); OapEnv.sparkSession }
+  override def sparkSession: SparkSession = _sparkSession
+  // For Unit Test: setting TestOapSession
+  private[sql] def setTestSession(session: SparkSession): Unit = {
+    _sparkSession = session
```
JkSelf: @gczsjdy , whether set the session to sparkSession not _sparkSession?

gczsjdy: the former is a `def`

src/main/scala/org/apache/spark/sql/hive/thriftserver/OapEnv.scala
gczsjdy: Good catch! I will add it.

gczsjdy: Actually no synchronization will not create errors, because the upper function level has done it.
But it might be better to have it due to you can't rely on the caller.

I  have updated https://github.com/Intel-bigdata/OAP/pull/841 but not this, due to they have some overlap.

gczsjdy:
cc  @linhong-intel @carsonwang  @LuciferYang

src/main/spark2.1/scala/org/apache/spark/sql/hive/thriftserver/OapEnv.scala
```diff
@@ -77,6 +79,7 @@ private[spark] object OapEnv extends Logging {
   def initWithoutCreatingOapSession(): Unit = synchronized {
     if (!initialized && !Utils.isTesting) {
       val sc = SparkContext.getOrCreate()
+      this.sparkSession = SparkSession.getActiveSession.get
```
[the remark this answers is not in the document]
gczsjdy: Actually this method is called in non-SQL conditions, see `WithoutCreatingOapSession`, it indicates that a SparkSession shouldn't be created due to Spark has already done that. So if this `get` throws an Exception, there are real problems.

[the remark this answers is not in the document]
LuciferYang: got it ~, do we need to handle this exception ?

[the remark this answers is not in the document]
gczsjdy: I think we don't need to handle it by code, because normally we should have an existed `SparkSession` here.
If not, we should let it fail here.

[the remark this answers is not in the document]
LuciferYang: ok ~

src/main/scala/org/apache/spark/sql/oap/OapRuntime.scala
```diff
@@ -45,19 +48,25 @@ private[oap] trait OapRuntime extends Logging {
...
   // is created. For example, OAP UI tab will show at the moment you read a Parquet file to OAP
   // cache
   OapEnv.initWithoutCreatingOapSession()
 
+  private var _sparkSession = OapEnv.sparkSession
+  override def sparkSession: SparkSession = _sparkSession
+  // For Unit Test: setting TestOapSession
+  private[sql] def setTestSession(session: SparkSession): Unit = {
+    _sparkSession = session
+  }
```
linhong-intel: This looks strange.
After you `setTestSession`, it can never restore back to "normal" sparkSession.
I don't think it's a good idea to mess OapRuntime with test code.
You may want a TestOapRuntime, or TestOapEnv

linhong-intel:
lgtm.
left a comment about the test code (not a big deal)

gczsjdy:
Thanks much @linhong-intel , it makes sense.

How to make `TestOapRuntime` `TestOapEnv` solve this? 
I think we shouldn't use another `Runtime`, due to existed OAP logic are written in `OapRuntime.xx`.

The main issue here is
- During UT we created a TestSession: ts (locally)
- OapRuntime is an Object and can be globally accessed
If the latter needs to reference to the right former, seems we can only call `OapRuntime.setTestSession` from where the TestSession is created?

gczsjdy:
Let's merge this first so as not to block the PRs depending on it. Thanks @linhong-intel @LuciferYang @JkSelf

PR #839, opened 2018-08-17

ConeyLiu:

Abstract the MemoryManager to support more memory type in future(such as Intel Optane Data Center Persistent Memory).

## How was this patch tested?

`mvn clean test`


src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
```diff
@@ -141,12 +144,16 @@ case class FiberCache(protected val fiberData: MemoryBlock) extends Logging {
   }
 
   def size(): Long = fiberData.size()
+
+  // Usable size maybe different the size, it should be the actual occupied size of memory.
```
winningsix: Rewording to "Return the allocated size and it's typically larger than the required data size due to memory alignments from underlying allocator"

PR #842, opened 2018-08-20
spark 2.3 support - import spark 2.3 source code

zhixingheyi-tian:
Import Spark source code from spark 2.3.0. No code changes in this PR.
This will be helpful to review changes we make in OAP in follow up PRs.
The following three code files are added to the OAP project compared to the imported source code for the spark 2.2 version.

OnHeapColumnVector.java
Added interface method: getChildColumn(int ordinal) to be used by OAP related code.
In spark 2.3, the ColumnVector class and its subclasses change too much. Member attributes and locations vary greatly. Some methods have been removed.

OneApplicationResource.scala
Add an http service method to be used by the OAP UI.

SaveAsHiveFile.scala
The write method in the FileFormatWriter class of spark adds the return parameter Seq[WriteResult], which affects the use of spark's saveAsHiveFile class. Therefore, change the use of FileFormatWriter.write in SaveAsHiveFile.

In addition:
Spark 2.2 source code ApiRootResource.scala, was not imported into OAP's version of spark 2.3.
The code content that ApiRootResource.scala needs to add is reflected in OneApplicationResource.scala. This is because the http service code architecture of spark 2.3 has been greatly adjusted.


gczsjdy:
Thanks @zhixingheyi-tian , I remembered you mentioned `InsertIntoHadoopFsRelationCommand.scala` in OAP is actually the same with Spark code. We can double check on that
- If yes, please remove it all in OAP source
- if no, please track it in issues

Also, remember to update the test section of PR description. It's good to take use of the template.

jerrychenhf:
There are many files needs to be maintained for different versions. Let's keep this method in OAP 0.5 and we can revisit whether the multiple version problem can be simplified.

To support ORC format, Li Lei also added some classes from Spark 2.3 including ColumnBatch, ColumnVector, OnHeapColumnVector. Not sure whether this makes things more complicated. Let's discuss.

zhixingheyi-tian:
@carsonwang 
OnHeapColumnVector does not need to be imported in now. OAP needs to use the method " getChildColumn(int ordinal) ", which can be replaced by "getChild(int ordinal)"

carsonwang:
Thanks for the investigation. Please remove the file in this pr.

carsonwang:
Thanks @zhixingheyi-tian . Merging this.

gczsjdy:
@zhixingheyi-tian Could you create issues to track the 2 added Spark code using label 
Spark API change: 2.3 ?

gczsjdy:
Also, any update for `InsertIntoHadoopFsRelationCommand`?

PR #844, opened 2018-08-20

ConeyLiu:

* Abstract MemoryManager to support more memory type
* Add `NativeLoader` to load a native library.

## How was this patch tested?

Manual test.

### DIFF
--- a/src/main/java/org/apache/spark/unsafe/NativeLoader.java
+++ b/src/main/java/org/apache/spark/unsafe/NativeLoader.java
@@ -1,0 +1,119 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.unsafe;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.*;
+
+public class NativeLoader {
+  private static final Logger logger = LoggerFactory.getLogger(NativeLoader.class);
+
+  public static void loadLibrary(String libName) {
+    assertOsArchSupport();
+    try {
+      logger.info("Try to load library {} from system library path.", libName);
+      System.loadLibrary(libName);
+      return;
+    } catch (UnsatisfiedLinkError e) {
+      logger.warn("Can't load library {} from system library path.", libName);
+    }
+
+    logger.info("Try to load library {} from package.", libName);
+    loadFromPackage(libName);
+  }
+
+  private static void loadFromPackage(String libName) {
+    String fullName = appendPrefixAndSuffix(libName);
+    String path = "/lib/linux64/" + fullName;
+    InputStream input = NativeLoader.class.getResourceAsStream(path);
+    if (input == null) {
+      throw new RuntimeException("The library " + path + " doesn't exist");
+    }
+
+    File tmpFile = null;
+    OutputStream output = null;
+    try {
+      tmpFile = File.createTempFile("lib", libName + ".so.tmp");
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+
+    try {
+      output = new FileOutputStream(tmpFile);
+      byte[] buffer = new byte[1024];
+      int len = -1;
+      while ((len = input.read(buffer)) != -1) {
+        output.write(buffer, 0, len);
+      }
+
+      try {
+        output.flush();
+        output.close();
+      } catch (Exception e) {
+        // ignore it
+      }
+
+      System.load(tmpFile.getCanonicalPath());
+    } catch (FileNotFoundException e) {
+      throw new RuntimeException(e);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    } catch (UnsatisfiedLinkError e) {
+      throw new RuntimeException(e);
+    } finally {
+      if (input != null) {
+        try {
+          input.close();
+          input = null;
+        } catch (Exception e) {
+          // ignore it
+        }
+      }
+
+      if (output != null) {
+        try {
+          output.close();
+          output = null;
+        } catch (Exception e) {
+          // ignore it
+        }
+      }
+
+      if (tmpFile != null && tmpFile.exists()) {
+        tmpFile.delete();
+        tmpFile = null;
+      }
+    }
+  }
+
+  private static void assertOsArchSupport() {
+    String osProp = System.getProperty("os.name");
+    String archProp = System.getProperty("os.arch");
+    if (!osProp.contains("Linux") && !archProp.contains("64")) {
+      throw new UnsupportedOperationException("Only linux64 is supported. It doesn't support on "
+                                                + osProp + archProp + "currently");
+    }
+  }
+
+  private static String appendPrefixAndSuffix(String libName) {
+    // Currently, we only support linux64
+    return "lib" + libName + ".so";
+  }
+}
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
@@ -145,7 +145,8 @@ case class FiberCache(
 
   def size(): Long = fiberData.size()
 
-  // Usable size maybe different the size, it should be the actual occupied size of memory.
+  // Return the allocated size and it's typically larger than the required data size due to memory
+  // alignments from underlying allocator
   def getUsableSize(): Long = usableSize
 }
 
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
@@ -27,7 +27,10 @@ import org.apache.spark.unsafe.Platform
 import org.apache.spark.unsafe.memory.MemoryBlock
 import org.apache.spark.unsafe.types.UTF8String
 
-case class FiberCache(protected val fiberData: MemoryBlock) extends Logging {
+// TODO: Store usableSize in MemoryBlock would be better.
+case class FiberCache(
+    protected val fiberData: MemoryBlock,
+    private val usableSize: Long) extends Logging {
 
   // This is and only is set in `cache() of OapCache`
   // TODO: make it immutable
@@ -141,12 +144,16 @@ case class FiberCache(protected val fiberData: MemoryBlock) extends Logging {
   }
 
   def size(): Long = fiberData.size()
+
+  // Usable size maybe different the size, it should be the actual occupied size of memory.
+  def getUsableSize(): Long = usableSize
 }
 
 object FiberCache {
   //  For test purpose :convert Array[Byte] to FiberCache
   private[oap] def apply(data: Array[Byte]): FiberCache = {
     val memoryBlock = new MemoryBlock(data, Platform.BYTE_ARRAY_OFFSET, data.length)
-    FiberCache(memoryBlock)
+    // We store the length of the array as the usable size.
+    FiberCache(memoryBlock, data.length)
   }
 }
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
@@ -64,23 +64,48 @@ private[sql] abstract class MemoryManager {
    */
   protected def getUsableSize(address: Long, requestedSize: Long): Long
 
+  @inline protected def toFiberCache(bytes: Array[Byte]): FiberCache = {
+    val memoryBlock = allocate(bytes.length)
+    Platform.copyMemory(
+      bytes,
+      Platform.BYTE_ARRAY_OFFSET,
+      memoryBlock.getBaseObject,
+      memoryBlock.getBaseOffset,
+      bytes.length)
+    val usableSize = getUsableSize(memoryBlock.getBaseOffset, bytes.length)
+    FiberCache(memoryBlock, usableSize)
+  }
+
   /**
    * Used by IndexFile
    */
-  def toIndexFiberCache(in: FSDataInputStream, position: Long, length: Int): FiberCache
+  def toIndexFiberCache(in: FSDataInputStream, position: Long, length: Int): FiberCache = {
+    val bytes = new Array[Byte](length)
+    in.readFully(position, bytes)
+    toFiberCache(bytes)
+  }
 
   /**
    * Used by IndexFile. For decompressed data
    */
-  def toIndexFiberCache(bytes: Array[Byte]): FiberCache
+  def toIndexFiberCache(bytes: Array[Byte]): FiberCache = {
+    toFiberCache(bytes)
+  }
 
   /**
    * Used by OapDataFile since we need to parse the raw data in on-heap memory before put it into
    * off-heap memory
    */
-  def toDataFiberCache(bytes: Array[Byte]): FiberCache
-  def getEmptyDataFiberCache(length: Long): FiberCache
-  def stop(): Unit
+  def toDataFiberCache(bytes: Array[Byte]): FiberCache = {
+    toFiberCache(bytes)
+  }
+
+  def getEmptyDataFiberCache(length: Long): FiberCache = {
+    val memoryBlock = allocate(length)
+    val usableSize = getUsableSize(memoryBlock.getBaseOffset, memoryBlock.size())
+    FiberCache(memoryBlock, usableSize)
+  }
+  def stop(): Unit = {}
 }
 
 private[sql] object MemoryManager {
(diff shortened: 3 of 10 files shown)


ConeyLiu:
cc @winningsix

ConeyLiu:
#844

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -30,12 +30,114 @@ import org.apache.spark.storage.{BlockManager, TestBlockId}
...
+   * with the requested size, that's depends on the underlying implementation.
+   * @param size requested size of memory block
+   */
+  private[filecache] def allocate(size: Long): MemoryBlock
+  private[filecache] def free(block: MemoryBlock): Unit
+
+  /**
+   * Get the actual occupied size of requested memory. The meaning should be equal with
+   * 'malloc.malloc_usable_size'. This method is valuable, because we may allocate memory based
+   * on different implementation (eg: jemalloc).
```
gczsjdy: Could you illustrate this more?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -63,71 +165,35 @@ private[sql] class MemoryManager(sparkEnv: SparkEnv) extends Logging {
...
-    Platform.copyMemory(
-      bytes,
-      Platform.BYTE_ARRAY_OFFSET,
-      memoryBlock.getBaseObject,
-      memoryBlock.getBaseOffset,
-      bytes.length)
-    FiberCache(memoryBlock)
+  override private[filecache] def free(block: MemoryBlock): Unit = {
+    // TODO: get the usable size from the parameter? However, it's a bit ugly.
+    val occupiedSize = getUsableSize(block.getBaseOffset, block.size())
```
gczsjdy: This is in the subclass, so we don't need to call `getUsableSize`, the original way is ok

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -30,12 +30,114 @@ import org.apache.spark.storage.{BlockManager, TestBlockId}
...
+private[sql] abstract class MemoryManager {
+  /**
+   * Return the total memory used until now.
+   */
+  def memoryUsed: Long
+
+  /**
+   * The memory size used for cache.
+   */
+  def getCacheMemorySize: Long
```
gczsjdy: I think we don't need to follow java fashion `getxxx` here, the original naming is fine.
Also, maybe we don't need the comments here and above, the naming is already clear.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -136,8 +136,10 @@ class GuavaOapCache(cacheMemory: Long, cacheGuardianMemory: Long) extends OapCac
   }
 
   private val weigher = new Weigher[FiberId, FiberCache] {
-    override def weigh(key: FiberId, value: FiberCache): Int =
-      math.ceil(value.size() / KB).toInt
+    override def weigh(key: FiberId, value: FiberCache): Int = {
+      // We should calculate the weigh with the usable size of the block.
+      math.ceil(value.getUsableSize() / KB).toInt
```
gczsjdy: Will the actually used size stay consistent? 
For example, when a `FiberCache` is constructed, it takes Xkb, but will the alignment become worse? Then it actually takes 2Xkb.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
```diff
@@ -27,7 +27,10 @@ import org.apache.spark.unsafe.Platform
 import org.apache.spark.unsafe.memory.MemoryBlock
 import org.apache.spark.unsafe.types.UTF8String
 
-case class FiberCache(protected val fiberData: MemoryBlock) extends Logging {
+// TODO: Store usableSize in MemoryBlock would be better.
+case class FiberCache(
+    protected val fiberData: MemoryBlock,
+    private val usableSize: Long) extends Logging {
```
gczsjdy: We should merge it with size(dataSize), I think we only need to care about the former one.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -30,12 +30,114 @@ import org.apache.spark.storage.{BlockManager, TestBlockId}
...
+   * Get the actual occupied size of requested memory. The meaning should be equal with
+   * 'malloc.malloc_usable_size'. This method is valuable, because we may allocate memory based
+   * on different implementation (eg: jemalloc).
+   * @param address the address of the block.
+   * @param requestedSize the requested size of the block.
+   * @return the actual occupied size.
+   */
+  protected def getUsableSize(address: Long, requestedSize: Long): Long
+
+  @inline protected def toFiberCache(bytes: Array[Byte]): FiberCache = {
```
gczsjdy: Do we need this annotation?

src/main/java/org/apache/spark/unsafe/NativeLoader.java
```diff
@@ -0,0 +1,119 @@
+/*
```
gczsjdy: Dead code?

src/main/java/org/apache/spark/unsafe/NativeLoader.java
```diff
@@ -0,0 +1,119 @@
...
+      logger.warn("Can't load library {} from system library path.", libName);
+    }
+
+    logger.info("Try to load library {} from package.", libName);
+    loadFromPackage(libName);
+  }
+
+  private static void loadFromPackage(String libName) {
+    String fullName = appendPrefixAndSuffix(libName);
+    String path = "/lib/linux64/" + fullName;
```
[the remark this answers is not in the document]
ConeyLiu: This is a path under the `resources` for the native library. The native library will be uploaded in the following patches. There don't config for the path.

src/main/java/org/apache/spark/unsafe/NativeLoader.java
```diff
@@ -0,0 +1,119 @@
...
+          // ignore it
+        }
+      }
+
+      if (output != null) {
+        try {
+          output.close();
+          output = null;
+        } catch (Exception e) {
+          // ignore it
```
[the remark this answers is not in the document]
ConeyLiu: Not sure, `outputStream.close` throws `Exception`. We need to catch it, but don't need care about it when failed again in this scope. There would be better add a log warning.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
```diff
@@ -141,12 +144,17 @@ case class FiberCache(protected val fiberData: MemoryBlock) extends Logging {
   }
 
   def size(): Long = fiberData.size()
+
+  // Return the allocated size and it's typically larger than the required data size due to memory
+  // alignments from underlying allocator
+  def getUsableSize(): Long = usableSize
```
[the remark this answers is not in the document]
ConeyLiu: I think the `usableSize` should be removed from the field of `FiberCache`. We need rewrite the `MemoryBlock` to support that.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
ConeyLiu: Same meaning with [`malloc.get_usable_size`](http://man7.org/linux/man-pages/man3/malloc_usable_size.3.html)

src/main/java/org/apache/spark/unsafe/NativeLoader.java
ConeyLiu: Used to load a native library.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
ConeyLiu: The `size` and `usableSize` are different, `usabeSize` should be greater or equal to size. The `usableSize` would be better stored in `MemoryBlock`. At present, this practice is only a compromise.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
ConeyLiu: Here, the function is defined in the abstract class. I just think the function should name with a verb. Very small problem, no problem changing back...

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
ConeyLiu: This method is called very often. Add the annotation just suggest JVM to inline it which can reduce once function call.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
ConeyLiu: Not very sure we need record the `memoryUsed` with `requested size` or `usable size`. The `requested size` shows the user request. However, the `usable size` is the actual occupied size of the cache.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
ConeyLiu: Yeah, you are right. The requested size not meaning the occupied size. That depends on the memory allocator implement.

src/main/java/org/apache/spark/unsafe/NativeLoader.java
```diff
@@ -0,0 +1,119 @@
...
+
+package org.apache.spark.unsafe;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.*;
+
+public class NativeLoader {
+  private static final Logger logger = LoggerFactory.getLogger(NativeLoader.class);
```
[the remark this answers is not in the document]
gczsjdy: ok

src/main/java/org/apache/spark/unsafe/NativeLoader.java
gczsjdy: I meant it's better to have this with your persistent memory PR. And name it to `PersistenMemoryNativeLoader`?

src/main/java/org/apache/spark/unsafe/NativeLoader.java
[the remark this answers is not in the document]
gczsjdy: ok

src/main/java/org/apache/spark/unsafe/NativeLoader.java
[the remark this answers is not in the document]
gczsjdy: Could we at least remove the if (!isNull)? This is too c and not necessary.
For the try-close-catch, small suggestion:
- A function to wrap this, small overhead and less blocking reading...

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
gczsjdy: For DRAM, they are the same value, right?
For PM, usable >= size, but when do we care about the `size`?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
gczsjdy: Maybe I think too much, I think it's too Java... Scala has not many `getxx`
Let's use the original naming.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
gczsjdy: Interesting, out of curiosity, then does it have side effect?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
gczsjdy: I mean does the 'actually occupied size' change(both for PM, time 1 & time2)?

src/main/java/org/apache/spark/unsafe/NativeLoader.java
```diff
@@ -0,0 +1,119 @@
...
+          input = null;
+        } catch (Exception e) {
+          // ignore it
+        }
+      }
+
+      if (output != null) {
+        try {
+          output.close();
+          output = null;
```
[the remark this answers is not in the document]
gczsjdy: ok but let us remove it for code readers : )

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
jerrychenhf: I suggest to rename "usableSize" to "capacity" just like a Array implementation. size is the number of actual element and capacity is the space available to put elements.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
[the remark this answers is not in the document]
jerrychenhf: Agree. It would be nice for "capacity" to be in MemoryBlock

src/main/java/org/apache/spark/unsafe/NativeLoader.java
winningsix: Yes, it's related to persistent memory patch. The major reason we put it in pre-pull-request is to simplify the complexity of AEP enabling since this native loader shall be common use.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
winningsix: I think it should be the same side effect of inline method in C++. (e.g. put a very complex logic as an inline method).

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
ConeyLiu: OK

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
ConeyLiu: no problem to change back

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
ConeyLiu: maybe talk offline

src/main/java/org/apache/spark/unsafe/NativeLoader.java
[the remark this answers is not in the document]
LuciferYang: Good habit~ but `JIT` will do this ·set null· to optimize GC automatically

src/main/java/org/apache/spark/unsafe/NativeLoader.java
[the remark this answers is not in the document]
LuciferYang: `org.apache.hadoop.io.IOUtils.close` or `org.apache.commons.io.IOUtils.closeQuietly`, there are many Utils to use.

src/main/java/org/apache/spark/unsafe/NativeLoader.java
```diff
@@ -0,0 +1,119 @@
...
+        // ignore it
+      }
+
+      System.load(tmpFile.getCanonicalPath());
+    } catch (FileNotFoundException e) {
+      throw new RuntimeException(e);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    } catch (UnsatisfiedLinkError e) {
+      throw new RuntimeException(e);
```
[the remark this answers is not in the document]
LuciferYang: should we `Fast failure` if load from ld lib path and loadFromPackage both failed？Is there other fault tolerance mechanism after load failed？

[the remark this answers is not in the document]
ConeyLiu: No other mechanism to process. The executor process should be exited due to the exception. You mean `System.exit()` ?

jerrychenhf:
@gczsjdy @carsonwang Xianyang mentioned to have address the comments. Please have a look and if there is no major issue, @carsonwang please help commit. 
We  are speeding up the process for being able to kick off the evaluation in Baidu in the second or third week of Sep

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -30,12 +30,114 @@ import org.apache.spark.storage.{BlockManager, TestBlockId}
...
+   * The memory size used for cache guardian.
+   */
+  def cacheGuardianMemory: Long
+
+  /**
+   * Allocate a block of memory with given size. The actual occupied size of memory maybe different
+   * with the requested size, that's depends on the underlying implementation.
+   * @param size requested size of memory block
+   */
+  private[filecache] def allocate(size: Long): MemoryBlock
```
carsonwang: Will it look better by returning a new class like xxxMemoryBlock which wraps the MemoryBlock and capacity in it? getCapacity can be removed then.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
```diff
@@ -27,7 +27,15 @@ import org.apache.spark.unsafe.Platform
 import org.apache.spark.unsafe.memory.MemoryBlock
 import org.apache.spark.unsafe.types.UTF8String
 
-case class FiberCache(protected val fiberData: MemoryBlock) extends Logging {
+/**
+ * Basic unit of OAP cache.
+ * TODO: Store capacity in MemoryBlock would be better.
+ * @param fiberData the row memory block which store the data
```
gczsjdy: :nit raw

carsonwang:
Just noticed this was not merged to master. Why did this target to another branch? @ConeyLiu

ConeyLiu:
#863

gczsjdy:
Same confusion with @carsonwang , why do we need a `persistentmemory` branch?

ConeyLiu:
Putting it on a separate branch as discussed before. I'll submit to master and delete the extra branch.

Issue #838, opened 2018-08-17
Abstract MemoryManager to support more memory type


ConeyLiu:
Abstract the `MemoryManager` to support more memory type(such as Intel Optane Data Center Persistent Memory).

ConeyLiu:
#844

PR #863, opened 2018-09-03

ConeyLiu:

What changes were proposed in this pull request?
Issue track #838

## How was this patch tested?

Manual test.


ConeyLiu:
#844

src/main/java/org/apache/spark/unsafe/NativeLoader.java
```diff
@@ -0,0 +1,111 @@
...
+      logger.warn("Can't load library {} from system library path.", libName);
+    }
+
+    logger.info("Try to load library {} from package.", libName);
+    loadFromPackage(libName);
+  }
+
+  private static void loadFromPackage(String libName) {
+    String fullName = appendPrefixAndSuffix(libName);
+    String path = "/lib/linux64/" + fullName;
```
winningsix: Could we make the path as a static variable?

src/main/java/org/apache/spark/unsafe/NativeLoader.java
```diff
@@ -0,0 +1,111 @@
...
+      int len = -1;
+      while ((len = input.read(buffer)) != -1) {
+        output.write(buffer, 0, len);
+      }
+
+      try {
+        output.flush();
+        output.close();
+      } catch (Exception e) {
+        // ignore it
```
winningsix: LOG warning here please.

src/main/java/org/apache/spark/unsafe/NativeLoader.java
```diff
@@ -0,0 +1,111 @@
...
+      }
+
+      System.load(tmpFile.getCanonicalPath());
+    } catch (FileNotFoundException e) {
```
winningsix: Since no specified handling for each exception, why not catch with Throwable exception here?

src/main/java/org/apache/spark/unsafe/NativeLoader.java
ConeyLiu: I'll remove the `NativeLoader.java`, and using [Native-lib-loader](https://github.com/scijava/native-lib-loader)

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
```diff
@@ -97,7 +99,8 @@ case class FiberCache(protected val fiberData: MemoryBlock) extends Logging {
   // For debugging
   def toArray: Array[Byte] = {
     // TODO: Handle overflow
-    val bytes = new Array[Byte](fiberData.size().toInt)
+    val intSize = Ints.checkedCast(size())
```
gczsjdy: Good check!

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
```diff
@@ -140,13 +143,19 @@ case class FiberCache(protected val fiberData: MemoryBlock) extends Logging {
       getBaseObj, getBaseOffset + offset, dst, Platform.BYTE_ARRAY_OFFSET, dst.length)
   }
 
-  def size(): Long = fiberData.size()
+  def size(): Long = fiberData.memoryBlock.size()
+
+  // Return the allocated size and it's typically larger than the required data size due to memory
+  // alignments from underlying allocator
+  def getCapacity(): Long = fiberData.capacity
```
gczsjdy: Actually capacity is 
> the amount that can be held or produced by something

But it's not the meaning we want, I think `occupiedSize` may be better

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
```diff
@@ -140,13 +143,19 @@ case class FiberCache(protected val fiberData: MemoryBlock) extends Logging {
...
 }
 
 object FiberCache {
   //  For test purpose :convert Array[Byte] to FiberCache
   private[oap] def apply(data: Array[Byte]): FiberCache = {
     val memoryBlock = new MemoryBlock(data, Platform.BYTE_ARRAY_OFFSET, data.length)
-    FiberCache(memoryBlock)
+    // We store the length of the array as the capacity.
```
gczsjdy: We don't need this? It's just the code says... : )

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
gczsjdy: `requiredSize` & `occupiedSize` should distinguish the two semantics?

src/test/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManagerSuite.scala
```diff
@@ -82,12 +82,16 @@ class MemoryManagerSuite extends SharedOapContext {
...
+      val fiberId = TestFiberId(null, "Test fiber for MemoryManagerSuite#0")
+      val cache = memoryManager.toDataFiberCache(buf.toByteArray)
+      cache.fiberId = fiberId
+      cache
     }
   }
 
   protected override def afterAll(): Unit = {
     super.afterAll()
+    fiberCache.realDispose()
```
gczsjdy: Thanks ! : )

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
ConeyLiu: OK, `occupiedSize`

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
ConeyLiu: For DRAM, we consider the requested size as occupied size. Here, the requested size should be the length of the Array[Byte]

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -31,11 +31,107 @@ import org.apache.spark.unsafe.Platform
...
 /**
- * Memory Manager
- *
- * Acquire fixed amount of memory from spark during initialization.
+ * This just wrap the memory block with capacity field. The capacity used to get the actual
+ * capacity of the memoryBlock. For DRAM OFF_HEAP memory, this should be same as the block size.
+ * For other types of memory, that's depends on the implementation of memory allocation.
  */
-private[sql] class MemoryManager(sparkEnv: SparkEnv) extends Logging {
+private[filecache] case class MemoryBlockWithCapacity(memoryBlock: MemoryBlock, capacity: Long)
```
winningsix: Could we just extend MemoryBlock here by adding capacity as a parameter?

ConeyLiu: We can do that, but that means the new class need depend on the interface of `MemoryBlock`. In Spark 2.3, there are a big refactor of MemoryBlock. And also, the combination is better than inheritance from the perspective of design patterns.

winningsix: To my understanding, `MemoryBlockWithCapacity` is a kind of extension of `MemoryBlock` with the capacity attribution. By doing it, we can simplify most points without different handling logic with  `MemoryBlockWithCapacity` and `MemoryBlock`.

ConeyLiu:
If no other comments, I'm going to merge this patch after test finished.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
ConeyLiu: We need do the different implementation for the different version of Spark if doing that.

ConeyLiu:
Merge this, further comments will be addressed in following patches.

PR #865, opened 2018-09-05
Add support Intel Optane DC persistent memory cache

ConeyLiu:

Finished:
- Add new memory manager based on Intel Optane DC persistent memory.
- Add NativeLibrarayLoader to load the native library.
- Add new conf/persistent-memory.xml.template to describe the persistent memory config.
- CMakeLists.txt used to make native code.
- new pom profile to active native code compile and produce share library.

Compile & Package:
`mvn clean package -Pspark-2.1 -Ppersistent-memory`

TODO:

- [x]  Maybe need modify the tracis-ci to support the compile.
- [ ]  How to add test case ?

## How was this patch tested?

Manually local test.


src/main/java/org/apache/spark/util/NativeLibraryLoader.java
```diff
@@ -0,0 +1,121 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.*;
+
+/**
+ * A native library loader which used to load persistent memory native library.
+ */
+public class NativeLibraryLoader {
```
ConeyLiu: [native-lib-loader](https://github.com/scijava/native-lib-loader) seems not a good device.

ConeyLiu:
cc @carsonwang , @gczsjdy , @jerrychenhf, @winningsix, pls help to review in advance. I'll finish the code and locally test by the end of today. Thanks a lot.

conf/persistent-memory.xml.template
```diff
@@ -0,0 +1,35 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+
+<persistentMemoryPool>
+  <!--The numa id-->
+  <numanode id="0">
+    <!--The initial path for Intel Optane DC persistent memory-->
+    <initialPath>/mnt/pmem0/spark</initialPath>
```
winningsix: Hint: "/mnt/pmem0/spark" is more likely an example. Please add it an example and make it empty in case users mis-configure it by using the default file.

conf/persistent-memory.xml.template
```diff
@@ -0,0 +1,35 @@
...
+    <!--The initial size for the given path-->
+    <initialSize>100g</initialSize>
```
winningsix: The same as above.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -25,19 +26,26 @@ import org.apache.spark.SparkEnv
...
+ * type of memory, the field of occupiedSize should be same as the block length. For Intel Optane
+ * DC persistent memory, the occupied size is typically larger than length because the
+ * heap management is based on the jemalloc.
+ * @param baseObject null for OFF_HEAP or Intel Optane DC persistent memory
+ * @param length the requested size of the block
+ * @param occupiedSize the actual occupied size of the memory block
  */
-private[filecache] case class MemoryBlockWithOccupiedSize(
-    memoryBlock: MemoryBlock, occupiedSize: Long)
+private[filecache] case class MemoryBlockHolder(
```
ConeyLiu: Seems we don't need to rely on Spark's `MemoryBlock`. In Spark, all the OFF_HEAP storage are operated by the raw address and the length of the block. The `MemoryBlock` and the feature of `MemoryBlock` are only used in computation memory.

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
```diff
@@ -0,0 +1,121 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.spark.util;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.*;
```
winningsix: Import only needed classes please.

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        if (tmpLib != null && !tmpLib.delete()) {
+          logger.warn("Delete tmp lib: {} failed, ignore it.", tmpLib.getAbsolutePath());
+        }
+      }
+    }
+  }
+
+  private static void closeWithWarn(Closeable closeable, String warnMsg) {
```
winningsix: Should be `safeClose` method? It will throw no warning if close successfully.

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
```diff
@@ -0,0 +1,121 @@
...
+          int read = is.read(buf);
+          if (read == -1) {
+            break;
+          }
+          out.write(buf, 0, read);
+        }
+
+        out.flush();
+        out.close();
+        out = null;
```
winningsix: Not needed.

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
```diff
@@ -0,0 +1,121 @@
...
+        while (true) {
+          int read = is.read(buf);
+          if (read == -1) {
+            break;
+          }
+          out.write(buf, 0, read);
+        }
+
+        out.flush();
+        out.close();
```
winningsix: Should put to final block.

jerrychenhf: @ConeyLiu  since you close the out stream in finally, we don't need close call here. 

if(out != null ) {
  closeWithWarn(out, ...)
}

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
```diff
@@ -0,0 +1,121 @@
...
+        out = null;
+        loaded = true;
```
jerrychenhf: I don't see the code the load library from the temporary lib file. Then loaded is set to true.
Do we miss something here?

src/main/native/org_apache_spark_unsafe_PMPlatform.cpp
```diff
@@ -0,0 +1,97 @@
...
+JNIEXPORT jlong JNICALL Java_org_apache_spark_unsafe_PMPlatform_getUsableSize
+  (JNIEnv *env, jclass clazz, jlong address) {
+  check(env);
+  void *p = addr_from_java(address);
+  return memkind_malloc_usable_size(pmemkind, p);
+}
+
+JNIEXPORT void JNICALL Java_org_apache_spark_unsafe_PMPlatform_freeMemory
+  (JNIEnv *env, jclass clazz, jlong address) {
+  if (NULL == memkind) {
```
jerrychenhf: Do we really need the memkind == NULL check here?
As I don't see the same check for allocateMemory function implementation,  I have this question.
Please consistently check or not check.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -168,17 +177,18 @@ private[filecache] class OffHeapMemoryManager(sparkEnv: SparkEnv)
...
-  override private[filecache] def allocate(size: Long): MemoryBlockWithOccupiedSize = {
-    val block = MemoryAllocator.UNSAFE.allocate(size)
+  override private[filecache] def allocate(size: Long): MemoryBlockHolder = {
+    val address = Platform.allocateMemory(size)
     _memoryUsed.getAndAdd(size)
     logDebug(s"request allocate $size memory, actual occupied size: " +
       s"${size}, used: $memoryUsed")
     // For OFF_HEAP, occupied size also equal to the size.
-    MemoryBlockWithOccupiedSize(block, size)
+    MemoryBlockHolder(null, address, size, size)
```
jerrychenhf: shall "address" pass as the first parameter as baseObject here?
You pass "null" for baseObject.

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
ConeyLiu: need import too many.

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
ConeyLiu: sure

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
ConeyLiu: we need it, to identify the out is closed or not.

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
ConeyLiu: We should flush the data to tmpLib before we load it.

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
ConeyLiu: sure, will fix it

src/main/native/org_apache_spark_unsafe_PMPlatform.cpp
```diff
@@ -0,0 +1,97 @@
...
+  // Initialize persistent memory
+  int error = memkind_create_pmem(str, sz, &pmemkind);
+  if (error) {
+    jclass exceptionCls = env->FindClass("java/lang/Exception");
+    env->ThrowNew(exceptionCls,
+      "Persistent initialize failed! Please check the path permission.");
+  }
+
+  env->ReleaseStringUTFChars(path, str);
+}
```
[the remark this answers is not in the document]
ConeyLiu: The Exception will be catched in java level, and the clear error message will be included and throw a `ExceptionInInitializerError`

src/main/java/org/apache/spark/unsafe/PMPlatform.java
```diff
@@ -0,0 +1,83 @@
...
+          "initial path should be a directory");
+        Preconditions.checkArgument(size > 0,
+          "Persistent memory initial size must be a positive number");
+        try {
+          initializeNative(path, size);
+        } catch (Exception e) {
+          throw new ExceptionInInitializerError("Persistent memory initialize (path: " + path +
+            ", size: " + size + ") failed. Please check the path permission and initial size.");
+        }
+        initialized = true;
```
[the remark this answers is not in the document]
ConeyLiu: The info is print in the called of `initialize`

src/main/native/org_apache_spark_unsafe_PMPlatform.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+JNIEXPORT jlong JNICALL Java_org_apache_spark_unsafe_PMPlatform_allocateMemory
+  (JNIEnv *env, jclass clazz, jlong size) {
+  check(env);
+
+  size_t sz = (size_t)size;
+  void *p = memkind_malloc(pmemkind, sz);
+  if (p == null) {
+    jclass errorCls = env->FindClass("java/lang/OutOfMemoryError");
+    env->ThrowNew(errorCls,
```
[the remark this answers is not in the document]
ConeyLiu: will add it

src/main/native/org_apache_spark_unsafe_PMPlatform.cpp
ConeyLiu: Others are checked will method `check`, will change to the same.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -187,3 +197,62 @@ private[filecache] class OffHeapMemoryManager(sparkEnv: SparkEnv)
...
+    // The NUMA id should be set when the executor process start up. However, Spark don't
+    // support NUMA binding currently.
+    var numaId = conf.getInt("spark.executor.numa.id", -1)
+    val executorId = sparkEnv.executorId.toInt
+    val map = XmlUtils.parseConfig(conf)
+    if (numaId == -1) {
+      logWarning(s"Executor ${executorId} is not bind with NUMA. It would be better to bind " +
+        s"executor with NUMA when cache data to Intel Optane DC persistent memory.")
+      // Just round the executorId to the total NUMA number.
+      numaId = executorId % XmlUtils.totalNumaNode(conf)
```
jerrychenhf: What if there are not persistent memory config file? Will totalNumaNode return 0 in this case or throw an exception?
If return 0, will cause an unrelated exception (divide by zero).  We may be able to handle this more clearly.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/XmlUtils.scala
```diff
@@ -0,0 +1,76 @@
...
+
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.internal.oap.OapConf
+import org.apache.spark.util.Utils
+
+/**
+ * A xml parser used for parse persistent memory config.
+ */
+@NotThreadSafe
+object XmlUtils {
```
jerrychenhf: The class/object name is XmlUtils. But it is actually a utils specific to persistent memory config handling.
We should rename, for example PersistentConfigUtils.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/XmlUtils.scala
```diff
@@ -0,0 +1,76 @@
...
+  private val RESERVED_SIZE_PROPERTY = "reservedSize"
+  private val numaToPMProperty = new mutable.HashMap[Int, PMProperty]()
+
+  def parseConfig(conf: SparkConf): mutable.HashMap[Int, PMProperty] = {
+    val configFile = conf.get(OapConf.OAP_FIBERCACHE_PERSISTENT_MEMORY_CONFIG_FILE.key,
+      DEFAULT_PERSISTENT_MEMORY_CONFIG_FILE)
+    // If already parsed, just return it
+    if (numaToPMProperty.size == 0) {
+      val xml = XML.load(
+        Utils.getSparkClassLoader.getResourceAsStream(configFile)
```
jerrychenhf: Is getResourceAsStream extract config file from Jar?
This persistent configuration should be changeable/editable by the user in the runtime. Can this satisfy the need?

src/main/java/org/apache/spark/unsafe/PMPlatform.java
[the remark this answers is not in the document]
ConeyLiu: only succeed can set true right? If failed the process will be killed.

conf/persistent-memory.xml.template
```diff
@@ -0,0 +1,35 @@
...
+    <initialSize>the initial size</initialSize>
+    <!--The reserved size of the total initial size which can't count into total available size.-->
+    <reservedSize>the reserved size</reservedSize>
+  </numanode>
+  <numanode id="1">
+    <initialPath>path to initial path</initialPath>
+    <initialSize>the initial size</initialSize>
+    <reservedSize>the reserved size</reservedSize>
+  </numanode>
+</persistentMemoryPool>
```
gczsjdy: :nit empty line

pom.xml
```diff
@@ -740,6 +740,38 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>pm</id>
```
gczsjdy: Full name: persistent-memory is better. Like `hive-thriftserver`.

pom.xml
```diff
@@ -740,6 +740,38 @@
...
+            <version>1.6.0</version>
+            <executions>
+              <execution>
+                <id>Compile native code and produce share library</id>
+                <phase>generate-resources</phase>
+                <goals>
+                  <goal>exec</goal>
+                </goals>
+                <configuration>
+                  <executable>${basedir}/src/main/native/compile.sh</executable>
```
gczsjdy: ${basedir}/bin/native-compile.sh?

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
```diff
@@ -0,0 +1,130 @@
...
+            break;
+          }
+          out.write(buf, 0, read);
+        }
+
+        out.flush();
+        out.close();
+        out = null;
+
+        System.load(tmpLib.getAbsolutePath());
```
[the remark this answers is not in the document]
ConeyLiu: may load failed

src/main/java/org/apache/spark/unsafe/PMPlatform.java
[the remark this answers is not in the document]
winningsix: My fault. Should be after L50.

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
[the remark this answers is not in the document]
winningsix: There're 2 levels try-catch here. I mean we move the try-catch out of the 1st level of block since the next IOException is only happening when loaded failed.

pom.xml
```diff
@@ -740,6 +740,38 @@
...
+        </resources>
+        <plugins>
+          <plugin>
+            <artifactId>exec-maven-plugin</artifactId>
+            <groupId>org.codehaus.mojo</groupId>
+            <version>1.6.0</version>
+            <executions>
+              <execution>
+                <id>Compile native code and produce share library</id>
+                <phase>generate-resources</phase>
```
gczsjdy: Why not phase `compile`?

src/main/java/org/apache/spark/unsafe/PMPlatform.java
```diff
@@ -0,0 +1,83 @@
...
+
+import com.google.common.base.Preconditions;
+import org.apache.spark.util.NativeLibraryLoader;
+
+import java.io.File;
+
+/**
+ * A platform used to allocate/free volatile memory from Intel Optane DC persistent memory.
+ */
+public class PMPlatform {
```
gczsjdy: I prefer we use full name. : )

src/main/java/org/apache/spark/unsafe/PMPlatform.java
```diff
@@ -0,0 +1,83 @@
+/*
```
gczsjdy: Do we take use of unsafe API in this class? 
Then why this package?

src/main/native/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+#
```
gczsjdy: What about moving to /conf? /native looks weird

src/main/native/compile.sh
```diff
@@ -0,0 +1,68 @@
...
+if [ -d build ]; then
+    rm -r build
+fi
+mkdir build
+cd build
+cmake -DCMAKE_INSTALL_PREFIX=${RESOURCES_DIR} ../
+make
+make install
+
+set +eu
```
gczsjdy: :nit new line

src/main/native/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+
+SET(SOURCE_FILES org_apache_spark_unsafe_PMPlatform.cpp)
+
+ADD_LIBRARY(pmplatform SHARED ${SOURCE_FILES})
+
+INSTALL(TARGETS pmplatform LIBRARY DESTINATION lib)
+
+TARGET_LINK_LIBRARIES(pmplatform memkind)
```
gczsjdy: :nit new line

src/main/native/org_apache_spark_unsafe_PMPlatform.cpp
```diff
@@ -0,0 +1,98 @@
+/*
```
gczsjdy: I'm not sure about the /native directory, any other opensource project uses like this?

pom.xml
ConeyLiu: this will generate the share library under the resources directory, so it needs to process before compile.

src/main/java/org/apache/spark/unsafe/PMPlatform.java
ConeyLiu: Just follow the `Platform`

src/main/java/org/apache/spark/unsafe/PMPlatform.java
ConeyLiu: `PersistentMemoryPlatform` ? Too noise?

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
```diff
@@ -0,0 +1,130 @@
...
+    String arch = System.getProperty("os.arch");
+    if (arch.contains("64")) {
+      arch = "64";
+    } else {
+      arch = "32";
+    }
+    return arch;
+  }
+
+  private static String resourceName() {
```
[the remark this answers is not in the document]
ConeyLiu: Why need it?

src/main/native/org_apache_spark_unsafe_PMPlatform.cpp
winningsix: That should be OK. https://github.com/apache/hadoop/blob/a55d6bba71c81c1c4e9d8cd11f55c78f10a548b0/hadoop-hdfs-project/hadoop-hdfs-native-client/src/main/native/libhdfspp/lib/common/auth_info.h

src/main/native/CMakeLists.txt
ConeyLiu: Why native weird? I seem much jni code follows this format. And the `CMakeLists.txt` is only used to compile the native code, not the overall project. Move to /conf will misleading people.

src/main/native/org_apache_spark_unsafe_PMPlatform.cpp
ConeyLiu: Just like we distinguish the java code and scala code, we need also do this with native code.

src/main/native/org_apache_spark_unsafe_PMPlatform.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include <memkind.h>
+#include <string>
+#include <cstdio>
+#include <cstdint>
+#include <cstdlib>
+#include <cassert>
+#include <stdexcept>
+#include "org_apache_spark_unsafe_PMPlatform.h"
+
+using memkind = struct memkind;
```
ConeyLiu: Currently, we rely on the version of `memkind`. Here maybe need a version check, I'll add a todo.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
ConeyLiu: I'll refactor here.

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
```diff
@@ -0,0 +1,129 @@
...
+          break;
+        }
+        out.write(buf, 0, read);
+      }
+
+      out.flush();
+      out.close();
+      out = null;
+
+      System.load(tmpLib.getAbsolutePath());
```
winningsix: Can we move L102-L103 to L119? Then we can make code much clean by avoiding unnecessary null value check.

src/main/scala/org/apache/spark/sql/internal/oap/OapConf.scala
```diff
@@ -122,10 +121,19 @@ object OapConf {
   val OAP_FIBERCACHE_MEMORY_MANAGER =
     SqlConfAdapter.buildConf("spark.sql.oap.fiberCache.memory.manager")
       .internal()
-      .doc("Sets the implement of memory manager, it only supports off heap currently.")
+      .doc("Sets the implement of memory manager, it only supports offheap(DRAM OFF_HEAP) and " +
+        "(pm) Intel Optane DC persistent memory currently.")
```
winningsix: pm -> PM

src/main/scala/org/apache/spark/sql/internal/oap/OapConf.scala
```diff
@@ -122,10 +121,19 @@ object OapConf {
   val OAP_FIBERCACHE_MEMORY_MANAGER =
     SqlConfAdapter.buildConf("spark.sql.oap.fiberCache.memory.manager")
       .internal()
-      .doc("Sets the implement of memory manager, it only supports off heap currently.")
+      .doc("Sets the implement of memory manager, it only supports offheap(DRAM OFF_HEAP) and " +
```
winningsix: space before (

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
ConeyLiu: No, we can't. Firstly, there still potential failed with `System.load(tmpLib.getAbsolutePath())`, eg: can't found `libmemkind.so`.  Secondly, we need delete the `tmpLib` even with exception.

src/main/scala/org/apache/spark/sql/internal/oap/OapConf.scala
ConeyLiu: where

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -187,3 +197,66 @@ private[filecache] class OffHeapMemoryManager(sparkEnv: SparkEnv)
...
+      logWarning(s"Executor ${executorId} is not bind with NUMA. It would be better to bind " +
+        s"executor with NUMA when cache data to Intel Optane DC persistent memory.")
+      // Just round the executorId to the total NUMA number.
+      // TODO: improve here
+      numaId = executorId % PersistentConfigUtils.totalNumaNode(conf)
+    }
+
+    val initialPath = map.get(numaId).get
+    val initialSize = conf.get(OapConf.OAP_FIBERCACHE_PERSISTENT_MEMORY_INITIAL_SIZE)
+    val reservedSize = conf.get(OapConf.OAP_FIBERCACHE_PERSISTENT_MEMORY_RESERVED_SIZE)
```
ConeyLiu: Hi @jerrychenhf, now we use the `persistent-memory.xml` to config the mapping between NUMA and the initial path. And use `OAP_FIBERCACHE_PERSISTENT_MEMORY_INITIAL_SIZE ` and `OAP_FIBERCACHE_PERSISTENT_MEMORY_RESERVED_SIZE ` to set the initial size and reserved size. 

Actually, the `persistent-memory.xml` should be located in `${SPARK_HOME}/conf`, so there are two choices to set the config file.:

- Config the `persistent-memory.xml` and compile the OAP source code with `-Ppersistent-memory`, the config file should be wrapped into jar.
- Config the `persistent-memory.xml` and put it into the `${SPARK_HOME}/conf`.

The `persistent-memory.xml` should not be modified at runtime just like other Spark configure files, the user should only use the other two parameters to set the memory size.

src/main/native/org_apache_spark_unsafe_PMPlatform.cpp
gczsjdy: I like it : )

gczsjdy:
Also add `-Ppersistent-memory` to .travis.yml?

src/main/java/org/apache/spark/unsafe/PMPlatform.java
```diff
@@ -0,0 +1,84 @@
...
+
+package org.apache.spark.unsafe;
+
+import com.google.common.base.Preconditions;
+import org.apache.spark.util.NativeLibraryLoader;
+
+import java.io.File;
+
+/**
+ * A platform used to allocate/free volatile memory from Persistent Memory e.g. Intel Optane DC
```
gczsjdy: Will this cause confusion? volatile memory from Persistent Memory? cc @winningsix @ConeyLiu

gczsjdy: It's easier for us to understand than others... Is `Persistent Memory` an official name? 
Anyway I suggest more illustration.

winningsix: Valid suggestion. Instead of adding too much comments about persistent memory, I tend to add a reference to the wiki here : https://en.wikipedia.org/wiki/Persistent_memory

ConeyLiu:
@gczsjdy Travis ci is OK to compile with the persistent-memory profile.

src/main/java/org/apache/spark/unsafe/PMPlatform.java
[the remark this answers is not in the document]
jerrychenhf: I see. Thanks!

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
gczsjdy: :nit order

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
gczsjdy: We can add != null check to `safeClose`, the annoying null value check will be saved.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -25,19 +26,26 @@ import org.apache.spark.SparkEnv
...
+ * @param length the requested size of the block
+ * @param occupiedSize the actual occupied size of the memory block
  */
-private[filecache] case class MemoryBlockWithOccupiedSize(
-    memoryBlock: MemoryBlock, occupiedSize: Long)
+private[filecache] case class MemoryBlockHolder(
+    baseObject: AnyRef,
+    baseOffset: Long,
+    length: Long,
+    occupiedSize: Long)
```
gczsjdy: I think it's better not to do this, plus, what's the benefit ?
We will lose encapsulation of `MemoryBlock`, and what if users want to call memoryBlock.fill?
I suggest we just -> OapMemoryBlock

src/main/scala/org/apache/spark/sql/execution/datasources/oap/utils/PersistentConfigUtils.scala
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.execution.datasources.OapException
+import org.apache.spark.sql.internal.oap.OapConf
+import org.apache.spark.util.Utils
+
+/**
+ * A xml parser used for parse persistent memory config.
+ */
+@NotThreadSafe
+object PersistentConfigUtils extends Logging {
```
gczsjdy: -> PersistentMemoryConfigUtils

src/main/scala/org/apache/spark/sql/internal/oap/OapConf.scala
```diff
@@ -122,10 +121,37 @@ object OapConf {
...
+      .internal()
+      .doc("Used to set the initial size of Intel Optane DC persistent memory.")
+      .stringConf
+      .createWithDefault("0b")
+
+  val OAP_FIBERCACHE_PERSISTENT_MEMORY_RESERVED_SIZE =
+    SqlConfAdapter.buildConf("spark.sql.oap.fiberCache.persistent.memory.reserved.size")
+      .internal()
+      .doc("Used to set the reserved size of Intel Optane DC persistent memory. Because the " +
+        "heap management of Intel Optane DC persistent memory are based on jemalloc, so we " +
```
gczsjdy: Still don't understand what this config means, users of OAP shouldn't be required to understand jemalloc to know how to use this...

src/main/scala/org/apache/spark/sql/internal/oap/OapConf.scala
```diff
@@ -122,10 +121,37 @@ object OapConf {
...
+      .internal()
+      .doc("A config file used to config the Intel Optane DC persistent memory initial path," +
+        " and mapping with NUMA node.")
+      .stringConf
+      .createWithDefault("persistent-memory.xml")
+
+  val OAP_FIBERCACHE_PERSISTENT_MEMORY_INITIAL_SIZE =
+    SqlConfAdapter.buildConf("spark.sql.oap.fiberCache.persistent.memory.initial.size")
+      .internal()
+      .doc("Used to set the initial size of Intel Optane DC persistent memory.")
```
gczsjdy: ...Could we explain what initial means here? Maybe just paraphrase?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -187,3 +197,68 @@ private[filecache] class OffHeapMemoryManager(sparkEnv: SparkEnv)
     memoryManager.releaseStorageMemory(oapMemory, MemoryMode.OFF_HEAP)
   }
 }
+
+/**
+ * A memory management which support allocate/free volatile memory from Intel Optane DC
```
gczsjdy: ->A memory manager which supports to ...

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -187,3 +197,68 @@ private[filecache] class OffHeapMemoryManager(sparkEnv: SparkEnv)
...
+ * persistent memory.
+ */
+private[filecache] class PersistentMemoryManager(sparkEnv: SparkEnv)
+  extends MemoryManager with Logging {
+
+  private val conf = sparkEnv.conf
```
gczsjdy: :nit put it inside init

src/main/native/org_apache_spark_unsafe_PMPlatform.cpp
```diff
@@ -0,0 +1,97 @@
...
+  check(env);
+  void *p = addr_from_java(address);
+  return memkind_malloc_usable_size(pmemkind, p);
+}
+
+JNIEXPORT void JNICALL Java_org_apache_spark_unsafe_PMPlatform_freeMemory
+  (JNIEnv *env, jclass clazz, jlong address) {
+  check(env);
+  memkind_free(pmemkind, addr_from_java(address));
+}
```
gczsjdy: :nit new line

.travis.yml
```diff
@@ -1,6 +1,17 @@
 sudo: required
 dist: trusty
 language: java
+before_install:
+- sudo apt-get install cmake
+- sudo apt-get install libpthread-stubs0-dev
+- sudo apt-get install libnuma-dev
+before_script:
```
gczsjdy: why not install?

.travis.yml
```diff
@@ -1,6 +1,17 @@
 sudo: required
 dist: trusty
 language: java
+before_install:
```
gczsjdy: Not about this patch, I see travis guides to put such apt-get in before_install, but why not install...

src/main/java/org/apache/spark/unsafe/PMPlatform.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.spark.util.NativeLibraryLoader;
+
+import java.io.File;
+
+/**
+ * A platform used to allocate/free volatile memory from
+ * <a href="https://en.wikipedia.org/wiki/Persistent_memory>Persistent Memory</a></a>
+ * e.g. Intel Optane DC persistent memory.
+ */
+public class PMPlatform {
```
gczsjdy: Good for `Platform`! But I suggest we don't save letters here, because there aren't any mapping of PM -> Persistent Memory in many developers mind. : )

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
gczsjdy: How about making this a general NativeLibLoader? We add a function named `loadPersistentMemoryLib` and wrap some of current code, also remove the comments on head of class containing `persistent memory`?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -168,17 +177,18 @@ private[filecache] class OffHeapMemoryManager(sparkEnv: SparkEnv)
 
   override def cacheGuardianMemory: Long = _cacheGuardianMemory
 
-  override private[filecache] def allocate(size: Long): MemoryBlockWithOccupiedSize = {
-    val block = MemoryAllocator.UNSAFE.allocate(size)
+  override private[filecache] def allocate(size: Long): MemoryBlockHolder = {
+    val address = Platform.allocateMemory(size)
```
gczsjdy: If we still use the previous `MemoryBlock` instead of replacing it with baseObj, offset and length, it's easier here to extract some common code.

Different concrete MemoryManagers just have to implement a `protected def allocateMemoryBlock(size: Long)`(new function added), and don't need to bother writing L182-184 and L251-254.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -187,3 +197,68 @@ private[filecache] class OffHeapMemoryManager(sparkEnv: SparkEnv)
...
+  extends MemoryManager with Logging {
+
+  private val conf = sparkEnv.conf
+  private val (_cacheMemory, _cacheGuardianMemory) = init()
+
+  private val _memoryUsed = new AtomicLong(0)
+
+  private def init(): (Long, Long) = {
+    // The NUMA id should be set when the executor process start up. However, Spark don't
+    // support NUMA binding currently.
```
gczsjdy: So actually it needs certain patch patched on Spark?

src/main/java/org/apache/spark/unsafe/PMPlatform.java
```diff
@@ -0,0 +1,85 @@
...
+
+  private static native void initializeNative(String path, long size);
+
+  /**
+   * Allocate volatile memory from persistent memory.
+   * @param size the requested size
+   * @return the address which same as Platform.allocateMemory, it can be operated by
+   * Platform which same as OFF_HEAP memory.
+   */
+  public static native long allocateMemory(long size);
```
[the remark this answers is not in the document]
ConeyLiu: np. All this method are used internal and shouldn't be opened to users.

.travis.yml
ConeyLiu: not sure the reason

.travis.yml
ConeyLiu: just follow the doc of travis

src/main/java/org/apache/spark/util/NativeLibraryLoader.java
ConeyLiu: np

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
ConeyLiu: I don't think we need to the dependency of `MemoryBlock` from Spark. You nedd to see this https://github.com/apache/spark/pull/19222. We need implement different `PMMemoryBlock` and `PMMemoryAllocator` for different version of Spark if we depend on `MemoryBlock`. And the code is very straightforward and not incomprehensible.

.travis.yml
```diff
@@ -1,7 +1,18 @@
...
+- git clone https://github.com/memkind/memkind.git
+- cd memkind && ./build.sh
+- make
+- sudo make install
+- cd ${TRAVIS_BUILD_DIR}
 script:
-- mvn clean -q -Pspark-2.1 test
-- mvn clean -q -Pspark-2.2 test
-- mvn clean -q -Pspark-2.3 test
+- mvn clean -Pspark-2.1 -Ppersistent-memory test
```
gczsjdy: -q 
: )

gczsjdy:
Thanks @ConeyLiu @winningsix @jerrychenhf LGTM

PR #866, opened 2018-09-05
Refactor OapConf to several categories

gczsjdy:

a certain entry of `OapConf` will be in one of the following categories: 
- parquet
- io
- index
- cache
- strategy
- rpc
- metrics
- oapfileformat

## How was this patch tested?

N/A


gczsjdy:
cc @linhong-intel @carsonwang @LuciferYang

gczsjdy:
Yes I will give a list here.
@deprecated is mainly for API? I think for conf we don't need to do that. See Spark SQLConf for reference.

PR #1035, opened 2019-04-11
OAP should create .oap.meta_bk file in data file dir

lidinghao:

At present ,  when create index, OAP will create all .oap.meta_bk files of different partitions in a fixed path (user.dir), which will cause indexing job fail if we create index on multiple partitions concurrently. The proper way is creating .oap.meta_bk in the same path as partition

## How was this patch tested?

add unit tests


lidinghao:
@LuciferYang  Could you review this PR,  thanks.

lidinghao:
if we revert this patch, the UT will fails as one of following case:
image
image

PR #1059, opened 2019-06-28
[Do not merge this] A fix for evenly calculating persistent memory initial path.

shaowenzh:
When spark is running on yarn mode, the executor Id is randomly allocated to the executor running on the same machine, the old way assign persistent memory based on if the id is odd or even under default config. But this will come to the situation that if two executors on the same machine are both even. It will only utilize one persistent memory if we have two installed on the server. 

This fix will have the executor trying to retrieve the its count number on the same machine by rpc.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
@@ -252,11 +252,14 @@ private[filecache] class PersistentMemoryManager(sparkEnv: SparkEnv)
     val executorId = sparkEnv.executorId.toInt
     val map = PersistentMemoryConfigUtils.parseConfig(conf)
     if (numaId == -1) {
+      val idNum = conf.get(s"spark.sql.oap.${executorId}.numa.num").toInt
       logWarning(s"Executor ${executorId} is not bind with NUMA. It would be better to bind " +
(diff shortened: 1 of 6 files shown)


shaowenzh:
@LuciferYang do you see this issue when test the persistent memory?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -249,14 +251,23 @@ private[filecache] class PersistentMemoryManager(sparkEnv: SparkEnv)
...
+        OapPmRpcDriverEndpoint.DRIVER_PM_ENDPOINT_NAME, sparkEnv.conf, sparkEnv.rpcEnv)
+      val driverReply = oapPmRpcDriverEndpoint.askSync[ReplyExecutorIdNumPerHost](
+        AskExecutorIdNumPerHost(executorId, Utils.localHostName()))
+      val idNum = driverReply.num
+
       logWarning(s"Executor ${executorId} is not bind with NUMA. It would be better to bind " +
         s"executor with NUMA when cache data to Intel Optane DC persistent memory.")
       // Just round the executorId to the total NUMA number.
       // TODO: improve here
-      numaId = executorId % PersistentMemoryConfigUtils.totalNumaNode(conf)
```
shaowenzh: Here we calculate the numa id by mod. But the id is not evenly distributed

PR #1063, opened 2019-07-10
support create index and refresh index on multi partitions with different location

lidinghao:
When partitions in same table have different locations (user may call "ALTER TABLE xx PARTITION xx SET LOCATION XX" to modify partition location ), create and refresh index command will failed, we should fix this.

## How was this patch tested?
new UTs

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexUtils.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexUtils.scala
@@ -22,18 +22,36 @@ import java.io.OutputStream
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
 
+import org.apache.spark.internal.Logging
+import org.apache.spark.internal.io.FileCommitProtocol
+import org.apache.spark.sql.Dataset
 import org.apache.spark.sql.RuntimeConfig
+import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.execution.datasources.FileIndex
+import org.apache.spark.sql.catalyst.catalog.CatalogTable
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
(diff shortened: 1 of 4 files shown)


lidinghao:
@LuciferYang @zhixingheyi-tian Could you review this patch? Thanks

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/IndexUtils.scala
```diff
@@ -227,6 +245,37 @@ private[oap] object IndexUtils {
     }
   }
 
+  def getOutputPathBasedOnConf(
+      partitionDirs: Seq[PartitionDirectory], fileIndex: FileIndex, conf: RuntimeConfig): Path = {
+
+    def getTableBaseDir(path: Path, times: Int): Path = {
+      if (times > 0) getTableBaseDir(path.getParent, times - 1)
+      else path
```
zhixingheyi-tian: Is there a better way to get the parent directory?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/index/indexPlans.scala
```diff
@@ -235,6 +142,56 @@ case class CreateIndexCommand(
...
+    val bAndP = buildPartitionMeta(sparkSession, identifier,
+      partitions, readerClassName,
+      schema, time)
+    if (bAndP.isEmpty) {
+      return Nil
+    }
+
+    val retMap = baseDirSet.flatMap{ baseDir =>
+      val partitions = partitionsGroupByBaseDir(baseDir)
+      // TODO currently we ignore empty partitions, so each partition may have different indexes,
```
zhixingheyi-tian: @lidinghao 
Here, trigger different jobs for different locations.  
Can it be further optimized here?

lidinghao: Because different partition location means different job output dir, so for,  one spark job  can have only on output dir.
Maybe we can optimized it later.

PR #1066, opened 2019-07-22
Support Persistent Memory destroy method and fiberCacheManager.resetAll()

shaowenzh:
Due to memkind memory allocate issue. This patch will allow user to reinitialize persistent memory cache pool.

Added destoryKind() method in JNI to destory memkind persistent memory pool. Added a resetAll method in fiberCacheManager to reinitialize persistent memory. Added sql grammer "RESET FIBER CACHE" to call fiberCacheManager.resetAll to reset PM cache pool.

## How was this patch tested?
ET


shaowenzh:
@LuciferYang Please take a look at this PR, see if this reset method could work for your scenario.

PR #1079, opened 2019-08-16
Using Address 0 to indicate DCPMM allocation failure

shaowenzh:
According to Haifeng's suggestion, memory address won't be 0. So in this PR we change the code to return address 0 when there is failed memory allocation.

## How was this patch tested?
ET


shaowenzh:
@LuciferYang This should be able to avoid the exception thrown by the native code.

PR #1097, opened 2019-11-07
Update the notify threshold from 0 to the dcpmm wait threshold

JkSelf:
Currently the thread will wait when the `_pendingFiberCapacity` > `dcpmmWaitingThreshold`, and it will be notified when the `_pendingFiberCapacity` ==0. This is not a fair wait/notify mechanism. It will cause the blocked thread will be notified when all other threads done in some extreme situation. This PR update the notify threshold from 0 to the dcpmm wait threshold. 

## How was this patch tested?
9 I/O intensive queries in 3TB TPC-DS


JkSelf:
With this PR, the total wait time can degrade from 23.4h to 13.4h after run 3 rounds 9 I/O intensive queries in 3TB TPC-DS.

PR #1098, opened 2019-11-12
sync free the fiber cache when the reference count =0 and if failed, still keep the origin release code path

JkSelf:
The currently single thread released method can cause performance degradation when heavily eviction. This PR implement the sync free when the reference count =0. And if released failed, we still go the single thread released mode.

## How was this patch tested?
9 I/O intensive queries.


JkSelf:
Here is the benchmark result.

  |   | With PR#1098 | Without PR#1098
-- | -- | -- | --
3TB | Round 1 | 671.94 | 781.64
||Round 2 | 565.75 | 1133.64
||Round 3 | 478.02 | 1593.32

JkSelf:
@zhixingheyi-tian Please help review. Thanks

PR #1111, opened 2019-12-06
improve the robustness of sync free

JkSelf:
When sync the fiber and the refCount is not 0,  it may cause the other thread to read the removing fiber if we wait `Thread.sleep(100)`.  This PR will directly return when the refCount is to 0 and not wait.

## How was this patch tested?
Manual test in TPC-DS

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
@@ -96,6 +96,39 @@ case class FiberCache(fiberData: MemoryBlockHolder) extends Logging {
     _refCount.decrementAndGet()
   }
 
+  def tryDisposeWithoutWait(): Boolean = {
+    require(fiberId != null, "FiberId shouldn't be null for this FiberCache")
+    val startTime = System.currentTimeMillis()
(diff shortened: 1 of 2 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCache.scala
```diff
@@ -96,6 +96,39 @@ case class FiberCache(fiberData: MemoryBlockHolder) extends Logging {
...
 
+  def tryDisposeWithoutWait(): Boolean = {
+    require(fiberId != null, "FiberId shouldn't be null for this FiberCache")
+    val startTime = System.currentTimeMillis()
+    val writeLockOp = OapRuntime.get.map(_.fiberLockManager.getFiberLock(fiberId).writeLock())
+    writeLockOp match {
+      case None => return true // already stopped OapRuntime
+      case Some(writeLock) =>
+        // Give caller a chance to deal with the long wait case.
+        while (System.currentTimeMillis() - startTime <= DISPOSE_TIMEOUT) {
```
jerrychenhf: We don't need the while loop for handling the time and wait things. So also remove the val startTime = ... line

JkSelf: @jerrychenhf  Thanks fro your comments. Updated.

JkSelf:
@zhixingheyi-tian @jerrychenhf Do you have any further comments?

PR #1121, opened 2020-01-03
import parquet 1.10.1 related source code for further binary cache use

ivoson:

Import ParquetFileReader source code with version 1.10.1 for binary cache use, since we need to visit some `private` fields and methods in binary cache reader as a subclass. 

This patch is one of the sub tasks for binary cache as discussed in  [#1047 ](https://github.com/Intel-bigdata/OAP/pull/1047).

## How was this patch tested?

existing ut.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -646,6 +646,7 @@
               <sources>
                 <source>src/main/spark2.4.4/java</source>
                 <source>src/main/spark2.4.4/scala</source>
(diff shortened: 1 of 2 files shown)


ivoson:
> @ivoson LGTM & we should add a new issue to refactor ParquetXXXReader, then we should decouple this file

sure, I will create a new issue to follow this.

PR #1156, opened 2020-03-12
Add RemoteShuffle codebase to OAP

gczsjdy:

Same as the title.

## How was this patch tested?

Tests under remote-shuffle/src/test.


gczsjdy:
cc @carsonwang

carsonwang:
Thanks! Merging this.

zhixingheyi-tian:
Hi  @gczsjdy @carsonwang 

Because .travis.yml in the subdirectory does not take effect. Would you please add test scripts to the .travis.yml in the root directory as needed?

Thanks.

gczsjdy:
@zhixingheyi-tian Thanks for the reminding.
Yes we need to do this, but I saw hardcoding `spark2.4.4` in root travis.yml. But current remote shuffle is compatible with Spark 2.4.0(there are API changes between 2.4.0 & 2.4.3). So the CI may not pass.
Also do we need to integrate remote-shuffle as an OAP module? (Indicate the hierarchy in pom.xml?)

gczsjdy:
@carsonwang I saw we merged this by squashing the commits to one... I am afraid in this way people cannot see the evolving history of RemoteShuffle, just a chunk commit...

zhixingheyi-tian:
Hi @gczsjdy ,

I have ever tried the way of mvn-like structure ,but Travis-CI doesn't support this currently. Refer to https://github.com/travis-ci/travis-ci/issues/3540

Also, we can use Matrix in Travis-CI.
A build matrix is made up by several multiple jobs that run in parallel.

Each module uses independent jobs so that they do not conflict with each other.

I will update the  travis structure.

Thanks.

gczsjdy:
@zhixingheyi-tian I didn't mean to have multiple .travis.yml, but: including the submodules in the upper-level pom.xml declaration.

Matrix is great. My latest PR #1167 takes use of it.

PR #1159, opened 2020-03-16
index/data cache seperation with different cache backend

yeyuqiang:
Support different cache backend for memory manager. Previous, guava cache is the only one strategy, but now VMemcache is included. So we need this feature.

## How was this patch tested?
A modified unit test IndexDataCacheSeparationSuite
tpcds query testing, add index for table store_sales
conf
oap_tab

TPCDS test
The first two part is for no sepration and mix .
 the thrid part is for mix. We should set spark.sql.oap.mix.index.memory.manager to mix and spark.oap.cache.strategy to mix
image

Add another test for share offheap/pm memorymanager in separate cache.
We should  set spark.oap.cache.strategy=mix and spark.sql.oap.index.data.cache.separation.enable to true
Capture

image

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
@@ -29,7 +29,8 @@ private[filecache] class CacheMemoryAllocator(sparkEnv: SparkEnv)
   extends Logging {
   private val _separateMemory = checkSeparateMemory()
   private val (memoryManager, indexMemoryManager) = init()
-  private val (_dataCacheMemory, _indexCacheMemory, _cacheGuardianMemory) = calculateSizes()
+  private val (_dataCacheMemory, _indexCacheMemory,
+              _dataCacheGuardianMemory, _indexCacheGuardianMemory) = calculateSizes()
 
   private def checkSeparateMemory(): Boolean = {
     val separateCache = sparkEnv.conf.getBoolean(
@@ -48,11 +49,11 @@ private[filecache] class CacheMemoryAllocator(sparkEnv: SparkEnv)
       case _ => false
     }
   }
-  private def calculateSizes(): (Long, Long, Long) = {
+  private def calculateSizes(): (Long, Long, Long, Long) = {
     // TO DO: make the 0.9 : 0.1 ration configurable
     if (_separateMemory) {
       ((memoryManager.memorySize * 0.9).toLong, (indexMemoryManager.memorySize * 0.9).toLong,
-        (memoryManager.memorySize * 0.1).toLong + (indexMemoryManager.memorySize * 0.1).toLong)
+        (memoryManager.memorySize * 0.1).toLong, (indexMemoryManager.memorySize * 0.1).toLong)
     } else {
       val cacheRatio = sparkEnv.conf.getDouble(
         OapConf.OAP_DATAFIBER_USE_FIBERCACHE_RATIO.key,
@@ -63,7 +64,7 @@ private[filecache] class CacheMemoryAllocator(sparkEnv: SparkEnv)
       val memorySize = memoryManager.memorySize
       ((memorySize * 0.9 * cacheRatio).toLong,
         (memorySize * 0.9 * (1 - cacheRatio)).toLong,
-        (memorySize * 0.1).toLong)
+        (memorySize * 0.1).toLong, 0)
     }
   }
 
@@ -121,7 +122,8 @@ private[filecache] class CacheMemoryAllocator(sparkEnv: SparkEnv)
 
   def dataCacheMemory: Long = _dataCacheMemory
   def indexCacheMemory: Long = _indexCacheMemory
-  def cacheGuardianMemory: Long = _cacheGuardianMemory
+  def dataCacheGuardianMemory: Long = _dataCacheGuardianMemory
+  def indexCacheGuardianMemory: Long = _indexCacheGuardianMemory
   def separateMemory: Boolean = _separateMemory
 }
 
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
@@ -29,8 +29,8 @@ private[filecache] class CacheMemoryAllocator(sparkEnv: SparkEnv)
   extends Logging {
   private val _separateMemory = checkSeparateMemory()
   private val (memoryManager, indexMemoryManager) = init()
-  private val (_dataCacheMemory, _indexCacheMemory,
-              _dataCacheGuardianMemory, _indexCacheGuardianMemory) = calculateSizes()
+  private val (_dataCacheMemorySize, _indexCacheMemorySize,
+              _dataCacheGuardianMemorySize, _indexCacheGuardianMemorySize) = calculateSizes()
 
   private def checkSeparateMemory(): Boolean = {
     val memoryManagerOpt =
@@ -119,10 +119,10 @@ private[filecache] class CacheMemoryAllocator(sparkEnv: SparkEnv)
     }
   }
 
-  def dataCacheMemory: Long = _dataCacheMemory
-  def indexCacheMemory: Long = _indexCacheMemory
-  def dataCacheGuardianMemory: Long = _dataCacheGuardianMemory
-  def indexCacheGuardianMemory: Long = _indexCacheGuardianMemory
+  def dataCacheMemorySize: Long = _dataCacheMemorySize
+  def indexCacheMemorySize: Long = _indexCacheMemorySize
+  def dataCacheGuardianMemorySize: Long = _dataCacheGuardianMemorySize
+  def indexCacheGuardianMemorySize: Long = _indexCacheGuardianMemorySize
   def separateMemory: Boolean = _separateMemory
 }
 
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
@@ -33,18 +33,17 @@ private[filecache] class CacheMemoryAllocator(sparkEnv: SparkEnv)
               _dataCacheGuardianMemory, _indexCacheGuardianMemory) = calculateSizes()
 
   private def checkSeparateMemory(): Boolean = {
-    val separateCache = sparkEnv.conf.getBoolean(
-      OapConf.OAP_INDEX_DATA_SEPARATION_ENABLE.key,
-      OapConf.OAP_INDEX_DATA_SEPARATION_ENABLE.defaultValue.get
-    )
     val memoryManagerOpt =
       sparkEnv.conf.get(OapConf.OAP_FIBERCACHE_MEMORY_MANAGER.key, "offheap").toLowerCase
+    val cacheName =
+      sparkEnv.conf.get(OapConf.OAP_FIBERCACHE_STRATEGY.key, "guava").toLowerCase
+
     memoryManagerOpt match {
-      case "mix" => if (separateCache) {
+      case ("mix") => if (cacheName.equals("mix")) {
         true
       } else {
         throw new UnsupportedOperationException("In order to enable mix memory manager," +
-          "you need to set to spark.sql.oap.index.data.cache.separation.enable to true")
+          "you need also to set to spark.oap.cache.strategy to mix")
       }
       case _ => false
     }
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
@@ -27,7 +27,7 @@ import org.apache.spark.sql.internal.oap.OapConf
  */
 private[filecache] class CacheMemoryAllocator(sparkEnv: SparkEnv)
   extends Logging {
-  private val separateMemory = checkSeparateMemory()
+  private val _separateMemory = checkSeparateMemory()
   private val (memoryManager, indexMemoryManager) = init()
   private val (_dataCacheMemory, _indexCacheMemory, _cacheGuardianMemory) = calculateSizes()
 
@@ -50,7 +50,7 @@ private[filecache] class CacheMemoryAllocator(sparkEnv: SparkEnv)
   }
   private def calculateSizes(): (Long, Long, Long) = {
     // TO DO: make the 0.9 : 0.1 ration configurable
-    if (separateMemory) {
+    if (_separateMemory) {
       ((memoryManager.memorySize * 0.9).toLong, (indexMemoryManager.memorySize * 0.9).toLong,
         (memoryManager.memorySize * 0.1).toLong + (indexMemoryManager.memorySize * 0.1).toLong)
     } else {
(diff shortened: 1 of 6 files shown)


yeyuqiang:
@zhixingheyi-tian @winningsix Please help to review.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
```diff
@@ -36,21 +36,21 @@ private[filecache] class CacheMemoryAllocator(sparkEnv: SparkEnv)
       OapConf.OAP_INDEX_DATA_SEPARATION_ENABLE.key,
       OapConf.OAP_INDEX_DATA_SEPARATION_ENABLE.defaultValue.get
     )
-    val memoryManagerOpt =
+    val memoryManagerOpt =
```
winningsix: Why does it show some diffs here?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -510,23 +538,45 @@ class VMemCache extends OapCache with Logging {
...
+
+    if (fiberType == FiberType.INDEX) {
+      CacheStats(
+        0, // dataFiberCount
+        0, // dataFiberSize JNIGet
+        cacheTotalCount, // indexFiberCount
+        cacheTotalSize, // indexFiberSize
+        cacheGuardian.pendingFiberCount, // pendingFiberCount
+        cacheGuardian.pendingFiberSize, // pendingFiberSize
+        0L,
```
winningsix: Avoiding magic numbers here.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -587,150 +634,87 @@ class GuavaOapCache(
...
-        dataStats.missCount(),
-        dataStats.loadCount(),
-        dataStats.totalLoadTime(),
-        dataStats.evictionCount(),
-        indexStats.hitCount(),
-        indexStats.missCount(),
-        indexStats.loadCount(),
-        indexStats.totalLoadTime(),
-        indexStats.evictionCount()
+        0L,
```
winningsix: Please avoid magic numbers.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -65,47 +60,61 @@ private[sql] class FiberCacheManager(
...
-      new NonEvictPMCache(dataCacheMemory, cacheGuardianMemory)
-    } else if (cacheName.equals(VMEM_CACHE)) {
-      new VMemCache()
+  private val separateMemory = cacheAllocator.separateMemory
+  private val (cacheBackend, indexCacheBackend) = init()
+
+  private def init(): (OapCache, OapCache) = {
+    if (separateMemory) {
+      val dataCacheBackend = OapCache(sparkEnv, OapConf.OAP_MIX_DATA_CACHE_BACKEND,
+        dataCacheMemory, cacheGuardianMemory, FiberType.DATA);
```
winningsix: ```cacheGuardianMemory``` should be different for index and data cache backend.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
```diff
@@ -226,26 +239,40 @@ private[sql] class FiberCacheManager(
...
     }
   }
 
-  // Only used by test suite
-  private[filecache] def enableGuavaCacheSeparation(): Unit = {
-    if (cacheBackend.isInstanceOf[GuavaOapCache]) {
-      cacheBackend.asInstanceOf[GuavaOapCache].enableCacheSeparation()
+  // Used by test suite
+  private[oap] def clearAllFibers(): Unit = {
+    if (separateMemory) {
```
winningsix: Too many duplicated branch handles for ```separateMemory```.  In parallel to other cache backend, I am suggesting to introduce a new cache backend saying ```separatedCacheBackend```. Then you can make ```FiberCacheManager``` much cleaner.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -202,6 +204,31 @@ private[filecache] class CacheGuardian(maxMemory: Long) extends Thread with Logg
...
 }
 
+private[filecache] object OapCache {
+
+  def apply(sparkEnv: SparkEnv, cacheMemory: Long,
+            cacheGuardianMemory: Long, fiberType: FiberType): OapCache = {
+    apply(sparkEnv, OapConf.OAP_FIBERCACHE_STRATEGY, cacheMemory, cacheGuardianMemory, fiberType)
+  }
+
+  def apply(sparkEnv: SparkEnv, configEntry: ConfigEntry[String],
```
winningsix: could we use a ```parameter group``` instead of multiple parameters to simplify the code?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
yeyuqiang: Not clearly about it. I should not change any here. Looks like there's also no redundant space here

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
yeyuqiang: here have to pass a zero. I give some comments to make it meaningful

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
yeyuqiang: Give another cacheGuardianMemory for index cache

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/FiberCacheManager.scala
yeyuqiang: Refact the code. Add a MixCache  in OapCache which will instantiate two cache backend instance

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -510,23 +538,33 @@ class VMemCache extends OapCache with Logging {
...
-      cacheEvictCount, // dataEvictionCount
-      0, // indexFiberHitCount
-      0, // indexFiberMissCount
-      0, // indexFiberLoadCount
-      0, // indexFiberLoadTime
-      0) // indexEvictionCount JNIGet
+
+    if (fiberType == FiberType.INDEX) {
+      CacheStats(
+        dataFiberCount.get(), dataFiberSize.get(),
```
winningsix: Could we pass in an ```Option``` type avoiding long parameter lists? Considering most of them are primitive type ```Int```.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -766,15 +717,125 @@ class GuavaOapCache(
...
+class MixCache(dataCacheMemory: Long,
+               indexCacheMemory: Long,
+               dataCacheGuardianMemory: Long,
+               indexCacheGuardianMemory: Long,
+               sparkEnv: SparkEnv)
+  extends OapCache with Logging {
+
+  private val (dataCacheBackend, indexCacheBackend) = init()
+
+  private def isCompatibleWithMemoryManager() = {
```
winningsix: Great to see we have such utility method. Can we make it more fancy? Saying make Memory manager configuration as a next level configuration for cache manager. Example, ```memkind.manager``` and for ```vmemcache```, no need to configure memory manager.

winningsix: @yma11 @jikunshang your thoughts?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
zhixingheyi-tian: There may be a difference in line breaks

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
```diff
@@ -27,32 +27,33 @@ import org.apache.spark.sql.internal.oap.OapConf
...
-      case _ => false
+    memoryManagerOpt match {
+      case "mix" => if (separateCache) {
+        true
+      } else {
+        throw new UnsupportedOperationException("In order to enable mix memory manager," +
+          "you need to set to spark.sql.oap.index.data.cache.separation.enable to true")
+      }
+      case _ => false
     }
```
zhixingheyi-tian: The same as above.
Please confirm the cause for the diffs.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -587,111 +622,52 @@ class GuavaOapCache(
...
-          fiberCache
-        }
+      val fiberCache = cacheInstance.get(fiber)
+      // Avoid loading a fiber larger than MAX_WEIGHT / CONCURRENCY_LEVEL
+      assert(fiberCache.size() <= MAX_WEIGHT / CONCURRENCY_LEVEL,
+        s"Failed to cache fiber(${Utils.bytesToString(fiberCache.size())}) " +
+          s"with cache's MAX_WEIGHT" +
+          s"(${Utils.bytesToString(MAX_WEIGHT)}) / $CONCURRENCY_LEVEL")
+      fiberCache.occupy()
+      fiberCache
```
zhixingheyi-tian: Why remove so many lines of code?

src/main/scala/org/apache/spark/sql/internal/oap/OapConf.scala
```diff
@@ -196,6 +207,20 @@ object OapConf {
...
+      .doc("Sets the implement of index cache backend in mix mode.")
+      .stringConf
+      .createWithDefault("guava")
+
+  val OAP_MIX_DATA_CACHE_BACKEND =
+    SqlConfAdapter.buildConf("spark.sql.oap.mix.data.cache.backend")
+      .internal()
+      .doc("Sets the implement of data memory manager in mix mode.")
+      .stringConf
+      .createWithDefault("guava")
```
zhixingheyi-tian: Is OAP_MIX_DATA_CACHE_BACKEND Vmem or guava?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
yeyuqiang: Previous code will exist lot ^M which should cause by the some editor. Here,  looks like my editor remove ^M automatically

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
yeyuqiang: MixCache will do separation logic. So no need to separation in GuavaCache

src/main/scala/org/apache/spark/sql/internal/oap/OapConf.scala
yeyuqiang: The user can set it to vmem. But by default, guava will be also ok

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -766,15 +717,125 @@ class GuavaOapCache(
...
+  }
+
+  override def dataCacheCount: Long = {
+    cacheInstance.asMap().keySet().asScala
+      .count(fiber => fiber.isInstanceOf[DataFiberId] || fiber.isInstanceOf[TestDataFiberId])
+  }
+}
+
+class MixCache(dataCacheMemory: Long,
+               indexCacheMemory: Long,
```
zhixingheyi-tian: About MixCache, this is some kind of refactoring. Please ensure that it does not affect the original two separation functionality:  index and data are on the same medium; on the different medium.

yeyuqiang: Yes, I've already test for it. Will post the test data soon.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
winningsix: Perhaps you can try ```git config --global core.autocrlf false``` to ignore those changes.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/CacheMemoryAllocator.scala
```diff
@@ -114,14 +115,16 @@ private[filecache] class CacheMemoryAllocator(sparkEnv: SparkEnv)
...
-    if (separateMemory) {
+    if (_separateMemory) {
       indexMemoryManager.stop()
     }
   }
 
   def dataCacheMemory: Long = _dataCacheMemory
   def indexCacheMemory: Long = _indexCacheMemory
-  def cacheGuardianMemory: Long = _cacheGuardianMemory
+  def dataCacheGuardianMemory: Long = _dataCacheGuardianMemory
```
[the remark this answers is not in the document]
winningsix: This one is even better: ```dataCacheGuardianMemorySizeAsLong```

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -510,23 +538,33 @@ class VMemCache extends OapCache with Logging {
...
-      0, // indexFiberHitCount
-      0, // indexFiberMissCount
-      0, // indexFiberLoadCount
-      0, // indexFiberLoadTime
-      0) // indexEvictionCount JNIGet
+
+    if (fiberType == FiberType.INDEX) {
+      CacheStats(
+        0, 0,
+        cacheTotalCount, cacheTotalSize,
```
winningsix: The same as above, passing as ```option```

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
jikunshang: I think it's doable. OapCache add a memoryManager, and subclass could have a default impl. We need modify some class constructor. Saying OapCache needs a "SparkEnv" for initial memoryManager. This should be in another PR since code would be a lot mess.

winningsix: Agreed, it definitely should exist in another PR.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
yeyuqiang: Here will always 0 for index because this branch is for data cache.

winningsix: To be more clear, you can change ```CacheStats``` constructor into something like ```CacheStats(Option[ConfigEntry] dataCacheStats, Option[ConfigEntry] indexCacheStats)```.

winningsix:
@yeyuqiang any idea why performance is dropping at the 3rd iteration in the performance data above?

yeyuqiang:
> @yeyuqiang any idea why performance is dropping at the 3rd iteration in the performance data above?

Looks like there still be some variation. I run serveral time. The dropping in 3rd can not be produced every time

winningsix:
>Looks like there still be some variation. I run serveral time. The dropping in 3rd can not be produced every time

Fairly enough. I will merge it once comments are addressed.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
yeyuqiang: Let's fire another PR to do this refact. This change would impact lines of source code and units.

PR #1169, opened 2020-03-25
update user guide organization on cache strategies part

yma11:

Reorg DCPM cache part based on issues found in customer OAP POC.

## How was this patch tested?

N/A


winningsix:
LGTM. I will commit it if no further comments left.

PR #1172, opened 2020-03-27
[remote-shuffle] Introduce new performance evaluation tool and deprecate the old micro-benchmark

gczsjdy:

The former micro-benchmark tool has following disadvantages:
- mainly focuses on the comparison between upstream Spark's shuffle facility and that of the new one;
- is single-threaded;
- most configurations are hardcoded, like the read/write concurrency, data volume, etc;
- fails to directly measure the remote-shuffle facility with a throughput number (in the user-specific environment)

This PR introduces a new performance evaluation tool that starts multiple threads to saturate the I/O bottleneck, solving the above issues.

This PR squashed the commits in https://github.com/Intel-bigdata/RemoteShuffle/pull/48:

* Add shuffle writer benchmark

* Infinite mem & can also test vanilla local shuffle & row size configurable

* Shuffle dir, log level configurable

* Allow disable transfer to

* Fix

* Generate before testing

* Int as key, bytebuffer as payLoad

* Logging task memory manager acquiring bytes

* Correctly setup TaskMemoryManager

* Clean up after finishing

* Block size & reducers configurable

* Fix remote resolver, print detailed log to console & file

* Add shuffle read benchmark

* Print more messages after finishing

* Use the same pair of SparkConfs from beginning to end

* Allow only testing shuffle write

* 2 data sizes: raw and compressed

* Fix

* Data payload more random, remove TestMemoryManager logs

* Style

* Fix: use Long to prevent overflow

* Make each generated row different & some refactor

* Modify travis

* Load spark configurations from spark.conf

* Support other 2 shuffle writers

* Print more info: records, load per record

* Logs

* Pool size configurable, simulating running multi-rounds

* Printing and recording SparkConf for each test

* Style

* Only display duration for successful running & refactor

* Print debugging string for mydebug level

* Fix: shuffle reader gets right info from MapOutputTracker

* Fix support for other 2 writers

* Use same set of SparkConfs

* Close file stream

* Fix multi-thread Mock stubbing using ThreadLocal

* No concurrent stubbing

## How was this patch tested?

Manual tests.


gczsjdy:
There are some weird conflicts, are you resolving this? @ccobb6 @LuciferYang @winningsix

carsonwang:
@winningsix did you have forced push to the branch? I think all commits should be merged by PRs, no direct push.

winningsix:
@gczsjdy I rebase the 0.8 branch with 0.7. Let me check it.

winningsix:
@jikunshang how did you resolve the conflict? By rebasing and creating a new PR?

gczsjdy:
Resolved, thanks all! @carsonwang @zjf2012 Could you please help me review this? Any advice is welcome!

carsonwang:
@gczsjdy , what about adding a simple doc about how to run this perf tool and an example output?

carsonwang:
Merge this first.

PR #1174, opened 2020-03-27
fix cache guardian use too much memory

jikunshang:

TmpMemmroyManager will use too much memory and throw exception when allocate size is more than 'spark.sql.oap.cache.guardian.memory.size'.  TmpMemoryManager.allocate() will sleep for a while and will not throw exception directly.

## How was this patch tested?

I tested this patch with tpc-ds 9 query on Parquet 1TB, 3TB dataset (generated by spark-sql-perf 0.5.1). Test passed while previous version failed.

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
@@ -169,7 +169,9 @@ private[filecache] class TmpDramMemoryManager(sparkEnv: SparkEnv)
     sparkEnv.conf.get(OapConf.OAP_CACHE_GUARDIAN_MEMORY_SIZE)
   val cacheGuardianMemory = Utils.byteStringAsBytes(cacheGuardianMemorySizeStr)
   logInfo(s"cacheGuardian total use $cacheGuardianMemory bytes memory")
-  val cacheGuardianRetryTime = sparkEnv.conf.get(OapConf.OAP_CACHE_GUARDIAN_RETRY_TIME_IN_MS) / 10
+  val cacheGuardianRetrySleepTimeInMs = 10
+  val cacheGuardianRetryTime =
+    sparkEnv.conf.get(OapConf.OAP_CACHE_GUARDIAN_RETRY_TIME_IN_MS) / cacheGuardianRetrySleepTimeInMs
 
   private val _memoryUsed = new AtomicLong(0)
   override def memoryUsed: Long = _memoryUsed.get()
@@ -179,11 +181,12 @@ private[filecache] class TmpDramMemoryManager(sparkEnv: SparkEnv)
     var retryTime: Int = 0
     while(memoryUsed + size > cacheGuardianMemory) {
       retryTime += 1
-      Thread.sleep(10)
+      Thread.sleep(cacheGuardianRetrySleepTimeInMs)
       if (retryTime > cacheGuardianRetryTime) {
         throw new OapException("cache guardian use too much memory over " +
-          cacheGuardianRetryTime * 10 + "ms," +
-          " please consider increase cache guardian size")
+          cacheGuardianRetryTime * cacheGuardianRetrySleepTimeInMs + "ms, please consider " +
+          "increase memory size by 'spark.sql.oap.cache.guardian.memory.size' configuration or " +
+          "increase retry time by 'spark.sql.oap.cache.guardian.retry.time.in.ms' configuration")
       }
     }
     val startTime = System.currentTimeMillis()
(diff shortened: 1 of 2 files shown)


jikunshang:
@zhixingheyi-tian  @yma11  @yeyuqiang @winningsix  Please take a review. Thanks!

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -175,8 +175,14 @@ private[filecache] class TmpDramMemoryManager(sparkEnv: SparkEnv)
   override def memorySize: Long = cacheGuardianMemory
 
   override private[filecache] def allocate(size: Long): MemoryBlockHolder = {
-    if (memoryUsed + size > cacheGuardianMemory) {
-      throw new OapException("cache guardian use too much memory")
+    var retryTime: Int = 0
+    while(memoryUsed + size > cacheGuardianMemory) {
+      retryTime += 1
+      Thread.sleep(10)
+      if (retryTime > 100) {
```
winningsix: Can we make it configurable and making 100 as default value?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -175,8 +175,14 @@ private[filecache] class TmpDramMemoryManager(sparkEnv: SparkEnv)
...
+        throw new OapException("cache guardian use too much memory over 1000 ms," +
```
winningsix: Instead of throwing exception, we can reserve more memory for guardian memory?

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -169,14 +169,22 @@ private[filecache] class TmpDramMemoryManager(sparkEnv: SparkEnv)
...
   override def memoryUsed: Long = _memoryUsed.get()
   override def memorySize: Long = cacheGuardianMemory
 
   override private[filecache] def allocate(size: Long): MemoryBlockHolder = {
-    if (memoryUsed + size > cacheGuardianMemory) {
-      throw new OapException("cache guardian use too much memory")
+    var retryTime: Int = 0
+    while(memoryUsed + size > cacheGuardianMemory) {
+      retryTime += 1
+      Thread.sleep(10)
```
winningsix: We should define 10 as a constant and use it to calculate ```cacheGuardianRetryTime```. 
```
val CACHE_GUARDIAN_RETRY_SLEEP_TIME_IN_MS = 10
val cacheGuardianRetryTime = sparkEnv.conf.get(OapConf.OAP_CACHE_GUARDIAN_RETRY_TIME_IN_MS) / WAIT_TIME_IN_MS 
```
```
while(memoryUsed + size > cacheGuardianMemory) {
      retryTime += 1
      Thread.sleep(CACHE_GUARDIAN_RETRY_SLEEP_TIME_IN_MS)
```

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -169,14 +169,22 @@ private[filecache] class TmpDramMemoryManager(sparkEnv: SparkEnv)
...
+      if (retryTime > cacheGuardianRetryTime) {
+        throw new OapException("cache guardian use too much memory over " +
+          cacheGuardianRetryTime * 10 + "ms," +
```
winningsix: Avoiding magic num. See last comment.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -169,14 +169,22 @@ private[filecache] class TmpDramMemoryManager(sparkEnv: SparkEnv)
...
+          " please consider increase cache guardian size")
```
winningsix: Please explicitly mention how to do it by saying ```increase memory size by XXX configuration```.

winningsix:
LGTM. Will merge it if no other comments left.

src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -169,14 +169,25 @@ private[filecache] class TmpDramMemoryManager(sparkEnv: SparkEnv)
...
   override def memorySize: Long = cacheGuardianMemory
 
   override private[filecache] def allocate(size: Long): MemoryBlockHolder = {
-    if (memoryUsed + size > cacheGuardianMemory) {
-      throw new OapException("cache guardian use too much memory")
+    var retryTime: Int = 0
+    while(memoryUsed + size > cacheGuardianMemory) {
+      retryTime += 1
+      Thread.sleep(cacheGuardianRetrySleepTimeInMs)
+      if (retryTime > cacheGuardianRetryTime) {
```
zhixingheyi-tian: Is there a better way to replace sleep, such as notify / wait?

zhixingheyi-tian:
@jikunshang 

Have you verified the performance impact of add this "sleep"?

jikunshang:
> @jikunshang
> 
> Have you verified the performance impact of adding this "sleep"?

@zhixingheyi-tian For memory enough case, it has no impact. For memory not enough case, some queries will fail with previous implementation, and it won't fail with this patch.

winningsix:
Another option is to on hold the memory request when we reach a threshold. You can just maintain a atomic value for memory usage in allocate and free sides. Thoughts?

PR #1175, opened 2020-03-30
update user guide organization on cache strategies part

yma11:

Reorg DCPM cache part based on issues found in customer OAP POC.

## How was this patch tested?

N/A

### DIFF
--- a/docs/OAP-User-Guide.md
+++ b/docs/OAP-User-Guide.md
@@ -163,17 +163,58 @@ Step 5. To verify that the cache functionality is in effect, you can open Spark 
 #### Prerequisites
 Before configuring in OAP to use DCPMM cache, you need to make sure the following:
 
-- DCPMM hardwares are installed, formatted and mounted correctly on every cluster worker node. You will get a mounted directory to use if you have done this. Usually, the DCPMM on each socket will be mounted as a directory. For example, on a two sockets system, we may get two mounted directories named `/mnt/pmem0` and `/mnt/pmem1`.
+- DCPMM hardwares are installed, formatted and mounted correctly on every cluster worker node with AppDirect mode(refer [guide](https://software.intel.com/en-us/articles/quick-start-guide-configure-intel-optane-dc-persistent-memory-on-linux)). You will get a mounted directory to use if you have done this. Usually, the DCPMM on each socket will be mounted as a directory. For example, on a two sockets system, we may get two mounted directories named `/mnt/pmem0` and `/mnt/pmem1`.
+```
+	// use impctl command to show topology and dimm info of DCPM
+	impctl show -topology
+	impctl show -dimm
+	// provision dcpm in app direct mode
+	ipmctl create -goal PersistentMemoryType=AppDirect
+	// reboot system to make configuration take affect
+	reboot
+	// check capacity provisioned for app direct mode(AppDirectCapacity)
+	impctl show -memoryresources
+	// show the DCPM region information
+	impctl show -region
+	// create namespace based on the region, multi namespaces can be created on a single region
+	ndctl create-namespace -m fsdax -r region0
+	ndctl create-namespace -m fsdax -r region1
+	// show the created namespaces
+	fdisk -l
+	// create and mount file system
+	mount -o dax /dev/pmem0 /mnt/pmem0
+	mount -o dax /dev/pmem1 /mnt/pmem1
+```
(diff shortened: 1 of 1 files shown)


docs/OAP-User-Guide.md
```diff
@@ -161,16 +163,36 @@ Step 5. To verify that the cache functionality is in effect, you can open Spark
...
+```
+     git clone https://github.com/memkind/memkind
+     cd memkind
+     ./autogen.sh
+     ./configure
+     make
+     make install
+```
+- [Vmemcache](https://github.com/pmem/vmemcache) library has been installed on every cluster worker node if vmemcache strategy is chosen for DCPM cache. You can follow the build/install steps from vmemcache website and make sure libvmemcache.so exist in '/lib64' directory in each worker node. To build vmemcache lib from source, you can (for RPM-based linux as example):
+```
```
winningsix: Hmm, we should consider making it as a dedicated script. Thoughts?

yma11: Yes. We should put as many operations as we can in a setup script. What I am thinking is, numactl/memkind/vmemcache installation, autogenerating pmem configuration file after confirm numa nodes in system as well as pmem mount paths, spark-default conf auto update based on cache strategy choose/data format choose, classpath setting of OAP jar, etc.

docs/OAP-User-Guide.md
```diff
@@ -117,6 +117,8 @@ Step 1. Make the following configuration changes in Spark configuration file `$S
 ```
 spark.memory.offHeap.enabled                true
 spark.memory.offHeap.size                   80g      # half of total memory size
+spark.sql.oap.fiberCache.memory.manager     offheap  # for dram cache, use 'offheap' as the memory manager
+conf spark.oap.cache.strategy               guava    # dram cache uses guava as the cache strategy
 spark.sql.oap.parquet.data.cache.enable     true     # for parquet fileformat
```
zhixingheyi-tian: These two configuration items have default values

docs/OAP-User-Guide.md
```diff
@@ -194,35 +216,48 @@ spark.yarn.numa.enabled                                    true            # ena
...
-##### Non-evictable Cache strategy
+Guava cache is based on memkind library, built on top of jemalloc and provides memory characteristics. To use it in your workload, follow [prerequisites](#prerequisites-1) to set up DCPMM hardware and memkind library correctly. Then follow bellow configurations.
 
-Another cache strategy named Non-evictable is also supported in OAP based on memkind library for DCPMM.
+For Parquet data format, provides following conf options:
+```
+--conf spark.sql.oap.parquet.data.cache.enable=true \
+--conf spark.sql.oap.fiberCache.memory.manager=pm \
+--conf spark.oap.cache.strategy=guava
+--conf spark.sql.oap.fiberCache.persistent.memory.initial.size=*g \
```
zhixingheyi-tian: The configuration item does not have "=", in spark-default.conf

docs/OAP-User-Guide.md
winningsix: It's great to have this. I would suggest to provide a couple of tool script as beginning. Say, one script for numa setting check, one script for memkind installation, etc. Moving forwards, we can explore the possibility of having a fully atomically script.

docs/OAP-User-Guide.md
winningsix: Inspired by the comments, could we check the correctness of user settings? For incorrect settings, we can directly throw exception. This can avoid unnecessary troubleshooting for us. @yma11

yma11: Instead of check existing OAP configurations of user, seems it should be more reliable to let them use the script in a clean environment without any OAP conf settings. Detecting incorrect setting is no so easy.

docs/OAP-User-Guide.md
yma11: update this part in new commit.

docs/OAP-User-Guide.md
yma11: What about keep doc like this for the time being? the installation part is very simple and it should be okay for user to copy these commands. I would like to start with a better automatic script that we can provide POC customer first. Then we can submit here after some refinement.

winningsix: Yes, I am talking about long term plan. Feel free to address it in separated PR.

winningsix:
LGTM + 1

PR #1185, opened 2020-04-07
[CPP] Fixed a avg with groupby issue, now Q17 is correct

xuechendi:


zhixingheyi-tian:
@xuechendi 

Would you  add the prefix for every PR‘s  title？ For example : [oap-native-sql]

xuechendi:
> @xuechendi
> 
> Would you add the prefix for every PR‘s title？ For example : [oap-native-sql]

@zhixingheyi-tian , sure, will follow this naming pattern in future. Thanks

PR #1193, opened 2020-04-10
[nativesql] load native lib from jar

zhouyuan:
mvn package -Dnative.cpp.build.dir=/path/to/native/lib/

package the native lib into target jar so we could load native lib from there

## How was this patch tested?


zhouyuan:
in #1205

PR #1197, opened 2020-04-10
update unit test for memory manager config options

yma11:

add more unit test for memory manager config options

## How was this patch tested?

N/A


src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -342,7 +350,14 @@ private[filecache] class HybridMemoryManager(sparkEnv: SparkEnv)
     // The NUMA id should be set when the executor process start up. However, Spark don't
     // support NUMA binding currently.
     var numaId = conf.getInt("spark.executor.numa.id", -1)
-    val executorId = sparkEnv.executorId.toInt
+    val executorIdStr = sparkEnv.executorId
+    scala.util.Try(executorIdStr.toInt) match {
+      case Success(_) => logDebug("valid executor id for numa binding.") ;
+      case _ =>
+        logInfo("invalid executor id for numa binding.")
```
winningsix, line 357: logWarning

src/test/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManagerConfigSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  test("guava cache with pm memory manager") {
+    val sparkEnv = SparkEnv.get
+    sparkEnv.conf.set("spark.oap.cache.strategy", "guava")
+    sparkEnv.conf.set("spark.sql.oap.fiberCache.memory.manager", "pm")
+    val memoryManager = MemoryManager(sparkEnv)
+    assert(memoryManager.isInstanceOf[PersistentMemoryManager])
+  }
+
+  test("vmem with tmp memory manager") {
```
winningsix, line 50: Please add one test case. ```Vemcache``` and set to ```PM```. We should still return ```TmpMemMgr```

src/test/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManagerConfigSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+    val sparkEnv = SparkEnv.get
+    sparkEnv.conf.set("spark.oap.cache.strategy", "vmem")
+    val memoryManager = MemoryManager(sparkEnv)
+    assert(memoryManager.isInstanceOf[TmpDramMemoryManager])
+  }
+
+  test("nonevict with hybrid memory manager") {
```
winningsix, line 57: The same as above.

src/test/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManagerConfigSuite.scala
```diff
@@ -0,0 +1,99 @@
...
+    val sparkEnv = SparkEnv.get
+    sparkEnv.conf.set("spark.oap.cache.strategy", "nonevict")
+    val memoryManager = MemoryManager(sparkEnv)
+    assert(memoryManager.isInstanceOf[HybridMemoryManager])
+  }
+
+  test("mix cache with offheap as index memory manager") {
```
winningsix, line 64: For mix cache, you should set cache backend and validate whether it's the right memory manager.

yma11: for mix cache, the memory manager is determined by "spark.sql.oap.mix.index.memory.manager" and "spark.sql.oap.mix.data.memory.manager". It has no relation with the backend setting from memorymanager perspective.

winningsix: Yes, I mean we should ensure those configurations are invalid when cache is decided.

winningsix:
LGTM +1

PR #1203, opened 2020-04-14
[NativeSql] DeCouple Gandiva protobuf and hashing dependency

xuechendi:
This PR is aim to decouple our dependency on gandiva protobuf and hashing.h inside arrow
Since we are intensively using this two feature and it is a little pain to keep in state with upstream arrow with cmake changes, we decided to maintain by our own of these two features.

### DIFF
--- a/oap-native-sql/cpp/src/CMakeLists.txt
+++ b/oap-native-sql/cpp/src/CMakeLists.txt
@@ -19,9 +19,39 @@ option(BENCHMARKS "Build the benchmarks" OFF)
 option(DEBUG "Enable Debug Info" OFF)
 
 find_package(JNI REQUIRED)
-
 set(source_root_directory ${CMAKE_CURRENT_SOURCE_DIR})
 
+# Gandiva protobuf
+
+find_package(Protobuf REQUIRED)
+file(MAKE_DIRECTORY ${root_directory}/src/proto)
+set(PROTO_OUTPUT_DIR "${CMAKE_CURRENT_BINARY_DIR}/proto")
+set(PROTO_OUTPUT_FILES "${PROTO_OUTPUT_DIR}/Exprs.pb.cc")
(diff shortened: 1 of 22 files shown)


zhouyuan:
1. revert the gandiva split patch
2. building patched arrow works
3. building native sql c++ lib works

The failed tests are not related

oap-native-sql/cpp/src/CMakeLists.txt
```diff
@@ -19,9 +19,39 @@ option(BENCHMARKS "Build the benchmarks" OFF)
...
+find_package(Protobuf REQUIRED)
+file(MAKE_DIRECTORY ${root_directory}/src/proto)
+set(PROTO_OUTPUT_DIR "${CMAKE_CURRENT_BINARY_DIR}/proto")
+set(PROTO_OUTPUT_FILES "${PROTO_OUTPUT_DIR}/Exprs.pb.cc")
+set(PROTO_OUTPUT_FILES ${PROTO_OUTPUT_FILES} "${PROTO_OUTPUT_DIR}/Exprs.pb.h")
+
+set_source_files_properties(${PROTO_OUTPUT_FILES} PROPERTIES GENERATED TRUE)
+
+get_filename_component(ABS_GANDIVA_PROTO ${CMAKE_CURRENT_SOURCE_DIR}/proto/Exprs.proto
+                       ABSOLUTE)
```
zhouyuan, line 35: instead of copying this file into our code layout, my initial thoughts are to define one env ARROW_CPP_DIR, and reference these files.

xuechendi:
Yes, but my concern is if once files in arrow updated, we need to fix ours as well and by copying, we can maintain by our own and even adding some optimization or new feature in future.

zhouyuan:
@xuechendi  no need to follow upstream, our release should target stable release, these codes won't change a lot

zhixingheyi-tian:
@xuechendi 

Please make sure travis can pass, before merge

Thanks

PR #1216, opened 2020-04-17
Minor updates to docs/Developer-Guide.md and docs/OAP-Architect-Overview.md

intelkevinputnam:

Small editorial changes.

## How was this patch tested?

Preview render on GitHub branch.

### DIFF
--- a/docs/Developer-Guide.md
+++ b/docs/Developer-Guide.md
@@ -1,49 +1,43 @@
 # OAP Developer Guide
 
-* [OAP Building](#OAP-Building)
-* [Integration with Spark](#integration-with-spark)
-* [Enable Numa binding for DCPMM in Spark](#enable-numa-binding-for-dcpmm-in-spark)
+* [Build OAP](#build_oap)
+* [Integrate with Spark](#integrate-with-spark)
+* [Enable NUMA binding for DCPMM in Spark](#enable-numa-binding-for-dcpmm-in-spark)
 
+## Build OAP
 
-
-## OAP Building
-
-#### Building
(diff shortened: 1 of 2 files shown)


zhixingheyi-tian:
cc @HongW2019

docs/Developer-Guide.md
```diff
@@ -1,83 +1,84 @@
...
+Add all options:
 ```
 mvn clean -q -Ppersistent-memory -Pvmemcache -DskipTests package
 ```
 
-## Integration with Spark
+## Integrate with Spark
 
-Although OAP acts as a plugin jar to Spark, there are still a few tricks to note when integration with Spark. Basically, OAP explored Spark extension & data source API to perform its core functionality. But there are other functionality aspects that cannot achieved by Spark extension and data source API. We made a few improvements or changes to the Spark internals to achieve the functionality. So when integrating OAP on Spark, you need to check whether you are running an unmodified Community Spark or a modified customized Spark.
+Although OAP acts as a plug-in `.jar` to Spark, there are still a few tricks to note when integrating with Spark. While the OAP project explored using a Spark extension and the data source API to add its core functions, some required functionality could not be achieved this way. As a result, we made some changes to Spark internals instead. Before you begin, check whether you are running an unmodified Community Spark or a customized version.
```
HongW2019, line 70: Maybe we can be more specific like this "While the OAP project explored Spark using a Spark Session extension and the data source API to add its core functions, "  :)

zhixingheyi-tian:
@intelkevinputnam 
Thanks your work.

PR #1181, opened 2020-04-03
[oap-cache/oap]Modified test case configs of oap-perf-suite

HongW2019:
## What changes were proposed in this pull request?

Add test case configs to oap-perf-suite.

## How was this patch tested?
Passed the daily test.

PR #1217, opened 2020-04-19
[oap-cache/oap]Add oap-perf-suite for oap-cache/oap

HongW2019:

 Add oap-perf-suite for oap-cache/oap, integrate #1155 and #1181


winningsix:
LGTM +1

PR #1218, opened 2020-04-20
[OAP-COMMON] rename common package

yeyuqiang:

rename common package

## How was this patch tested?

UT


yeyuqiang:
@winningsix

PR #1224, opened 2020-04-21
[OAP]Add a script to build OAP 

yao531441:
Add a script to build OAP and update pom.xml to enable build remote shuffle,


oap-cache/oap/pom.xml
```diff
@@ -22,18 +22,12 @@
     <relativePath>../../pom.xml</relativePath>
   </parent>
 
-  <artifactId>oap-cache</artifactId>
+  <artifactId>oap-cache-oap</artifactId>
```
winningsix, line 25: Any better name than ```oap-cache-oap```? It's a legacy problem at first refactor.

dev/make-distribution.sh
```diff
@@ -0,0 +1,181 @@
...
+        echo "Start building with all"
+        prepare_memkind
+        prepare_cmake
+        prepare_vmemcache
+        cd $dev_path
+        cd ..
+        mvn clean -q -Ppersistent-memory -Pvmemcache -DskipTests package
+        ;;
+        *)
+        echo "You should choose a existing options"
```
winningsix, line 176: Typo: ```an existed```

winningsix:
Could you also update the README? @yao531441

dev/make-distribution.sh
yao531441: Fixed

oap-cache/oap/pom.xml
yao531441: Do you have any advice?I have no idea how to name it better.

yao531441:
Add a README about the folder dev now.

oap-cache/oap/pom.xml
winningsix: OK, we can use it as the name for now.

winningsix:
LGTM

PR #1227, opened 2020-04-22
[NativeSql][Scala] memory leak track and fixes

xuechendi:
By this commit, observed zero inflight ColumnVector when closing operator.

### DIFF
--- a/oap-native-sql/cpp/src/codegen/arrow_compute/ext/conditioned_probe_kernel.cc
+++ b/oap-native-sql/cpp/src/codegen/arrow_compute/ext/conditioned_probe_kernel.cc
@@ -47,7 +47,7 @@ using ArrayList = std::vector<std::shared_ptr<arrow::Array>>;
 class ConditionedProbeArraysKernel::Impl {
  public:
   Impl() {}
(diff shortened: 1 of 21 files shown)


oap-native-sql/cpp/src/codegen/arrow_compute/ext/conditioned_probe_kernel.cc
```diff
@@ -47,7 +47,7 @@ using ArrayList = std::vector<std::shared_ptr<arrow::Array>>;
 class ConditionedProbeArraysKernel::Impl {
  public:
   Impl() {}
-  ~Impl() {}
+  virtual ~Impl() {}
```
zhouyuan, line 50: should we also virtual this method for other kernels?

xuechendi: @zhouyuan , you are right, I just found another class missed virtual....Also added, and I have fixed the issue on Jiayi's node, seems she is somehow testing the old code, I will merge this PR now.

PR #1230, opened 2020-04-23
[oap-native-sql]Wip refine protobuf install

zhouyuan:

This patch refines protobuf check

 - if not found, download protobuf and statically link to it
 - if found, reuse system level protobuf and dynamically link to it

## How was this patch tested?


zhouyuan:
@zhztheplayer here's the new patch to link with system level protobuf

PR #1234, opened 2020-04-24
[remote-shuffle]Reuse file handle in reduce stage

gczsjdy:

By switching on this feature, the file handles returned by Filesystem open operations will be cached/reused inside an executor(across different rounds of reduce tasks),  eliminating open overhead. This should improve the reduce stage performance only when file open operations occupy majority of the time, e.g. There is a large number of shuffle blocks, each reading a fairly small block of data, and there is no other compute in the reduce stage. Original PR @ https://github.com/Intel-bigdata/RemoteShuffle/pull/52

## How was this patch tested?

Existing tests.


gczsjdy:
cc @carsonwang

carsonwang:
Thanks!!

PR #1242, opened 2020-04-26
Rename nonevict to noevict branch 0.7

offthewall123:

Rename nonevict to noevict branch 0.7

## How was this patch tested?

### DIFF
--- a/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
+++ b/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
@@ -104,10 +104,10 @@ private[sql] object MemoryManager extends Logging {
           logWarning(s"current spark.sql.oap.fiberCache.memory.manager: ${memoryManagerOpt} " +
             "takes no effect, use 'tmp' as memory manager for vmem cache instead.")
         }
-      case "nonevict" =>
+      case "noevict" =>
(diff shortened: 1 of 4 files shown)


src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/MemoryManager.scala
```diff
@@ -104,10 +104,10 @@ private[sql] object MemoryManager extends Logging {
           logWarning(s"current spark.sql.oap.fiberCache.memory.manager: ${memoryManagerOpt} " +
             "takes no effect, use 'tmp' as memory manager for vmem cache instead.")
         }
-      case "nonevict" =>
+      case "noevict" =>
```
winningsix, line 107: Could we define those cache strategies like a Emun? So we we can maintain the code more easily?

offthewall123: Tried this.Seems not work in this scenario,cacheStrategyOpt is String and Scala only allows the same type in match...case...statements.

winningsix:
Merged. Which JIRA ID should we close?

offthewall123:
> Merged. Which JIRA ID should we close?

https://jira.devtools.intel.com/browse/POAE7-259
https://jira.devtools.intel.com/browse/POAE7-258

PR #1251, opened 2020-04-30
[NativeSql] googletest installation fixes

JiayiChen785:

Use yum install to avoid moving the header files in the previous method.


xuechendi:
@JiayiChen785 , just talked with Yuan, he will add a auto build in CMake to install googletest, so I will keep this PR until we fix googletest issue！

zhouyuan:
I think this should be fine, in the script we should always to check if there's correct system level gtest

zhouyuan:
@JiayiChen785 it looks like you accidentally used haodong's git settings, can you update this with your info?

`haodong committed 6 days ago`

PR #1256, opened 2020-05-06
[oa-native-sql][Scala] adding IsNull support

zhouyuan:

Adding IsNull Kernel 

## How was this patch tested?

manually tested


zhouyuan:
@mmoriaaa here's one simple example on adding support for different functions

PR #1258, opened 2020-05-07
init fs with table path when drop index.

ivoson:
Get fs with `path.getFileSystem(conf)` instead of `FileSystem.get(conf)` which init fs with fs.default, if the table location is not on defaut fs, we will fail droping index file.

## How was this patch tested?
have tested in baidu product env.


ivoson:
cc @zhixingheyi-tian @LuciferYang , please have a review.

ivoson:
> @ivoson create an issue, and then I will merge the patch

@LuciferYang done with issue https://github.com/Intel-bigdata/OAP/issues/1677

PR #1259, opened 2020-05-07
track distinct key set with a hash set in OapExternalSorter

ivoson:
add a hash set to track distinct key in OapExternalSorter, the original way may get wrong `countDistinct` if spill happens, as we can only get the right `countDistinct` after visit all the data.

## How was this patch tested?
add new ut.


ivoson:
cc @zhixingheyi-tian @LuciferYang , please have a review.

PR #1261, opened 2020-05-08
[oap-native-sql] Enable ColumnarSort kernel with code generation

xuechendi:
1. Use "spark.sql.columnar.sort true" to enable columnar sort
2. TPCH queries passed, multiple columns sort supported.
3. Hot function is std::sort which contributes most process time, we now observed current sort algorithm is slower than vanilar spark.

### DIFF
--- a/oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/ColumnarPlugin.scala
+++ b/oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/ColumnarPlugin.scala
@@ -3,16 +3,18 @@ package com.intel.sparkColumnarPlugin
 import com.intel.sparkColumnarPlugin.execution._
 
 import org.apache.spark.internal.Logging
+import org.apache.spark.SparkConf
 import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.execution._
-import org.apache.spark.sql.execution.RowToColumnarExec
+import org.apache.spark.sql.execution.{RowToColumnarExec, ColumnarToRowExec}
 import org.apache.spark.sql.execution.aggregate.HashAggregateExec
 import org.apache.spark.sql.execution.datasources.v2.BatchScanExec
 import org.apache.spark.sql.execution.exchange.ShuffleExchangeExec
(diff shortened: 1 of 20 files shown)


xuechendi:
All tpch quries are runable, while some queries produces incorrect result.
Correct queries are Q1, Q2, Q4, Q6, Q11, Q14, Q15, Q16, Q17, Q19, Q20, Q22
Incorrect queries are Q3, Q5, Q7, Q8, Q9, Q10, Q12, Q13, Q18, Q21
Need to look into this later

oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/ColumnarPlugin.scala
```diff
@@ -44,14 +46,16 @@ case class ColumnarPreOverrides() extends Rule[SparkPlan] {
...
-    /*case plan: SortExec =>
-      val child = replaceWithColumnarPlan(plan.child)
-      logDebug(s"Columnar Processing for ${plan.getClass} is currently supported.")
-      new ColumnarSortExec(
-        plan.sortOrder,
-        plan.global,
-        child,
-        plan.testSpillFrequency)*/
+    case plan: SortExec =>
+      if (columnarConf.enableColumnarSort) {
```
xuechendi, line 50: columnarSort is disabled by default. @zhouyuan

oap-native-sql/cpp/src/codegen/arrow_compute/ext/codegen_common.cc
```diff
@@ -0,0 +1,197 @@
...
+    std::cout << "cannot open " << cppfile << std::endl;
+    exit(EXIT_FAILURE);
+  }
+  out << codes;
+  out.flush();
+  out.close();
+
+  // compile the code
+  std::string cmd = "gcc -std=c++11 -Wall -Wextra " + cppfile + " -o " + libfile +
+                    " -O3 -shared -fPIC -larrow 2> " + logfile;
```
xuechendi, line 149: Need your fix in compiling cmd line, @zhouyuan

PR #1265, opened 2020-05-08
[DNM][oap-native-sql][Scala] fallback to row based execution on unsupported kernel

zhouyuan:

fallback to row based plan if there are unsupported columnar kernel

## How was this patch tested?
manually tested

### DIFF
--- a/oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/ColumnarPlugin.scala
+++ b/oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/ColumnarPlugin.scala
@@ -1,6 +1,10 @@
 package com.intel.sparkColumnarPlugin
 
 import com.intel.sparkColumnarPlugin.execution._
+import com.intel.sparkColumnarPlugin.expression._
+import org.apache.spark.sql.catalyst.expressions._
+
+import scala.util.control.Breaks._
 
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.rules.Rule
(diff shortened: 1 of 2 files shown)


oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/ColumnarPlugin.scala
```diff
@@ -22,17 +26,36 @@ case class ColumnarPreOverrides() extends Rule[SparkPlan] {
...
       val columnarPlan = replaceWithColumnarPlan(plan.child)
       val res = if (!columnarPlan.isInstanceOf[ColumnarConditionProjectExec]) {
-        new ColumnarConditionProjectExec(null, plan.projectList, columnarPlan)
+        val supp = checklist(plan.projectList)
+        if (supp) {
+          logDebug(s"Columnar Processing for ${plan.getClass} is currently supported.")
+          new ColumnarConditionProjectExec(null, plan.projectList, columnarPlan)
+        } else {
+          logDebug(s"Columnar Processing for ${plan.getClass} is currently not supported due to unsupported function in sub tasks")
+          plan
```
xuechendi, line 35: if we use plan here, all child plan won't be able to replace.
Same as below, we should use columnarPlan

zhouyuan: Indeed, since we fallback on proj/filter here, not sure if using row based execution will be better.

zhouyuan:
https://github.com/Intel-bigdata/OAP/pull/1549

PR #1271, opened 2020-05-09
[DNM][oap-native-sql]adding decimal support

zhouyuan:

Adding gandiva level decimal support(projection, filter)

## How was this patch tested?

manually tested

### DIFF
--- a/oap-native-sql/cpp/src/codegen/arrow_compute/ext/kernels_ext.cc
+++ b/oap-native-sql/cpp/src/codegen/arrow_compute/ext/kernels_ext.cc
@@ -382,6 +382,10 @@ class SumArrayKernel::Impl {
   } break;
       PROCESS_SUPPORTED_TYPES(PROCESS)
 #undef PROCESS
+    case arrow::Decimal128Type::type_id: {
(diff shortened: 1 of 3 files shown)


oap-native-sql/cpp/src/codegen/arrow_compute/ext/kernels_ext.cc
```diff
@@ -398,7 +402,23 @@ class SumArrayKernel::Impl {
     std::shared_ptr<arrow::Array> arr_out;
     std::shared_ptr<arrow::Scalar> scalar_out;
     scalar_out = arrow::MakeScalar(res);
-    RETURN_NOT_OK(arrow::MakeArrayFromScalar(*scalar_out.get(), 1, &arr_out));
+    arr_out = arrow::MakeArrayFromScalar(*scalar_out.get(), 1).ValueOrDie();
+    out->push_back(arr_out);
+    return arrow::Status::OK();
+  }
+
+  template <typename DataType>
```
xuechendi, line 410: If this function is only for Decimal128, we don't need to make it as a template function

PR #1272, opened 2020-05-10
[OAP] Enable to build the component of native-sql by shell

yao531441:
[OAP] Enable to build the component of native-sql by shell


JiayiChen785:
Since the function install_gcc7() is never called during the build of native-sql, exporting the absolute path of compiler set might cause trouble.

PR #1274, opened 2020-05-11
Support CDP 7.0.3

offthewall123:

Support CDP 7.0.3

## How was this patch tested?


oap-cache/oap/src/main/java/org/apache/parquet/hadoop/InternalOapRecordReader.java
```diff
@@ -35,10 +35,10 @@
 import org.apache.parquet.io.RecordReader;
 import org.apache.parquet.io.RecordReaderFactory;
 import org.apache.parquet.io.api.RecordMaterializer;
-import org.apache.parquet.it.unimi.dsi.fastutil.ints.IntList;
 import org.apache.parquet.schema.MessageType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import parquet.it.unimi.dsi.fastutil.ints.IntList;
```
winningsix, line 41: Change like this, could we just copy the code under the same package but code located in OAP?

offthewall123: No,it's only a part.I'm fixing others,not pushed yet.

PR #1275, opened 2020-05-11
Define Storage Store API

winningsix:

## How was this patch tested?


oap-common/src/main/java/com/intel/oap/common/storage/ChunkImplFactory.java
```diff
@@ -0,0 +1,18 @@
+package com.intel.oap.common.storage;
+
+import java.util.Properties;
+
+public class ChunkImplFactory {
+    public static ChunkAPI getChunkImplByType(Properties properties) {
+        ChunkType type = ChunkType.valueOf(properties.getProperty("oap.chunk.type"));
```
yma11, line 7: this oap.chunk.type only define a single type. but for compose, there should be 2 chunk types.

oap-common/src/main/java/com/intel/oap/common/storage/PMemMetaStore.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * provide trunk for output stream write, need update metadata for
+     * this stream, like chunkID++, totalsize, etc. need implement methods next()
+     * @param id
+     * @param chunkSize
+     * @return
+     */
+    Iterator<PMemBlock> getOutputChunkIterator();
+
```
yma11, line 33: If we support compose at metastore level, here we can't return an Iterator<PMemBlock>, we can return a Iterator<Trunk> and PMemBlock, FileTrunk extend Trunk?

oap-common/src/main/java/com/intel/oap/common/storage/PMemMetaStore.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     *
+     * @param id logical ID
+     * @return
+     */
+    Iterator<PMemBlock> getInputChunkIterator();
+
+    /**
+     * provide trunk for output stream write, need update metadata for
```
yeyuqiang, line 26: Why we should get chunk for output from meta store? If meta store already maitain the chunk info, then the chunk should be already store in pmem. I think we only need getChunkIterator.

oap-common/src/main/java/com/intel/oap/common/storage/StreamMeta.java
```diff
@@ -0,0 +1,11 @@
...
+package com.intel.oap.common.storage;
+
+public class StreamMeta {
+    // total size of bytes in this stream
+    public long totalSize;
+    // number of chunks in this stream
+    public int numOfChunks;
+    // need compose or not
+    boolean isComplete;
+}
```
yeyuqiang, line 11: Will we use stream meta to lookup chunks in pmem? If so, does these variable enough to do it? Or we need a list to record each chunks' address?

oap-common/src/main/java/com/intel/oap/common/storage/ChunkAPI.java
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * action when read done of current Chunk
+     */
+    public void release();
+
+    /**
+     * free pmem space when chunk life cycle end
+     * @param pMemBlock
+     */
+    public void free(PMemBlock pMemBlock);
```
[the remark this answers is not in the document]
winningsix: @yma11 Any comments here? I am still not following you well why we don't need id to free.

oap-common/src/main/java/com/intel/oap/common/storage/ChunkInputStream.java
```diff
@@ -0,0 +1,14 @@
+package com.intel.oap.common.storage;
+
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+
+public class ChunkInputStream extends FileInputStream {
```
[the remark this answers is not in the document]
winningsix: As last discussed, we will not reuse code but for minimize the code changes in Spark.

oap-common/src/main/java/com/intel/oap/common/storage/ChunkAPI.java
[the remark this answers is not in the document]
yma11: I have replied it in previous PR.  We should use ID to free here.  Current methodis not correct.

oap-common/src/main/java/com/intel/oap/common/storage/ChunkInputStream.java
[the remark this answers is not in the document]
yeyuqiang: I mean we should not inherit FileInput/OutputStream. That's not related code changes in Spark, in spark side(at least RDD Cache) , InputStream/OutputStream is enough

oap-common/src/main/java/com/intel/oap/common/storage/PMemChunk.java
```diff
@@ -0,0 +1,16 @@
+package com.intel.oap.common.storage;
+
+public class PMemChunk {
+    ChunkAPI chunkAPI;
+    byte[] id;
```
yma11, line 5: Shouldn't ID is already maintained in the ChunkAPI?

oap-common/src/main/java/com/intel/oap/common/storage/SPDKPMemDataStoreImpl.java
```diff
@@ -0,0 +1,30 @@
+package com.intel.oap.common.storage;
+
+import java.util.Iterator;
+
+public class SPDKPMemDataStoreImpl implements PMemDataStore {
```
yeyuqiang, line 5: extends instead of implements?

oap-common/src/test/java/com/intel/oap/common/storage/ChunkStreamAPITest.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.Properties;
+
+public class ChunkStreamAPITest {
+
+    @Test
```
yeyuqiang, line 11: We'd better ignore it if this ut depend on the implementation to make the PR pass the travis

PR #1288, opened 2020-05-15
[DataSource][Arrow] Close batch every time new batch is read to avoid…

zhztheplayer:
… possible leaks


zhouyuan:
ArrowDataSourceTPCHBasedTest.scala is this file missed in your commit?

zhztheplayer:
@zhouyuan the file is in the commit... Is there something wrong with it?

PR #1297, opened 2020-05-20
[oap-native-sql][C++] pass arrow/gcc  to runtime filter codegen

zhouyuan:
Make codegen work with customized arrow installation


zhouyuan:
fixes #1311

PR #1307, opened 2020-05-23
[oap-native-sql][Scala]adding if support

mmoriaaa:

### DIFF
--- a/oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/expression/ColumnarIfOperator.scala
+++ b/oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/expression/ColumnarIfOperator.scala
@@ -1,0 +1,41 @@
+package com.intel.sparkColumnarPlugin.expression
+
+import com.google.common.collect.Lists
+import com.google.common.collect.Sets
+import org.apache.arrow.gandiva.evaluator._
+import org.apache.arrow.gandiva.exceptions.GandivaException
+import org.apache.arrow.gandiva.expression._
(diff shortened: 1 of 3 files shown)


oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/expression/ColumnarIfOperator.scala
```diff
@@ -0,0 +1,41 @@
...
+    with Logging {
+  override def doColumnarCodeGen(args: java.lang.Object): (TreeNode, ArrowType) = {
+      val (predicate_node, predicateType): (TreeNode, ArrowType) =
+      predicate.asInstanceOf[ColumnarExpression].doColumnarCodeGen(args)
+      val (true_node, trueType): (TreeNode, ArrowType) =
+        trueValue.asInstanceOf[ColumnarExpression].doColumnarCodeGen(args)
+      val (false_node, falseType): (TreeNode, ArrowType) =
+        falseValue.asInstanceOf[ColumnarExpression].doColumnarCodeGen(args)
+
+      val funcNode = TreeBuilder.makeIf(predicate_node, true_node, false_node, trueType)
```
zhouyuan, line 27: not sure if this is necessary, will there be some issue if trueType != falseType?

mmoriaaa: this patch was used to solve if support error for q41, like: if (isnull(alwaysTrue#863)) 0 else item_cnt#850L. For now, we seemed to ignore the "trueType != falseType" issue in caseWhen and coalesce.

PR #1313, opened 2020-05-25
POAE7-319: reader/writer for memkind

yma11:

Implement reader/writer for memkind.

Tracked by JIRA ticket POAE7-319.

## How was this patch tested?

not tested yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ jobs:
       - ./install_memkind.sh
       - ../oap-common/src/native/memkind/compile.sh
       - sudo cp ../oap-common/src/resources/linux/64/lib/libpmplatform.so /usr/lib/
+      - sudo ldconfig /usr/lib
       script:
       - cd ${TRAVIS_BUILD_DIR}
       - mvn clean -q -Ppersistent-memory -Pvmemcache test
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ jobs:
       - cd ${TRAVIS_BUILD_DIR}/dev
       - ./install_vmemcache.sh
       - ./install_memkind.sh
+      - ../oap-common/src/native/memkind/compile.sh
+      - sudo cp ../oap-common/src/resources/linux/64/lib/libpmplatform.so /usr/lib/
       script:
       - cd ${TRAVIS_BUILD_DIR}
       - mvn clean -q -Ppersistent-memory -Pvmemcache test
--- a/oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
+++ b/oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
@@ -1,6 +1,10 @@
 package com.intel.oap.common.storage.memkind;
 
-import com.intel.oap.common.storage.stream.*;
+import com.intel.oap.common.storage.stream.ChunkReader;
+import com.intel.oap.common.storage.stream.MetaData;
+import com.intel.oap.common.storage.stream.PMemManager;
+import com.intel.oap.common.storage.stream.PMemMetaStore;
+import com.intel.oap.common.storage.stream.PMemPhysicalAddress;
 import com.intel.oap.common.unsafe.PersistentMemoryPlatform;
 import com.intel.oap.common.util.MemCopyUtil;
 import org.slf4j.Logger;
@@ -31,6 +35,7 @@ public class MemkindChunkReader extends ChunkReader {
 
     @Override
     protected void freeFromPMem() {
+        // TODO: may need refactor based on up-level usage
         // get metaFooter for this logicalID
         MetaData metaData = pMemMetaStore.getMetaFooter(logicalID);
         int currentTrunkID = 0;
--- a/oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
+++ b/oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
@@ -3,6 +3,7 @@ package com.intel.oap.common.storage.memkind;
 import com.intel.oap.common.storage.stream.PMemManager;
 import com.intel.oap.common.storage.stream.ChunkReader;
 import com.intel.oap.common.storage.stream.PMemPhysicalAddress;
+import com.intel.oap.common.util.MemCopyUtil;
 
 import java.nio.ByteBuffer;
 
@@ -12,7 +13,10 @@ public class MemkindChunkReader extends ChunkReader {
     }
 
     @Override
-    protected int readFromPMem(PMemPhysicalAddress id, ByteBuffer data) {
-        return 0;
+    protected int readFromPMem(PMemPhysicalAddress pMemPhysicalAddress, ByteBuffer data) {
+        long baseAddress = ((MemkindPMemPhysicalAddress) pMemPhysicalAddress).getBaseAddress();
+        int length = ((MemkindPMemPhysicalAddress) pMemPhysicalAddress).getOffset();
+        MemCopyUtil.copyMemory(null, baseAddress, data, data.arrayOffset(), length);
+        return length;
     }
 }
--- a/oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
+++ b/oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
@@ -1,13 +1,19 @@
 package com.intel.oap.common.storage.memkind;
 
+import com.intel.oap.common.storage.stream.MetaData;
 import com.intel.oap.common.storage.stream.PMemManager;
 import com.intel.oap.common.storage.stream.ChunkReader;
 import com.intel.oap.common.storage.stream.PMemPhysicalAddress;
+import com.intel.oap.common.unsafe.PersistentMemoryPlatform;
 import com.intel.oap.common.util.MemCopyUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.io.File;
 import java.nio.ByteBuffer;
 
 public class MemkindChunkReader extends ChunkReader {
+    private static final Logger logger = LoggerFactory.getLogger(MemkindChunkReader.class);
     public MemkindChunkReader(byte[] logicalID, PMemManager pMemManager) {
         super(logicalID, pMemManager);
     }
@@ -18,5 +24,26 @@ public class MemkindChunkReader extends ChunkReader {
         int length = ((MemkindPMemPhysicalAddress) pMemPhysicalAddress).getOffset();
         MemCopyUtil.copyMemory(null, baseAddress, data, data.arrayOffset(), length);
         return length;
+    }
+
+    @Override
+    protected void freeFromPMem(byte[] logicalID) {
+        // get metaFooter for this logicalID
+        MetaData metaData = pMemManager.getpMemMetaStore().getMetaFooter(logicalID);
+        int currentTrunk = 0;
+        while(currentTrunk < metaData.getTotalChunk()) {
+            PMemPhysicalAddress pMemPhysicalAddress = pMemManager.getpMemMetaStore().getPhysicalAddressByID(logicalID, currentTrunk);
+            PersistentMemoryPlatform.freeMemory(((MemkindPMemPhysicalAddress) pMemPhysicalAddress).getBaseAddress());
+            pMemManager.getpMemMetaStore().removePhysicalAddress(logicalID, currentTrunk, pMemPhysicalAddress);
+        }
+        if(metaData.isHasDiskData() == true) {
+            // delete file
+            File file = new File(new String(logicalID));
+            if (file != null && file.exists()) {
+                if (!file.delete()) {
+                    logger.error("Was unable to delete file {}", file.getAbsolutePath());
+                }
+            }
+        }
     }
 }
--- a/oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
+++ b/oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
@@ -1,49 +1,52 @@
 package com.intel.oap.common.storage.memkind;
 
-import com.intel.oap.common.storage.stream.MetaData;
-import com.intel.oap.common.storage.stream.PMemManager;
-import com.intel.oap.common.storage.stream.ChunkReader;
-import com.intel.oap.common.storage.stream.PMemPhysicalAddress;
+import com.intel.oap.common.storage.stream.*;
 import com.intel.oap.common.unsafe.PersistentMemoryPlatform;
 import com.intel.oap.common.util.MemCopyUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import sun.misc.Unsafe;
 
 import java.io.File;
 import java.nio.ByteBuffer;
 
 public class MemkindChunkReader extends ChunkReader {
     private static final Logger logger = LoggerFactory.getLogger(MemkindChunkReader.class);
+    private PMemMetaStore pMemMetaStore;
     public MemkindChunkReader(byte[] logicalID, PMemManager pMemManager) {
         super(logicalID, pMemManager);
+        this.pMemMetaStore = pMemManager.getpMemMetaStore();
     }
 
     @Override
     protected int readFromPMem(PMemPhysicalAddress pMemPhysicalAddress, ByteBuffer data) {
-        long baseAddress = ((MemkindPMemPhysicalAddress) pMemPhysicalAddress).getBaseAddress();
-        int length = ((MemkindPMemPhysicalAddress) pMemPhysicalAddress).getOffset();
-        MemCopyUtil.copyMemory(null, baseAddress, data, data.arrayOffset(), length);
+        MemkindPMemPhysicalAddress memkindPMemPhysicalAddress = (MemkindPMemPhysicalAddress) pMemPhysicalAddress;
+        long baseAddress = memkindPMemPhysicalAddress.getBaseAddress();
+        int length = memkindPMemPhysicalAddress.getOffset();
+        byte[] buffer = new byte[length];
+        MemCopyUtil.copyMemory(null, baseAddress, buffer, Unsafe.ARRAY_BYTE_BASE_OFFSET, length);
+        data.put(buffer);
         return length;
     }
 
     @Override
-    protected void freeFromPMem(byte[] logicalID) {
+    protected void freeFromPMem() {
         // get metaFooter for this logicalID
-        MetaData metaData = pMemManager.getpMemMetaStore().getMetaFooter(logicalID);
-        int currentTrunk = 0;
-        while(currentTrunk < metaData.getTotalChunk()) {
-            PMemPhysicalAddress pMemPhysicalAddress = pMemManager.getpMemMetaStore().getPhysicalAddressByID(logicalID, currentTrunk);
+        MetaData metaData = pMemMetaStore.getMetaFooter(logicalID);
+        int currentTrunkID = 0;
+        while(currentTrunkID < metaData.getTotalChunk()) {
+            PMemPhysicalAddress pMemPhysicalAddress = pMemMetaStore.getPhysicalAddressByID(logicalID, currentTrunkID);
             PersistentMemoryPlatform.freeMemory(((MemkindPMemPhysicalAddress) pMemPhysicalAddress).getBaseAddress());
-            pMemManager.getpMemMetaStore().removePhysicalAddress(logicalID, currentTrunk, pMemPhysicalAddress);
+            pMemMetaStore.removePhysicalAddress(logicalID, currentTrunkID, pMemPhysicalAddress);
+            currentTrunkID++;
         }
-        if(metaData.isHasDiskData() == true) {
-            // delete file
-            File file = new File(new String(logicalID));
-            if (file != null && file.exists()) {
-                if (!file.delete()) {
-                    logger.error("Was unable to delete file {}", file.getAbsolutePath());
-                }
+        // delete file
+        File file = new File(new String(logicalID));
+        if (file != null && file.exists()) {
+            if (!file.delete()) {
+                logger.error("Was unable to delete file {}", file.getAbsolutePath());
             }
         }
+        pMemMetaStore.removeMetaFooter(logicalID);
     }
 }
--- a/oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkWriter.java
+++ b/oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkWriter.java
@@ -5,6 +5,7 @@ import com.intel.oap.common.storage.stream.ChunkWriter;
 import com.intel.oap.common.storage.stream.PMemPhysicalAddress;
 import com.intel.oap.common.unsafe.PersistentMemoryPlatform;
 import com.intel.oap.common.util.MemCopyUtil;
+import sun.misc.Unsafe;
 
 import java.nio.ByteBuffer;
 
@@ -18,8 +19,8 @@ public class MemkindChunkWriter extends ChunkWriter {
         int dataSizeInByte = byteBuffer.position();
         long baseAddr = PersistentMemoryPlatform.allocateVolatileMemory(dataSizeInByte);
         MemkindPMemPhysicalAddress pMemPhysicalAddress = new MemkindPMemPhysicalAddress(baseAddr, dataSizeInByte);
-        // write the bytebuffer to PMem
-        MemCopyUtil.copyMemory(byteBuffer, byteBuffer.arrayOffset(), null, baseAddr, dataSizeInByte);
+        // write the byte buffer to PMem
+        MemCopyUtil.copyMemory(byteBuffer.array(), Unsafe.ARRAY_BYTE_BASE_OFFSET, null, baseAddr, dataSizeInByte);
         return  pMemPhysicalAddress;
     }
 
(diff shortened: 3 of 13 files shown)


oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemKindPMemBlock.java
```diff
@@ -0,0 +1,19 @@
+package com.intel.oap.common.storage.memkind;
+
+public class MemKindPMemBlock {
+    private long baseAddress;
+    private int offset;
```
jikunshang, line 5: `offset` here equals to length?

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkWriter.java
```diff
@@ -12,13 +13,19 @@ public MemkindChunkWriter(byte[] logicalID, PMemManager pMemManager) {
     }
 
     @Override
-    protected PMemID writeInternal(ByteBuffer byteBuffer) {
-        //TODO
-        return null;
+    protected void writeInternal(ByteBuffer byteBuffer) {
+        String id = super.pMemManager.getpMemMetaStore().getPMemIDByLogicalID(super.logicalID, super.chunkID);
+        int dataSizeInByte = byteBuffer.position();
+        long baseAddr = PersistentMemoryPlatform.allocateVolatileMemory(dataSizeInByte);
```
jikunshang, line 19: Have we considered when to call ` PersistentMemoryPlatform.free` ? cc @winningsix @yeyuqiang

yeyuqiang: Current design did not include remove/free phase

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemKindPMemBlock.java
yma11: Yes. It indicates the actual data size in this MemBlock.

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkWriter.java
yma11: From upper level, Pmem space should be freed when the files(current in spark) are deleted. We may design it separately from inputstream/output stream.

oap-common/src/main/java/com/intel/oap/common/storage/stream/ChunkWriter.java
```diff
@@ -46,10 +46,10 @@ private void flushBufferByChunk(ByteBuffer byteBuffer) throws IOException {
         int dataSizeInByte = byteBuffer.position();
         if (!fallbackTriggered && pMemManager.getStats().getRemainingSize() > dataSizeInByte) {
             try {
-                PMemID id = writeInternal(byteBuffer);
+                writeInternal(byteBuffer);
                 chunkID++;
                 pMemManager.getStats().increaseSize(dataSizeInByte);
-                pMemManager.getpMemMetaStore().putPMemID(logicalID, chunkID, id);
+                // pMemManager.getpMemMetaStore().putPMemID(logicalID, chunkID, id);
```
yeyuqiang, line 52: Is that possible to make writeInternal return a PMemBlock. then we can use putPMemBlock here?

oap-common/src/main/java/com/intel/oap/common/util/MemCopyUtil.java
```diff
@@ -0,0 +1,241 @@
...
+    static {
+        sun.misc.Unsafe unsafe;
+        try {
+            Field unsafeField = Unsafe.class.getDeclaredField("theUnsafe");
+            unsafeField.setAccessible(true);
+            unsafe = (sun.misc.Unsafe) unsafeField.get(null);
+        } catch (Throwable cause) {
+            unsafe = null;
+        }
+        _UNSAFE = unsafe;
```
yeyuqiang, line 220: Currently we should only need L211 - L220 and copyMemory method. No need to involve too much unnecessary code since we are borrowing from spark Platform

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkWriter.java
```diff
@@ -12,13 +13,19 @@ public MemkindChunkWriter(byte[] logicalID, PMemManager pMemManager) {
...
+        MemKindPMemBlock memKindPMemBlock = new MemKindPMemBlock(baseAddr, dataSizeInByte);
```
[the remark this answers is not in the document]
yma11: Probably. In that case we can change it to PMemBlock and share it.

oap-common/src/main/java/com/intel/oap/common/storage/stream/ChunkWriter.java
yma11: We can do this if assure PMemblock can be a common obj shared among Memkind/PMDK, etc. But I think do this put action once after the allocate happen also should be fine.

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkWriter.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Override
+    protected void writeInternal(ByteBuffer byteBuffer) {
+        String id = super.pMemManager.getpMemMetaStore().getPMemIDByLogicalID(super.logicalID, super.chunkID);
+        int dataSizeInByte = byteBuffer.position();
+        long baseAddr = PersistentMemoryPlatform.allocateVolatileMemory(dataSizeInByte);
+        MemKindPMemBlock memKindPMemBlock = new MemKindPMemBlock(baseAddr, dataSizeInByte);
+        // add this new allocated block in the map
+        ((MemkindMetaStore)super.pMemManager.getpMemMetaStore()).putPMemBlock(id, memKindPMemBlock);
```
yeyuqiang, line 22: Is it enough to only record the mapping relationship between physical id and pmem block (Here seems just a chunk). For example, RDD Cache, RDDBlock may 100+MB, our chunk may 1MB. How ChunkInputStream know which chunks belong to this RDD Block

oap-common/src/main/java/com/intel/oap/common/storage/stream/PMemMetaStore.java
```diff
@@ -1,11 +1,19 @@
 package com.intel.oap.common.storage.stream;
 
+import com.intel.oap.common.storage.memkind.MemKindPMemBlock;
```
winningsix, line 3: We should avoid depending anything from Memkind.

oap-common/src/main/java/com/intel/oap/common/storage/stream/PMemMetaStore.java
```diff
@@ -1,11 +1,19 @@
...
+
 public interface PMemMetaStore {
-     PMemID getPMemIDByLogicalID(byte[] id, int chunkID);
+     /**
+      * compose physical ID based on logical ID and chunkID
+      * @param id
+      * @param chunkID
+      * @return
+      */
+     String getPMemIDByLogicalID(byte[] id, int chunkID);
```
winningsix, line 12: cc @yeyuqiang Is string enough for you?

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkWriter.java
```diff
@@ -12,13 +13,19 @@ public MemkindChunkWriter(byte[] logicalID, PMemManager pMemManager) {
...
-    protected PMemID writeInternal(ByteBuffer byteBuffer) {
-        //TODO
-        return null;
+    protected void writeInternal(ByteBuffer byteBuffer) {
+        String id = pMemManager.getpMemMetaStore().getPMemIDByLogicalID(logicalID, chunkID);
+        int dataSizeInByte = byteBuffer.position();
+        long baseAddr = PersistentMemoryPlatform.allocateVolatileMemory(dataSizeInByte);
+        MemKindPMemBlock memKindPMemBlock = new MemKindPMemBlock(baseAddr, dataSizeInByte);
+        // add this new allocated block in the map
+        ((MemkindMetaStore) pMemManager.getpMemMetaStore()).putPMemBlock(id, memKindPMemBlock);
```
winningsix, line 22: Why do we remove this method from the parent class? Not follow this part. BTW, let's move L22 back to parent class so that other implementation can avoid copying the same code there.

winningsix, line 22: The original implementation should be cleaner. BTW, getPMemIDByLogicalID is not a util method. It's used to get the metadata by ID. The metadata is PMemID which is a string in your case.

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindMetaStore.java
```diff
@@ -1,33 +1,36 @@
...
 
 //TODO design point, how to store data on PMEM
 public class MemkindMetaStore implements PMemMetaStore {
-    ConcurrentHashMap<String, PMemID> PMemHashMap = new ConcurrentHashMap();
+    ConcurrentHashMap<String, MemKindPMemBlock> pMemHashMap = new ConcurrentHashMap();
     ConcurrentHashMap<String, MetaData> metaHashMap = new ConcurrentHashMap();
 
     @Override
-    public PMemID getPMemIDByLogicalID(byte[] id, int chunkID) {
+    public String getPMemIDByLogicalID(byte[] id, int chunkID) {
```
winningsix, line 14: I guess you misunderstand the method.

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkWriter.java
yma11: There are two mappings maintained. one is <PMemID, PMemBlock> and the other one is <logicalID, MetaData>. In the Metadata, it will record total trunks and whether spilled to disk.

oap-common/src/main/java/com/intel/oap/common/storage/stream/PMemMetaStore.java
yma11: This line can be deleted. MemkindPMemBlock is not actually used.

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkWriter.java
yma11: Will do it if the mapping <string PMemID, PMemBlock> also work in PMDK. @yeyuqiang

yma11: So what getMetaFooter() do? My understanding is there are 2 mappings. One to track <PMemID(physicalID), PMemBlock> and the other one is <logicalID, metaData>, right?

yma11:
@winningsix @yeyuqiang @jikunshang rebased. Please take a review.

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
```diff
@@ -1,18 +1,49 @@
...
         super(logicalID, pMemManager);
     }
 
     @Override
-    protected int readFromPMem(PMemPhysicalAddress id, ByteBuffer data) {
-        return 0;
+    protected int readFromPMem(PMemPhysicalAddress pMemPhysicalAddress, ByteBuffer data) {
+        long baseAddress = ((MemkindPMemPhysicalAddress) pMemPhysicalAddress).getBaseAddress();
+        int length = ((MemkindPMemPhysicalAddress) pMemPhysicalAddress).getOffset();
+        MemCopyUtil.copyMemory(null, baseAddress, data, data.arrayOffset(), length);
```
yeyuqiang, line 25: I am not sure whether copying to bytebuffer directly will work . Or here should create a byte array first, copy to the byte array, put this byte array to bytebuffer？ If we haven't tested yet, we can write some unit test after merged and move to elastic data fusion.

yeyuqiang:
LGTM overall.

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
```diff
@@ -1,18 +1,49 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(MemkindChunkReader.class);
     public MemkindChunkReader(byte[] logicalID, PMemManager pMemManager) {
         super(logicalID, pMemManager);
     }
 
     @Override
-    protected int readFromPMem(PMemPhysicalAddress id, ByteBuffer data) {
-        return 0;
+    protected int readFromPMem(PMemPhysicalAddress pMemPhysicalAddress, ByteBuffer data) {
+        long baseAddress = ((MemkindPMemPhysicalAddress) pMemPhysicalAddress).getBaseAddress();
```
winningsix, line 23: Let's do the convert once and reuse variable from that.

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
```diff
@@ -1,18 +1,49 @@
...
+        MemCopyUtil.copyMemory(null, baseAddress, data, data.arrayOffset(), length);
+        return length;
+    }
+
+    @Override
+    protected void freeFromPMem(byte[] logicalID) {
+        // get metaFooter for this logicalID
+        MetaData metaData = pMemManager.getpMemMetaStore().getMetaFooter(logicalID);
+        int currentTrunk = 0;
+        while(currentTrunk < metaData.getTotalChunk()) {
```
winningsix, line 34: will this loop exit? Didn't see any updates to ```currentTrunk```. And we should update it with physical ID instead.

oap-common/src/main/java/com/intel/oap/common/storage/stream/ChunkReader.java
```diff
@@ -65,5 +65,13 @@ private int readFromDisk(ByteBuffer remainingBuffer) throws IOException {
...
 
+    /**
+     * read data from a address mappted to physical ID
+     * @param id
+     * @param data
+     * @return
+     */
     protected abstract int readFromPMem(PMemPhysicalAddress id, ByteBuffer data);
+
+    protected abstract void freeFromPMem(byte[] logicalID);
```
winningsix, line 76: We should use ```PhysicalAddress``` here instead of ```logicalID```.

yma11: If PhysicalAddress used here, then we will need provide interface to delete the full file(logicalID) some where else, like in the chunkInputStream?

winningsix: Yes, we need.

winningsix:
Thanks @yma11 for the updates. Please help take a look at the failed unit tests.
```
Caused by: java.lang.RuntimeException: Can't load library: pmplatform from jar.
	at com.intel.oap.common.util.NativeLibraryLoader.loadFromPackage(NativeLibraryLoader.java:111)
	at com.intel.oap.common.util.NativeLibraryLoader.load(NativeLibraryLoader.java:74)
	at com.intel.oap.common.unsafe.PersistentMemoryPlatform.<clinit>(PersistentMemoryPlatform.java:40)
	... 31 more
Caused by: java.lang.UnsatisfiedLinkError: /tmp/pmplatform1114415800134548548.tmp: libmemkind.so.0: cannot open shared object file: No such file or directory
	at java.lang.ClassLoader$NativeLibrary.load(Native Method)
	at java.lang.ClassLoader.loadLibrary0(ClassLoader.java:1941)
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1824)
	at java.lang.Runtime.load0(Runtime.java:809)
	at java.lang.System.load(System.java:1086)
	at com.intel.oap.common.util.NativeLibraryLoader.loadFromPackage(NativeLibraryLoader.java:109)
	... 33 more
```

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
```diff
@@ -1,18 +1,52 @@
 package com.intel.oap.common.storage.memkind;
 
-import com.intel.oap.common.storage.stream.PMemManager;
-import com.intel.oap.common.storage.stream.ChunkReader;
-import com.intel.oap.common.storage.stream.PMemPhysicalAddress;
+import com.intel.oap.common.storage.stream.*;
```
winningsix, line 3: Please only import the needed package.

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
```diff
@@ -1,18 +1,52 @@
...
+        long baseAddress = memkindPMemPhysicalAddress.getBaseAddress();
+        int length = memkindPMemPhysicalAddress.getOffset();
+        byte[] buffer = new byte[length];
+        MemCopyUtil.copyMemory(null, baseAddress, buffer, Unsafe.ARRAY_BYTE_BASE_OFFSET, length);
+        data.put(buffer);
+        return length;
+    }
+
+    @Override
+    protected void freeFromPMem() {
```
winningsix, line 33: Please put a TODO here since we will refactor it later once we figure out how upper layer uses it.

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkReader.java
```diff
@@ -1,18 +1,52 @@
...
     @Override
-    protected int readFromPMem(PMemPhysicalAddress id, ByteBuffer data) {
-        return 0;
+    protected int readFromPMem(PMemPhysicalAddress pMemPhysicalAddress, ByteBuffer data) {
+        MemkindPMemPhysicalAddress memkindPMemPhysicalAddress = (MemkindPMemPhysicalAddress) pMemPhysicalAddress;
+        long baseAddress = memkindPMemPhysicalAddress.getBaseAddress();
+        int length = memkindPMemPhysicalAddress.getOffset();
+        byte[] buffer = new byte[length];
+        MemCopyUtil.copyMemory(null, baseAddress, buffer, Unsafe.ARRAY_BYTE_BASE_OFFSET, length);
+        data.put(buffer);
```
winningsix, line 28: Could we use a direct buffer here to avoid some GC pressure?

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkWriter.java
```diff
@@ -13,12 +16,16 @@ public MemkindChunkWriter(byte[] logicalID, PMemManager pMemManager) {
...
+        MemkindPMemPhysicalAddress pMemPhysicalAddress = new MemkindPMemPhysicalAddress(baseAddr, dataSizeInByte);
+        // write the byte buffer to PMem
+        MemCopyUtil.copyMemory(byteBuffer.array(), Unsafe.ARRAY_BYTE_BASE_OFFSET, null, baseAddr, dataSizeInByte);
+        return  pMemPhysicalAddress;
     }
 
     @Override
     protected void closeInternal() {
-        //TODO
+        //do nothing here for MemKind
```
winningsix, line 29: Remove this method since we did nothing.

oap-common/src/test/java/com/intel/oap/common/storage/ChunkStreamAPITest.java
```diff
@@ -38,4 +41,179 @@ public void testFileStreamReadWrite() throws IOException {
...
+        Properties p = new Properties();
+        p.put("totalSize", "1024");
+        p.put("chunkSize", "6");
+        PMemManager pMemManager = new PMemManager(p);
+        dataStore = new DataStore(pMemManager);
+        String initialPath = "target/tmp/";
+        String fileName = "target/pmem.file";
+        byte[] data = new byte[]{'a', 'b', 'c'};
+        PersistentMemoryPlatform.initialize(initialPath, 16 * 1024 *1024, 0);
+        ChunkOutputStream chunkOutputStream = new ChunkOutputStream(fileName, dataStore);
```
winningsix, line 76: Could we create a private method and reuse the code for each UTs?

oap-common/src/main/java/com/intel/oap/common/storage/memkind/MemkindChunkWriter.java
yma11: seems can't be removed since MemkindChunkWriter is no longer an abstract class and need implements this abstract method from parent.

.travis.yml
```diff
@@ -19,6 +19,8 @@ jobs:
       - cd ${TRAVIS_BUILD_DIR}/dev
       - ./install_vmemcache.sh
       - ./install_memkind.sh
+      - ../oap-common/src/native/memkind/compile.sh
+      - sudo cp ../oap-common/src/resources/linux/64/lib/libpmplatform.so /usr/lib/
```
jikunshang, line 23: try copy to `/usr/lib64`

yma11: Thanks for comment. /usr/lib doesn't work. let me try /usr/lib64.

oap-common/src/main/java/com/intel/oap/common/storage/stream/ChunkOutputStream.java
```diff
@@ -51,6 +51,7 @@ public void write(byte b[], int off, int len) throws IOException {
      */
     public void close() throws IOException {
         chunkWriter.close();
+        super.close();
```
winningsix, line 54: Usually super should be called as 1st statement.

oap-common/src/main/java/com/intel/oap/common/storage/stream/ChunkWriter.java
```diff
@@ -36,6 +36,7 @@ public void write(byte[] bytes) throws IOException {
                 // clear content of remainingBuffer
                 remainingBuffer.clear();
             }
+            // System.out.println("remainingbuffer" + remainingBuffer.capacity());
```
winningsix, line 39: Remove this or adding debug log.

oap-common/src/main/java/com/intel/oap/common/storage/stream/ChunkWriter.java
```diff
@@ -51,11 +52,13 @@ private void flushBufferByChunk(ByteBuffer byteBuffer) throws IOException {
         int dataSizeInByte = byteBuffer.position();
         if (!fallbackTriggered && pMemManager.getStats().getRemainingSize() >= dataSizeInByte) {
             try {
+                // System.out.println("writing to pmem" );
```
winningsix, line 55: Remove this or adding debug log.

oap-common/src/main/java/com/intel/oap/common/storage/stream/ChunkWriter.java
```diff
@@ -51,11 +52,13 @@ private void flushBufferByChunk(ByteBuffer byteBuffer) throws IOException {
...
                 PMemPhysicalAddress id = writeInternal(byteBuffer);
                 pMemManager.getStats().increaseSize(dataSizeInByte);
                 pMemManager.getpMemMetaStore().putPhysicalAddress(logicalID, chunkID, id);
                 chunkID++;
             } catch (RuntimeException re) {
+                // System.out.println("remaining" + pMemManager.getStats().getRemainingSize() );
```
winningsix, line 61: Remove this or adding debug log.

oap-common/src/test/java/com/intel/oap/common/storage/ChunkStreamAPITest.java
```diff
@@ -25,8 +26,20 @@ public void prepare() {
...
     }
 
-    private Boolean testChunkStream(String totalSize, String chunkSize, byte[] data,
-                                             int expectedTotalChunk, boolean expectedHasDiskData) throws IOException {
+    private boolean loadPmemLib() {
+        String LIBNAME = "pmplatform";
+        try {
+            NativeLibraryLoader.load(LIBNAME);
+        } catch (UnsatisfiedLinkError | RuntimeException e) {
+            System.out.println("pmplatform lib failed to load, PMem Stream Test will be skipped.");
```
winningsix, line 34: Use log instead.

oap-common/src/test/java/com/intel/oap/common/storage/ChunkStreamAPITest.java
```diff
@@ -25,8 +26,20 @@ public void prepare() {
...
+            libAvailable = false;
+        }
+        return libAvailable;
+    }
+
+    private Boolean testChunkStream(String totalSize, String chunkSize,
```
winningsix, line 40: code style: indents

oap-common/src/test/java/com/intel/oap/common/storage/ChunkStreamAPITest.java
```diff
@@ -72,42 +85,49 @@ public void testPMemStreamEmpty() throws IOException {
 
     @Test
     public void testFileStreamReadWrite() throws IOException {
+        assumeTrue(loadPmemLib());
```
winningsix, line 88: why not adding this into before all method? It seems all UTs are calling this method.

winningsix:
A few more comments left. Overall it's in good sharp.

oap-common/src/test/java/com/intel/oap/common/storage/ChunkStreamAPITest.java
yma11: The test case you wrote to test write file directly doesn't need load pmem lib.

oap-common/src/main/java/com/intel/oap/common/storage/stream/ChunkOutputStream.java
yma11: we can't do this since need call chunkwriter.close() to write last chunk into the outputstream before closing it.

PR #1315, opened 2020-05-25
[remote-shuffle] Hadoop shuffle I/O plugin for Spark 3.0

gczsjdy:

Added Spark 3.0 version remote-shuffle.

## How was this patch tested?

Added UTs.


oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleMapOutputWriter.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    def getCount: Int = count
+
+    override def write(b: Int): Unit = {
+      verifyNotClosed()
+      bufferedOutputStream.write(b)
+      count += 1
+    }
+
+    override def write(buf: Array[Byte], pos: Int, length: Int): Unit = {
```
carsonwang, line 107: also override `def write(byte[] b)`?

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleMapOutputWriter.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  class PartitionWriterStream(val partitionId: Int) extends OutputStream {
+    private var count = 0
+    private var isClosed = false
+
+    def getCount: Int = count
+
+    override def write(b: Int): Unit = {
+      verifyNotClosed()
+      bufferedOutputStream.write(b)
```
carsonwang, line 103: can you pass the `bufferedOutputStream` in as a class parameter?

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleMapOutputWriter.scala
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.spark.internal.{config, Logging}
+import org.apache.spark.shuffle.api.{ShuffleMapOutputWriter, ShufflePartitionWriter}
+import org.apache.spark.shuffle.api.metadata.MapOutputCommitMessage
+
+class RemoteHadoopShuffleMapOutputWriter(
+  shuffleId: Int,
+  mapId: Long,
+  numPartitions: Int,
+  resolver: RemoteShuffleBlockResolver,
+  conf: SparkConf) extends ShuffleMapOutputWriter with Logging {
```
carsonwang, line 34: need 4 spaces for the class parameters. also fix for other classes.

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleMapOutputWriter.scala
```diff
@@ -0,0 +1,124 @@
...
+  private val partitionLengths: Array[Long] = new Array[Long](numPartitions)
+  private val bufferSize = conf.get(config.SHUFFLE_UNSAFE_FILE_OUTPUT_BUFFER_SIZE).toInt * 1024
+
+  private val tempFile: Path = resolver.createTempShuffleBlock()._2
+  private var hadoopOutputStream: FSDataOutputStream = _
+  private var bufferedOutputStream: OutputStream = _
+
+  private def initStream(): Unit = {
+    if (hadoopOutputStream == null) {
+      hadoopOutputStream = fs.create(tempFile)
```
carsonwang, line 47: can we remove hadoopOutputStream and only keep bufferedOutputStream? can we init bufferedOutputStream when creating this RemoteHadoopShuffleMapOutputWriter?

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleMapOutputWriter.scala
gczsjdy: It seems the base class(`OutputStream`)'s write(byte[] b) is:
```
    public void write(byte b[]) throws IOException {
        write(b, 0, b.length);
    }
``` 
So we don't need to override that function.

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleMapOutputWriter.scala
gczsjdy: `bufferedOutputStream` is a field of the outside class, `PartitionWriterStream` is an internal class of the 'outside class', so it automatically has access to `bufferedOutputStream`

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleMapOutputWriter.scala
gczsjdy: Is that for less lines of code?
Actually I follow the implementations here: https://github.com/apache/spark/blob/b44acee95343f3508f12384b17e43cde65220924/core/src/main/java/org/apache/spark/shuffle/sort/io/LocalDiskShuffleMapOutputWriter.java#L139-L146.

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleOutputTracker.scala
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  override def registerMapOutput(
+    shuffleId: Int,
+    mapId: Int,
+    mapTaskAttemptId: Long,
+    mapOutputMetadata: MapOutputMetadata): Unit = {
+    val mapperToMetadata = shuffleStatuses.get(shuffleId)
+    mapperToMetadata.put(
+      mapTaskAttemptId, mapOutputMetadata.asInstanceOf[RemoteHadoopMapOutputMetadata])
```
carsonwang, line 50: Should be mapId instead of mapTaskAttempId, right? It seems map task attempts are not handled properly.

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleMetadata.scala
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package org.apache.spark.shuffle.sort.remote
+
+import java.util.concurrent.ConcurrentHashMap
+
+import org.apache.spark.shuffle.api.metadata.{MapOutputMetadata, ShuffleMetadata}
+import org.apache.spark.storage.BlockManagerId
+
+class RemoteHadoopShuffleMetadata(val map: ConcurrentHashMap[Long, RemoteHadoopMapOutputMetadata])
```
carsonwang, line 25: what about creating the map in the class instead of passing it by the creator?

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleOutputTracker.scala
```diff
@@ -0,0 +1,63 @@
...
+    shuffleStatuses.put(shuffleId,
+      new RemoteHadoopShuffleMetadata(new ConcurrentHashMap[Long, RemoteHadoopMapOutputMetadata]()))
+  }
+
+  /**
+    * Called when the shuffle with the given id is unregistered because it will no longer
+    * be used by Spark tasks.
+    */
+  override def unregisterShuffle(shuffleId: Int, blocking: Boolean): Unit = {
+    RemoteShuffleBlockResolver.active.removeDataByShuffleId(shuffleId)
```
carsonwang, line 40: should remove the shuffleId from shuffleStatuses

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleOutputTracker.scala
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+package org.apache.spark.shuffle.sort.remote
+
+import java.util.Optional
+import java.util.concurrent.ConcurrentHashMap
+
+import org.apache.spark.shuffle.api.metadata.{MapOutputMetadata, ShuffleMetadata, ShuffleOutputTracker}
+
+class RemoteHadoopShuffleOutputTracker extends ShuffleOutputTracker {
```
carsonwang, line 25: can you override all the methods? If there is no need to implement the method, just keep it. I think we should implement invalidateShuffle.

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleOutputTracker.scala
```diff
@@ -0,0 +1,63 @@
...
+  override def removeMapOutput(shuffleId: Int, mapId: Int, mapTaskAttemptId: Long): Unit = {}
+
+  override def getShuffleMetadata(shuffleId: Int): Optional[ShuffleMetadata] = {
+    Optional.of(shuffleStatuses.get(shuffleId))
+  }
+
+  override def isMapOutputAvailableExternally(
+    shuffleId: Int,
+    mapId: Int,
+    mapTaskAttemptId: Long): Boolean = true
```
carsonwang, line 62: need to check if the files do exist. If they are removed, then it should return false.

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteShuffleBlockResolver.scala
```diff
@@ -0,0 +1,459 @@
...
+    new RemoteShuffleTransferService(
+      conf,
+      env.securityManager,
+      env.blockManager.blockManagerId.host,
+      env.blockManager.blockManagerId.host,
+      env.conf.get(config.BLOCK_MANAGER_PORT),
+      conf.get(RemoteShuffleConf.NUM_TRANSFER_SERVICE_THREADS))
+  }
+  private[remote] lazy val shuffleServerId = {
+    remoteShuffleTransferService.asInstanceOf[RemoteShuffleTransferService].getShuffleServerId
```
carsonwang, line 78: If index cache is disabled, the remoteShuffleTransferService should never be used at all. Currently it will still be created.

oap-shuffle/remote-shuffle/LICENSE
```diff
@@ -0,0 +1,202 @@
+                                 Apache License
```
carsonwang, line 1: Do we still need this file as part of OAP?

oap-shuffle/remote-shuffle/scalastyle-config.xml
```diff
@@ -0,0 +1,386 @@
+<!--
```
carsonwang, line 1: Do we have scala style config at OAP level, then we don't need this file..

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleExecutorComponents.scala
```diff
@@ -0,0 +1,79 @@
...
+    shuffleMetadata: Optional[ShuffleMetadata]): lang.Iterable[ShuffleBlockInputStream] = {
+    val meta = shuffleMetadata.get().asInstanceOf[RemoteHadoopShuffleMetadata]
+    new lang.Iterable[ShuffleBlockInputStream] {
+      override def iterator(): util.Iterator[ShuffleBlockInputStream] =
+        new RemoteShuffleBlockIterator(
+          TaskContext.get(),
+          resolver.remoteShuffleTransferService,
+          resolver,
+          blockInfos.asScala.groupBy(shuffleBlockInfo =>
+            meta.get(shuffleBlockInfo.getMapTaskAttemptId).shuffleServerId)
```
carsonwang, line 59: MapTaskAttemptId or MapId?

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleExecutorComponents.scala
```diff
@@ -0,0 +1,79 @@
...
+            .map { case (blockManagerId, shuffleBlockInfos) =>
+              (blockManagerId,
+                shuffleBlockInfos.map(info =>
+                  (ShuffleBlockId(
+                    info.getShuffleId,
+                    info.getMapTaskAttemptId,
+                    info.getReduceId).asInstanceOf[BlockId],
+                    info.getBlockLength,
+                    info.getMapIndex)).toSeq)
+            }.iterator,
```
carsonwang, line 69: can you please move this part out, create a variable, to make it more readable?

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleExecutorComponents.scala
```diff
@@ -0,0 +1,79 @@
...
+            meta.get(shuffleBlockInfo.getMapTaskAttemptId).shuffleServerId)
+            .map { case (blockManagerId, shuffleBlockInfos) =>
+              (blockManagerId,
+                shuffleBlockInfos.map(info =>
+                  (ShuffleBlockId(
+                    info.getShuffleId,
+                    info.getMapTaskAttemptId,
+                    info.getReduceId).asInstanceOf[BlockId],
+                    info.getBlockLength,
+                    info.getMapIndex)).toSeq)
```
carsonwang, line 68: can you create a blockId variable to make this more readable? It looks like you are passing 5 variables to ShuffleBlockId()

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleMapOutputWriter.scala
carsonwang: I know. But it looks not quite clear.

carsonwang:
@gczsjdy , can you please test this implementation using repartition on DAOS and report the performance? Is this still the same to the previous shuffle manger plugin?

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleMapOutputWriter.scala
carsonwang: I see

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleMapOutputWriter.scala
gczsjdy: Introducing a class parameter adds one field for each object, also 2 lines... Actually don't understand why internal class is not clear...

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleOutputTracker.scala
gczsjdy: Good catch! Should be `mapId`. Seems we don't need to use map task attempts here?

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleMetadata.scala
gczsjdy: Makes sense!

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleOutputTracker.scala
gczsjdy: - I implemented all methods that don't have default implementations. I think the default implementations' design is for you to not overriding it if no needs, right?
- There is no invalidateShuffle in that class, `unregisterShuffle` was implemented

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleOutputTracker.scala
gczsjdy: Emm interesting catch. I think you're right

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteShuffleBlockIterator.scala
```diff
@@ -0,0 +1,546 @@
...
+    numBlocksProcessed += 1
+
+    var result: RemoteFetchResult = null
+    var input: InputStream = null
+    // Take the next fetched result and try to decompress it to detect data corruption,
+    // then fetch it one more time if it's corrupt, throw FailureFetchResult if the second fetch
+    // is also corrupt, so the previous stage could be retried.
+    // For local shuffle block, throw FailureFetchResult for the first IOException.
+    while (result == null) {
+      val startFetchWait = System.currentTimeMillis()
```
carsonwang, line 333: Please use nanoTime for accuracy

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteShuffleBlockIterator.scala
```diff
@@ -0,0 +1,546 @@
...
+              logError("Failed to create input stream from block backed by Hadoop file segment", e)
+              throwFetchFailedException(blockId, mapIndex, address, e)
+          }
+          var isStreamCopied: Boolean = false
+          // Detect ShuffleBlock corruption
+          try {
+            input = streamWrapper(blockId, in)
+            // Only copy the stream if it's wrapped by compression or encryption, also the size of
+            // block is small (the decompressed block is smaller than maxBytesInFlight)
+            if (detectCorrupt && !input.eq(in) && size < maxBytesInFlight / 3) {
```
carsonwang, line 381: this is the code from spark 2.4. There are much difference in 3.0

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/HadoopFileSegmentManagedBuffer.scala
```diff
@@ -0,0 +1,207 @@
...
+
+  override def size(): Long = length
+
+  override def nioByteBuffer(): ByteBuffer = ???
+
+  override def createInputStream(): InputStream = if (eagerRequirement) {
+    logInfo("Eagerly requiring this data input stream")
+    dataStream
+  } else {
+    new ByteArrayInputStream(dataInByteArray)
```
carsonwang, line 115: Eager means load the whole shuffle block in to memory before any further computation, right? `new ByteArrayInputStream(dataInByteArray)` seems to be eager execution here..

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/HadoopFileSegmentManagedBuffer.scala
```diff
@@ -0,0 +1,207 @@
...
+    }
+  }
+
+  override def hashCode(): Int = super.hashCode()
+
+  override def retain(): ManagedBuffer = this
+
+  override def release(): ManagedBuffer = this
+
+  override def convertToNetty(): AnyRef = nettyByteBuffer
```
carsonwang, line 133: throw exception if you have no implementation?

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/HadoopFileSegmentManagedBuffer.scala
```diff
@@ -0,0 +1,207 @@
...
+    if (! eagerRequirement) {
+      dataInByteArray
+    }
+  }
+
+  private lazy val nettyByteBuffer = ???
+
+  override def size(): Long = length
+
+  override def nioByteBuffer(): ByteBuffer = ???
```
carsonwang, line 109: Better to throw a meaningful exception instead of not implemented.

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteShuffleBlockResolver.scala
```diff
@@ -0,0 +1,459 @@
...
+    fs.delete(dir, true)
+    try {
+      HadoopFileSegmentManagedBuffer.handleCache.values().forEach {
+        new Consumer[mutable.HashMap[Path, FSDataInputStream]] {
+          override def accept(t: mutable.HashMap[Path, FSDataInputStream]): Unit = {
+            t.values.foreach(JavaUtils.closeQuietly)
+          }
+        }
+      }
+      JavaUtils.closeQuietly(remoteShuffleTransferService)    } catch {
```
carsonwang, line 396: move catch to the next line

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleOutputTracker.scala
gczsjdy: After some investigation, I doubt this.
I think this API was designed differently, instead it:
> Inspect whether or not all partitions from the given map output are available in an external
   system, i.e. offloaded from an executor.

and it's called

> If the executor that ran the map task is lost, all map outputs that were written by that
executor are checked against this method. For each map output that is not stored externally,
that map output has to be recomputed.

Imagine we store shuffle files in DFS(mostly replicated), turn on dynamic allocation, every time an executor is lost, we need to check all the files' existence(mostly it's existed), this is a waste of resource.

Here is where this `isMapOutputAvailableExternally` is called: https://github.com/gczsjdy/RemoteShuffle/blob/94bd8cedddbb66a7d0b022e7b9ce3ec8b1c9ae44/core/src/main/scala/org/apache/spark/scheduler/DAGScheduler.scala#L1881-L1885. And here `fileLost` is true.

gczsjdy: Also refer to https://github.com/Intel-bigdata/RemoteShuffle/blob/94bd8cedddbb66a7d0b022e7b9ce3ec8b1c9ae44/core/src/main/java/org/apache/spark/shuffle/sort/io/LocalDiskShuffleOutputTracker.java#L56-L59, the local-disk implementation also justifies the API doesn't involve checking every specific file's existence.

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteShuffleBlockResolver.scala
gczsjdy: Right!

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleExecutorComponents.scala
gczsjdy: Great advice!

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleExecutorComponents.scala
gczsjdy: Done!

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleExecutorComponents.scala
gczsjdy: mapId, modified.

oap-shuffle/remote-shuffle/LICENSE
gczsjdy: Removed.

oap-shuffle/remote-shuffle/scalastyle-config.xml
gczsjdy: Yep.

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleExecutorComponents.scala
gczsjdy: I still kept this style,because it's weird to create a variable in `map{}` and directly use it.

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteHadoopShuffleOutputTracker.scala
gczsjdy: Done.

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/HadoopFileSegmentManagedBuffer.scala
gczsjdy: - Well, both are correct understandings, eager here means eagerly get the `InputStream`, what you mentioned is another way of understanding. This is an internal configuration, users will not notice. Currently it always goes to line 115.
- Do you think branch around line-113('eager' in your meaning) can be faster for some storage?

oap-shuffle/remote-shuffle/scalastyle-config.xml
gczsjdy: Seems we don't have style configs on OAP level, so this remains a problem. I will create an issue to ask other OAP guys.

oap-shuffle/remote-shuffle/src/main/scala/org/apache/spark/shuffle/sort/remote/RemoteShuffleBlockIterator.scala
gczsjdy: Good ponit! Let me port important changes here.
I saw ~20 commits between Spark 2.4.0-present, significant commits including consecutive fetch(for Adaptive Execution), features like this are ones we definitely need.

Interestingly, the new shuffle I/O API still can't(mainly on read path) free developers from porting to newer upstream Spark.

gczsjdy: I have roughly gone through the upstream changes during the period, basically 3 features/improvements:
- AE-related consecutive fetch, there are a ton...
- Big block corruption detect
- Iterator fast release

gczsjdy:
@carsonwang Thanks for the review. Is it better to test after the 3 improvements(mentioned @ https://github.com/Intel-bigdata/OAP/pull/1315#discussion_r431175794) are ported to this PR?

gczsjdy:
Closing this, the decision is to not using Intel-specific Spark...

PR #1332, opened 2020-05-28
[oap-native-sql][Scala] adding round support

rui-mo:


mmoriaaa:
success running on Jenkins.

zhouyuan:
@rui-mo can you do a rebase on this?

rui-mo:
@zhouyuan rebased now, and this patch has some issue on decimal precision, may causing incorrect result.  Do we need to merge it?

zhouyuan:
@rui-mo I saw there's a TODO() we could fix this when decimal support is ready

PR #1333, opened 2020-05-28
[oap-native-sql][Scala] bypass nullPointerException

rui-mo:


mmoriaaa:
success running on Jenkins.

PR #1341, opened 2020-05-29
[oap-native-sql] Enable mvn package to build native libs

JiayiChen785:

Added the generate resource phase in spark columnar core to execute compile.sh and build the native library.

## How was this patch tested?

Manually tested.

### DIFF
--- a/oap-native-sql/cpp/compile.sh
+++ b/oap-native-sql/cpp/compile.sh
@@ -1,0 +1,18 @@
+#! /bin/bash
+
+CURRENT_DIR=$(cd "$(dirname "$BASH_SOURCE")"; pwd)
+echo $CURRENT_DIR
+
+cd ${CURRENT_DIR}
+if [ -d build ]; then
+    rm -r build
+fi
+mkdir build
+cd build
+cmake ..
+make -j
+make -j
+make install
+
+# cd $CURRENT_DIR
+# rm -r build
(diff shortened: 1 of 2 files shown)


oap-native-sql/cpp/compile.sh
```diff
@@ -0,0 +1,18 @@
...
+
+cd ${CURRENT_DIR}
+if [ -d build ]; then
+    rm -r build
+fi
+mkdir build
+cd build
+cmake ..
+make -j
+make -j
```
zhouyuan, line 14: looks like "make -j" will make trouble for travis tests

oap-native-sql/cpp/compile.sh
```diff
@@ -10,8 +10,8 @@ fi
 mkdir build
 cd build
 cmake ..
-make -j
-make -j
+make
+make
 make install
```
zhouyuan, line 17: should remove "make install"

JiayiChen785: Removing this line passes the test, but will it cause problems as the script will no longer update lib files under system dir, such as libspark_columnar_jni.so

zhouyuan: It should be ok as this lib will be packaged in the jar file, and is there any special objective by "make" twice?

JiayiChen785: I cleaned the compile.sh and removed one "make". When running the compile bash script alone, it needs to "make" twice as the first failed after building protobuf. After added as a mvn stage, running "make" once is enough.

zhouyuan:
@JiayiChen785 thanks for improving this!

PR #1358, opened 2020-06-09
[oap-native-sql][Scala] support PartialMerge mode for aggregate

rui-mo:
This patch was used to get aggregate result attrs from buffer attrs, so as to solve the avg attr missing error.

### DIFF
--- a/oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/expression/ColumnarProjection.scala
+++ b/oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/expression/ColumnarProjection.scala
@@ -73,10 +73,7 @@ class ColumnarProjection (
     case (expr, i) => {
       ColumnarExpressionConverter.reset()
       var columnarExpr: Expression =
-        ColumnarExpressionConverter.replaceWithColumnarExpression(expr, originalInputAttributes)
-      if (columnarExpr.isInstanceOf[AttributeReference]) {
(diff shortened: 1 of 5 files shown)


rui-mo:
successful running on Jenkins

oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/expression/ColumnarProjection.scala
```diff
@@ -87,7 +84,7 @@ class ColumnarProjection (
     }
   }
 
-  val (ordinalList, arrowSchema) = if (projPrepareList.size > 0 && check_if_no_calculation == false) {
```
rui-mo, line 90: to calculate for Alias and Attribute and avoid missing col.

rui-mo:
@xuechendi , could you help review this patch, thanks.

xuechendi:
Hi, @rui-mo , thanks for the great work, I just read the codes, and it makes sense to me, so the major change is to add the support for partial-merge which will use our expanded aggregateResultAttributes 
 , right? And once you checked with tpch queries and make sure the correctness of all queries? We can merge it.

rui-mo:
The correctness of all tpch queries were verified on jenkins, and no other issue was found in tpc-ds queries. We could merge it, thanks. @xuechendi @zhouyuan

zhouyuan:
@rui-mo thanks for improving aggregate!

PR #1367, opened 2020-06-12
[POAE7-335] libpmemblk backend and libpmemkv metastore implementation

yeyuqiang:


oap-common/src/native/libpmemblk/com_intel_oap_common_unsafe_PMemBlockPlatform.cpp
```diff
@@ -0,0 +1,136 @@
...
+ * Method:    read
+ * Signature: (Ljava/nio/ByteBuffer;I)V
+ */
+JNIEXPORT void JNICALL Java_com_intel_oap_common_unsafe_PMemBlockPlatform_read
+  (JNIEnv *env, jclass clazz, jbyteArray jbuf, jint jindex) {
+
+  jbyte* buf = env->GetByteArrayElements(jbuf, 0);
+  int index = (int) jindex;
+
+  if(pmemblk_read(pbp, buf, index) < 0) {
```
jikunshang, line 87: This API needs an extra copy like vmemcache, we can't use a libpmemblk address, right? If yes, there may be some overhead.

oap-common/src/test/java/com/intel/oap/common/unsafe/PMemBlockPlatformTest.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    @AfterClass
+    public static void close() {
+        PMemBlockPlatform.close();
+    }
+
+    private byte[] writeBlock(int index) {
+        byte[] bytesToWrite = new byte[ELEMENT_SIZE];
```
jikunshang, line 43: Do we support multi thread read/write?

oap-common/src/main/java/com/intel/oap/common/storage/pmemblk/PMemBlkChunkReader.java
```diff
@@ -0,0 +1,33 @@
...
+        this.chunkSize = pMemManager.getChunkSize();
+    }
+
+    @Override
+    protected int readFromPMem(PMemPhysicalAddress pMemPhysicalAddress, ByteBuffer data) {
+        PMemBlkPhysicalAddress pMemBlkPhysicalAddress = (PMemBlkPhysicalAddress) pMemPhysicalAddress;
+        int index = pMemBlkPhysicalAddress.getIndex();
+        byte[] buffer = new byte[chunkSize];
+        PMemBlockPlatform.read(buffer, index);
+        data.put(buffer);
```
yma11, line 25: Why should here always read whole data of this index? What will happen if chunkSize>data.size()?

oap-common/src/main/java/com/intel/oap/common/storage/pmemblk/PMemBlkPhysicalAddress.java
```diff
@@ -0,0 +1,17 @@
+package com.intel.oap.common.storage.pmemblk;
+
+import com.intel.oap.common.storage.stream.PMemPhysicalAddress;
+
+public class PMemBlkPhysicalAddress implements PMemPhysicalAddress {
+
+    private int index;
```
yma11, line 7: It makes sense we only track index of this pmemblk when it's not the last trunk of the stream. But for the last chunk, don't we need to track the actual size used of this pmemblk?

oap-common/src/test/java/com/intel/oap/common/unsafe/PMemBlockPlatformTest.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Test
+    public void testWriteMultipleBlock() {
+        for (int i = 0; i < 32; i++) {
+            byte[] writtenBlock = writeBlock(i);
+            byte[] readBlock = readBlock(i);
+            assertArrayEquals(writtenBlock, readBlock);
+        }
+    }
```
yma11, line 69: I think you only test the boundary cases that data size is X*chunkSize,  we need additional one to cover X*chunkSize + Y case, right?

oap-common/src/native/libpmemblk/com_intel_oap_common_unsafe_PMemBlockPlatform.cpp
yeyuqiang: Should be no extra copy. pmemblk_read will handle the copy from address(determine by index) to arrays of block (buf)

oap-common/src/test/java/com/intel/oap/common/unsafe/PMemBlockPlatformTest.java
yeyuqiang: Yes

oap-common/src/main/java/com/intel/oap/common/storage/pmemblk/PMemBlkChunkReader.java
yeyuqiang: pmemblk will always put a buf with chunkSize otherwise it will fail.

oap-common/src/test/java/com/intel/oap/common/unsafe/PMemBlockPlatformTest.java
yeyuqiang: For libpmemblk, the case should be (x+y)chunksize. Then testWriteBlockExceedMaximum is the expected test case

oap-common/src/main/java/com/intel/oap/common/storage/pmemblk/PMemBlkPhysicalAddress.java
yeyuqiang: I think metadata mechanism has already handle this problem.

yma11: Metafooter only record total trunks of this stream but not total size. There is no method to get actual size of last trunk.

yeyuqiang: memory stats is tracking on total size and usage. I should return the buffer size in readFromPMem for the last trun is < chunksize

yeyuqiang: and record the real chunk length in PhysicalAddress. Does it make sense?

yma11: I think so. I did that in memkind. and also you may need to rethink about the write/read part, that you may read a whole trunk but can't just put all of them into returned buffer?

yeyuqiang:
any other further comments? If not, I will merge it and open new PR for integration

PR #1370, opened 2020-06-13
[oap-native-sql]Add ColumnarBatch Combination on Shuffle Read Side

rongma1997:

Add a new operator CoalesceBatch to support ColumnarBatch combination on shuffle read side.

## How was this patch tested?

Existing unit tests. Manually test and pass all tpch queries.

### DIFF
--- a/oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/execution/CoalesceBatchesExec.scala
+++ b/oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/execution/CoalesceBatchesExec.scala
@@ -1,12 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


zhixingheyi-tian:
@rongma1997 
Would you please add the module name  in the PR title?

Thanks

oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/execution/CoalesceBatchesExec.scala
```diff
@@ -1,23 +1,152 @@
...
+
+    val recordsPerBatch = conf.arrowMaxRecordsPerBatch
+    val numInputRows = longMetric("numInputRows")
+    val numInputBatches = longMetric("numInputBatches")
+    val numOutputBatches = longMetric("numOutputBatches")
+    val concatTime = longMetric("concatTime")
+    val collectTime = longMetric("collectTime")
+    val avgCoalescedNumRows = longMetric("avgCoalescedNumRows")
+
+    child.executeColumnar().mapPartitions { iter =>
```
zhouyuan, line 65: I tried to increase this value to 20k, but from the statics of the operator, it's able to get 5k rows/batch. Is this the expected behavior?

rongma1997: which value?

zhouyuan: (clicked on the wrong line) 
I mean for this line L57
`val recordsPerBatch = conf.arrowMaxRecordsPerBatch`

rongma1997: Yes. For each reducer task, the actual combined size can be smaller than the configuration if there are no more data to get.

PR #1372, opened 2020-06-16
[oap-native-sql]Wip optimize sort

xuechendi:
If columnar sort operator is sort to a no payload single column, we will use inplace sort by ska_sort, which will boost performance by 3x in micro benchmark


xuechendi:
This PR is complete, I noticed that quite some queries will fall to single column no payload case, @zhouyuan

PR #1376, opened 2020-06-17
[DataSource][Arrow] Support reading dictionary encoded parquet values

zhztheplayer:
see also Intel-bigdata/arrow#73


zhztheplayer:
Hi @xuechendi @zhouyuan Could you please review API changes to writable vector to make sure we'll be able to access dictionary values conveniently from columnar plugin's code? Thanks.

PR #1391, opened 2020-06-22
Modify pom.xml to add RPMem shuffle as a sub project

Eugene-Mark:
## Add RPMem shuffle as a sub project in OAP's pom

The RPMem is added as a module. 
For now, the RPMem shuffle can be packaged specifically with:

```mvn package -pl com.intel.oap:oap-rpmem-shuffle -amd package -DskipTests```

### DIFF
--- a/oap-shuffle/RPMem-shuffle/core/pom.xml
+++ b/oap-shuffle/RPMem-shuffle/core/pom.xml
@@ -5,14 +5,13 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
-      <groupId>com.intel.spark-pmof</groupId>
-      <artifactId>spark-pmof</artifactId>
-      <version>1.0</version>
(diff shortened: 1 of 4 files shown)


Jian-Zhang:
Does this mean we need to update the docs about how to build RPMem-shuffle? 
instead of doing it under oap-shuffle/RPMem-shuffle folder, need to issue this command in oap folder?

Eugene-Mark:
> Does this mean we need to update the docs about how to build RPMem-shuffle?
> instead of doing it under oap-shuffle/RPMem-shuffle folder, need to issue this command in oap folder?

Either way is workable, for user who wants to use RPMem shuffle, she can build it at ${oap-root-folder} or ${oap-root-folder}/opa-shuffle/RPMem-shuffle, and if in the latter folder, the build command is exactly the same as before.

Jian-Zhang:
LGTM.

PR #1394, opened 2020-06-23
[oap-native-sql] compression use single thread

rongma1997:

Arrow's RecordBatch compression by default use multi-threads to compress. Need to disable multi-thread compression in our use case due to performance downgrade.

The decompression doesn't call the RecordBatch API and use single thread, so it doesn't need modification.

After disabling multi-thread compression, the native shuffle split + write micro benchmark improve ~60% (1330ms vs 825ms, input file 132M)

## How was this patch tested?

Unit tests. Pass TPCH queries.


rongma1997:
@zhouyuan @xuechendi

PR #1363, opened 2020-06-12
[Building] allow to build native sql from OAP parent folder

zhouyuan:
Signed-off-by: Yuan Zhou <yuan.zhou@intel.com>

## What changes were proposed in this pull request?

refine  pom to allow building from oap-parent folder

PR #1400, opened 2020-06-29
[OAP-1458][oap-native-sql] adding ExistenceJoin support

rui-mo:

### DIFF
--- a/oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarShuffledHashJoin.scala
+++ b/oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarShuffledHashJoin.scala
@@ -139,6 +139,8 @@ class ColumnarShuffledHashJoin(
 
   }
 
+  var existsField : Field = null
+  var existsIndex : Int = -1
   val (probe_func_name, build_output_field_list, stream_output_field_list) = joinType match {
     case _: InnerLike =>
       ("conditionedProbeArraysInner", build_input_field_list, stream_input_field_list)
@@ -150,6 +152,15 @@ class ColumnarShuffledHashJoin(
       ("conditionedProbeArraysOuter", build_input_field_list, stream_input_field_list)
     case LeftAnti =>
       ("conditionedProbeArraysAnti", List[Field](), stream_input_field_list)
+    case j : ExistenceJoin =>
+      val existsSchema = j.exists
+      existsField = Field.nullable(s"${existsSchema.name}",
+        CodeGeneration.getResultType(existsSchema.dataType))
+      // Use the last index for we cannot distinguish between two "exists" StructType
+      existsIndex = resultFieldList.lastIndexOf(existsField)
+      existsField = Field.nullable(s"${existsSchema.name}#${existsSchema.exprId.id}",
+        CodeGeneration.getResultType(existsSchema.dataType))
+      ("conditionedProbeArraysExistence", build_input_field_list, stream_input_field_list)
     case _ =>
       throw new UnsupportedOperationException(s"Join Type ${joinType} is not supported yet.")
   }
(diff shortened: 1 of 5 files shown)


rui-mo:
Verified tpch queries on Jenkins. Chendi, could you help review this patch? thanks. @xuechendi

zhouyuan:
the CI failure is due to some API change in upstream spark, it's fixed in #1363

oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/expression/ColumnarShuffledHashJoin.scala
```diff
@@ -182,7 +189,12 @@ class ColumnarShuffledHashJoin(
     case BuildLeft =>
       (build_input_field_list, build_output_field_list ::: stream_output_field_list)
     case BuildRight =>
-      (build_input_field_list, stream_output_field_list ::: build_output_field_list)
+      joinType match {
+        case ExistenceJoin(_) =>
+          (build_input_field_list, stream_output_field_list :+ existsField)
```
xuechendi, line 194: will build right also do existence join?

oap-native-sql/cpp/src/codegen/arrow_compute/ext/probe_kernel.cc
```diff
@@ -162,6 +163,10 @@ class ConditionedProbeArraysKernel::Impl {
         }
         i++;
       }
+      // ExistenceJoin
+      if (join_type == 4) {
+        (*result_schema_index_list).push_back(std::make_pair(1, i));
```
xuechendi, line 168: (1, i) means existence col will be showed as the last column in stream side, it seems to be quite fixed to me, a better way is to use spark result schema to decide the position of existence join column.

oap-native-sql/cpp/src/codegen/arrow_compute/ext/probe_kernel.cc
```diff
@@ -622,6 +627,59 @@ class ConditionedProbeArraysKernel::Impl {
...
+          index = hash_table_->GetNull();
+        }
+        if (index == -1) {
+          )" +
+           right_valid_ss.str() + right_not_exist_ss.str() + R"(
+          out_length += 1;
+        } else {
+          for (auto tmp : (*memo_index_to_arrayid_)[index]) {
+            )" +
+           shuffle_str + R"(
```
xuechendi, line 678: I don't think you need to add multiple rows to mark distinct left key is exists . Can you verify on this? I think existence row should only be one row for one key.

zhouyuan:
@rui-mo  needs rebase on the new code

oap-native-sql/cpp/src/codegen/arrow_compute/ext/probe_kernel.cc
rui-mo: yes, only distinct exists was needed

oap-native-sql/core/src/main/scala/com/intel/sparkColumnarPlugin/expression/ColumnarShuffledHashJoin.scala
rui-mo: in SparkStrategies.scala:
168     private def canBuildRight(joinType: JoinType): Boolean = joinType match {
169       case _: InnerLike | LeftOuter | LeftSemi | LeftAnti | _: ExistenceJoin => true
170       case _ => false
171     }
173     private def canBuildLeft(joinType: JoinType): Boolean = joinType match {
174       case _: InnerLike | RightOuter => true
175       case _ => false
176     }

oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarShuffledHashJoin.scala
```diff
@@ -150,6 +152,15 @@ class ColumnarShuffledHashJoin(
       ("conditionedProbeArraysOuter", build_input_field_list, stream_input_field_list)
     case LeftAnti =>
       ("conditionedProbeArraysAnti", List[Field](), stream_input_field_list)
+    case j : ExistenceJoin =>
+      val existsSchema = j.exists
+      existsField = Field.nullable(s"${existsSchema.name}",
+        CodeGeneration.getResultType(existsSchema.dataType))
+      // Use the last index for we cannot distinguish between two "exists" StructType
+      existsIndex = resultFieldList.lastIndexOf(existsField)
```
rui-mo, line 160: like this: resultSchema is StructType(StructField(c_customer_sk,IntegerType,true), StructField(c_current_cdemo_sk,IntegerType,true), StructField(c_current_addr_sk,IntegerType,true), StructField(exists,BooleanType,false), StructField(exists,BooleanType,false))

xuechendi:
@rui-mo, this PR Looks Good To Me

rui-mo:
we add ExistenceJoin support #https://github.com/Intel-bigdata/OAP/issues/1458 in ColumnarShuffledHashJoin, supporting cases that exists col locating in the middle of result_schema.

PR #1414, opened 2020-07-02
update pmemkv to stable 1.0 version

yeyuqiang:


oap-common/src/main/java/com/intel/oap/common/storage/pmemblk/PMemKVDatabase.java
```diff
@@ -29,3 +32,16 @@ private static boolean isCreated(String path) {
...
+class StringConverter implements Converter<String> {
+    public ByteBuffer toByteBuffer(String entry) {
+        return ByteBuffer.wrap(entry.getBytes());
+    }
+
+    public String fromByteBuffer(ByteBuffer entry) {
+        byte[] bytes;
+        bytes = new byte[entry.capacity()];
+        entry.get(bytes);
+        return new String(bytes);
```
xieqi, line 45: maybe use  new String(entry.array()) is more efficient.

yeyuqiang: This method just copy from pmemkv-java's example to adapt the pmemkv DataBase builder usage.  Updated

xieqi: Oh, maybe they think array() is not always available, we should call  hasArray() to check, if has we can simply this convert.

yeyuqiang: update

oap-cache/oap/pom.xml
```diff
@@ -117,6 +117,12 @@
       <groupId>com.intel</groupId>
       <artifactId>oap-common</artifactId>
       <version>${project.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>io.pmem</groupId>
+          <artifactId>libpmemkv-jni</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
```
yma11, line 126: is this exclusion done here because PMemKV only used in oap-common module and won't need access from higher level?

yeyuqiang: We dont need this artifact and maven assembly plugin does not support include artifact which type is so (libpmemkv-jni's type is so instead of jar usually)

yma11: Get it. LGTM.

yeyuqiang:
Merged. You can rebase your code now. @yma11

winningsix:
Do we need to download libpmemkv first in dev machine? It's not supported in OSX.

yeyuqiang:
> Do we need to download libpmemkv first in dev machine? It's not supported in OSX.

Yes, we have to install libpmemkv-java first or package it into. Looks like OSX is not proper to be dev machine here.

PR #1416, opened 2020-07-02
[ArrowDataSource][Scala] allow to specify batch size from Spark

zhouyuan:


zhztheplayer:
+1, This looks fine to me. Let's just make it configurable.

BTW did we have made decision about env var on this? I don't clearly remember it.

zhouyuan:
@zhztheplayer thanks! yes, the native code is pass `spark.sql.execution.arrow.maxRecordsPerBatch` to c++ code via registering env variable

PR #1428, opened 2020-07-03
Modify the artifact name with aligned version number and modify the e…

Eugene-Mark:
Modify the artifact name to be aligned with parent OAP project which is 0.8.1 and modify the enabling guide as per.

### DIFF
--- a/oap-shuffle/RPMem-shuffle/README.md
+++ b/oap-shuffle/RPMem-shuffle/README.md
@@ -724,8 +724,8 @@ spark.driver.memory    10g
 spark.yarn.driver.memoryOverhead 5g
 
 spark.io.compression.codec                                  snappy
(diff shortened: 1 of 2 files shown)


Jian-Zhang:
LGTM.

Eugene-Mark:
@Jian-Zhang Just updated commit, please try build again. I tested with version 0.8 before and didn't find issue. After short sync with test team we reach the agreement to change the version to 0.8.1, and I just updated the pom.xml without further verification. Howeveer, the issue happens because the combination between artifact oap-rpmem-shuffle and version 0.8.1 is duplicated both in rpmem shuffle parent project and java sub-project. I modified the sub-project's artifact ID with a "-java" suffix to distinguish it from parent project. Now, it should work find. Thanks.

Jian-Zhang:
LGTM.

PR #1434, opened 2020-07-03
[oap-native-sql] Some fixes for codegen groupby aggr

xuechendi:
1. use arrow::internal::HashTraits<arrow::StringType>::MemoTableType instead of ScalarMemoTable
2. update metrics for groupby hashaggr


xuechendi:
Hi, @zhouyuan , this is a fix for TPCH Q21, I realized that I used ScalarMemoTable<std::string> rather than BinaryMemoTable, which brought some incorrect hashmap result, fixed in this PR

zhouyuan:
Cool, let me see if we can create another PR to master, thanks!

Thanks, -yuan
________________________________

PR #1442, opened 2020-07-07
[OAP-1441] [POAE7-349] change com.intel.spark to org.apache.spark in oap-spark

yeyuqiang:


zhixingheyi-tian:
Parent pom.xml also need modify spark.internal.version

yeyuqiang:
> Parent pom.xml also need modify spark.internal.version

I've already remove it in this PR. Currently this property occurs across the modules. We can unified it in parent pom.xml later.

PR #1444, opened 2020-07-08
[OAP-MLLIB] support to use batch iterator to set data from jvm to native

JkSelf:
When convert the RDD[Vector] to RDD[HomogenNumericTable], the native set method will be called for each value. This PR aims to use the batch iterator and batch size to control the calls number of the native set method.

## How was this patch tested?

Manual test the kmeans-hibench and compared the results


JkSelf:
will re-open this pr based on master branch and close this one.

PR #1446, opened 2020-07-08
[OAP-1461][oap-data-source][arrow] Overwrite ParquetFileFormat's reader with Arrow b…

zhztheplayer:
…ased reader

Also, refactor directory tree and add module spark-arrow-datasource-parquet


zhztheplayer:
One typical case of this feature is to make ArrowDataSource work together with Delta Lake. Some PoC codes: https://github.com/zhztheplayer/delta/commit/81249fce7f45f44783c991ca577dd04b1d5b34ab

Issue #1462, opened 2020-07-10
[oap-native-sql] Adaptive query execution columnar support


rongma1997:
1. Modify some spark's class, so that columnar rules can be applied on columnar exchange operator when enabling AQE.
2. Modify columnar rules and add some relative columnar operators in native SQL for AQE.

PR #1463, opened 2020-07-10
[OAP-1462][oap-native-sql]Adaptive query execution columnar support

rongma1997:
1. Modify some spark's class, so that columnar rules can be applied on columnar exchange operator when enabling AQE.
2. Modify columnar rules and add some relative columnar operators in Native SQL for AQE.

## How was this patch tested?
UT. Pass all TPCH queries.

**CoalesceShufflePartitions & OptimzeLocalShuffleReader example (TPCH Q20)**
image

image

**OptimizeSkewedJoin example**
image


rongma1997:
Changes to spark are now included in this PR.
Also created https://github.com/Intel-bigdata/spark/pull/35 for easier review.

oap-native-sql/core/src/main/scala/com/intel/oap/ColumnarPlugin.scala
```diff
@@ -114,10 +136,23 @@ case class ColumnarPreOverrides(conf: SparkConf) extends Rule[SparkPlan] {
...
+      val exchange = ColumnarShuffleExchangeExec.exchanges.get(s) match {
+        case Some(e) => e
+        case None => throw new IllegalStateException("Reused exchange operator not found.")
+      }
+      ReusedExchangeExec(id, exchange)
+    // We need to discard c2r since ColumnarShuffleExchangeExec can't be the last operator with AQE enabled
+    case ColumnarToRowExec(child: SparkPlan)
+        if SQLConf.get.adaptiveExecutionEnabled && columnarConf.enableColumnarShuffle && child
+          .isInstanceOf[ColumnarShuffleExchangeExec] =>
+      replaceWithColumnarPlan(child)
```
zhouyuan, line 155: not sure if i understand this correctly, maybe we could use the row based shuffle in this case?

rongma1997: It's not a case to fall back to row based shuffle. It's to avoid extra c2r and r2c. 
For example, the correct plan should be like:

```+- *(3) ColumnarToRow
   +- ColumnarSort
      +- CoalesceBatches
         +- ColumnarCustomShuffleReader coalesced
            +- ShuffleQueryStage
               +- ColumnarExchange rangepartitioning
```
We know that there will always be a c2r after the last columnar plan, and in columnar AQE the last plan in each query stage is ColumnarExchange. So without discarding this c2r, the plan can be like:
```+- *(3) ColumnarToRow
   +- ColumnarSort
      +- CoalesceBatches
         +- ColumnarCustomShuffleReader coalesced
            +- RowToColumnar
               +- ShuffleQueryStage
                  +- ColumnarToRow
                     +- ColumnarExchange rangepartitioning
```

rongma1997: The code comments are misleading. I'll rewrite them.

zhouyuan:
locally verified, the perf gain on q20 is very promising.

rongma1997:
Closes #1462

PR #1465, opened 2020-07-13
[OAP-1460][rpmem-shuffle]Bug fix to support decision support query

Eugene-Mark:
The shuffle behavior of Spark3.0 has some changes due to https://issues.apache.org/jira/browse/SPARK-25341, it's used to better support rolling back a shuffle map stage and recover environment. 
Some interfaces of RPMem shuffle plugin needs to be changed correspondingly.

This PR has been tested by 500GB decision support benchmark


Jian-Zhang:
Looks good.

PR #1476, opened 2020-07-15
[oap-native-sql]Wip accelarate codegen compilation

xuechendi:
This PR is aim to reduce codegen compilation time:
1. reduce native compilation time by pre-compile/instantiate template classes
2. do prebuild in spark driver and distribute newly generated jars to executors


zhouyuan:
replaced by #1490

Issue #1464, opened 2020-07-13
[SQL Index and Data Source Cache] Remove ORC `spark.sql.orc.copyBatchToSpark` related codes to support Spark-3.0.0


HongW2019:
Spark-3.0.0[SPARK-26584] has removed the configuration `spark.sql.orc.copyBatchToSpark` and related ORC codes including tests and benchmarks. So OAP SQL Index and Data Source Cache needs made corresponding change to support Spark-3.0.0.

zhixingheyi-tian:
@HongW2019 , write the  write link here for this issue

PR #1481, opened 2020-07-16
[OAP-1464][SQL-Data-Source-Cache]Disable OAP Orc Column VecotorCache

HongW2019:
Spark-3.0 has removed internal ORC configuration "spark.sql.orc.copyBatchToSpark" to simplify the code path, so it won't copy the ORC columnar batch to Spark columnar bath in the vectorized ORC batch. Then OAP for now does not support ORC colunmnVectorCache, so made corresponding change and remove related tests.

## How was this patch tested?

Passed unit tests.


HongW2019:
Fix issue #1464

PR #1490, opened 2020-07-17
[oap-native-sql]Accelerate codegen compilation

xuechendi:

cpp changes: remove some of arrow headers in codegened cpp files, and use our precompiled cpp wrapper header instead, which brought compile time 2x faster.
scala side: add prebuild call in driver operator construction, so codegen will now happened in driver side and then gened dll will be made as jar and upload to a unified spark jar list. Executor then will use signature to get precompiled codegen dll for further work.

## How was this patch tested?

Tested with unittest and spark running TPCH


xuechendi:
@rui-mo , rui, do you think you can run a quick test see if existingJoin still works in this PR? I codes did a refactor to current codegen join on your codes, but I am not sure if it still works.

rui-mo:
hi, the test shows the below error about "arrow::BooleanArray".

/tmp/tmp/spark-columnar-plugin-codegen-38241a97deca9816.cc:190:17: error: ‘BooleanArray’ was not declared in this scope
 std::shared_ptr<BooleanArray> cached_1_exists_;
                 ^~~~~~~~~~~~
/tmp/tmp/spark-columnar-plugin-codegen-38241a97deca9816.cc:190:17: note: suggested alternative:
In file included from /usr/local/include/arrow/memory_pool.h:26:0,
                 from /usr/local/include/arrow/compute/context.h:23,
                 from /tmp/tmp/spark-columnar-plugin-codegen-38241a97deca9816.cc:2:
/usr/local/include/arrow/type_fwd.h:82:7: note:   ‘arrow::BooleanArray’
 class BooleanArray;
       ^~~~~~~~~~~~
/tmp/tmp/spark-columnar-plugin-codegen-38241a97deca9816.cc:190:29: error: template argument 1 is invalid
 std::shared_ptr<BooleanArray> cached_1_exists_;
                             ^
/tmp/tmp/spark-columnar-plugin-codegen-38241a97deca9816.cc:191:28: error: ‘BooleanArray’ does not name a type
 using ArrayType_1_exists = BooleanArray;

xuechendi:
Oh, I see, now we used a self-wrapped Array not arrow::Array, I think I forgot to wrap BooleanArray, will fix it later, thanks!

xuechendi:
@rui-mo , I added BooleanArray in precompile, can you do a quick test again?

rui-mo:
@xuechendi passed q10 and q35 with existenceJoin now

PR #1494, opened 2020-07-19
[OAP-1492]Change spark 3.0.0 from preview version to formal release version

Eugene-Mark:
Change the downloaded Spark binary from preview version to formal version.


Jian-Zhang:
LGTM.

Jian-Zhang:
Although we did not change Spark, let's keep it as it now - users can refer this as how RPMem can be build and installed.

PR #1495, opened 2020-07-20
[OAP-MLLIB] support to use batch iterator to set data from jvm to native

JkSelf:
When convert the RDD[Vector] to RDD[HomogenNumericTable], the native set method will be called for each value. This PR aims to use the batch iterator and batch size to control the calls number of the native set method.

## How was this patch tested?
Manual test the kmeans-hibench and compared the results

### DIFF
--- a/oap-mllib/mllib-dal/src/main/scala/org/apache/spark/ml/clustering/KMeansDALImpl.scala
+++ b/oap-mllib/mllib-dal/src/main/scala/org/apache/spark/ml/clustering/KMeansDALImpl.scala
@@ -61,15 +61,15 @@ class KMeansDALImpl (
     LibLoader.loadLibrary()
 
     println(s"KMeansDALImpl: Start data conversion")
+    import scala.collection.JavaConverters._
 
     val start = System.nanoTime
-    it.zipWithIndex.foreach {
-      case (v, rowIndex) =>
-        for (colIndex <- 0 until numCols)
-          // TODO: Add matrix.set API in DAL to replace this
-          // matrix.set(rowIndex, colIndex, row.getString(colIndex).toDouble)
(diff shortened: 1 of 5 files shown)


JkSelf:
@xwu99 @carsonwang

oap-mllib/mllib-dal/src/main/scala/org/apache/spark/ml/clustering/KMeansDALImpl.scala
```diff
@@ -61,15 +61,15 @@ class KMeansDALImpl (
...
     val start = System.nanoTime
-    it.zipWithIndex.foreach {
-      case (v, rowIndex) =>
-        for (colIndex <- 0 until numCols)
-          // TODO: Add matrix.set API in DAL to replace this
-          // matrix.set(rowIndex, colIndex, row.getString(colIndex).toDouble)
-          OneDAL.setNumericTableValue(localData.getCNumericTable, rowIndex, colIndex, v(colIndex))
+    val iter = it.map{
+                curVector => curVector.toArray
     }
```
xwu99, line 69: Does toArray result in data copy ?

oap-mllib/mllib-dal/src/main/scala/org/apache/spark/ml/clustering/KMeansDALImpl.scala
```diff
@@ -61,15 +61,15 @@ class KMeansDALImpl (
...
+    val batchSize = 8 << 10;
+    val batchIter = new DataBatch.BatchIterator(iter.asJava, batchSize)
+    OneDAL.cSetDoubleIterator(localData.getCNumericTable, batchIter)
```
xwu99, line 73: I think you don't need an extra BatchIterator class, you can use JNI to access the data of original iterator inside the C function.

PR #1504, opened 2020-07-21
[OAP-1508][oap-native-sql] Clean java tmp dir when doing compilation

### LINKED ISSUE #1508: [oap-native-sql] Clean lib and headers with codegen
After using runtime codegen method, we need to use original name libspark_columnar_jni.so instead of a random name, which will lead to some library conflict in future.

This issue to aim to fix this problem:
1. use a random folder name for each jni context, so any runtime copies lib, header and codegened codes will locate inside a random name folder.
2. when temp dir path is specified, we will use anything left in this specified folder for performance optimization and debug purpose.

xuechendi:

This PR is aim to clean java tmp dir in compilation, since we now do runtime codegen compile in driver side, and driver uses /tmp dir and won't clean this folder after test, we will try clean this DIR at next time codes compilation.

fixes #1508

### DIFF
--- a/oap-native-sql/core/src/main/java/com/intel/oap/vectorized/ExpressionEvaluator.java
+++ b/oap-native-sql/core/src/main/java/com/intel/oap/vectorized/ExpressionEvaluator.java
@@ -48,8 +48,9 @@ public class ExpressionEvaluator implements AutoCloseable {
       tmp_dir = System.getProperty("java.io.tmpdir");
     }
     jniWrapper = new ExpressionEvaluatorJniWrapper(tmp_dir, listJars);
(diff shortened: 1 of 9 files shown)


xuechendi:
@zhouyuan , I added some script to clean /tmp dir in compilation, do you think that makes sense to you?

zhouyuan:
@xuechendi I'm not sure if i understand this correctly, the compile work maybe done on other nodes?

xuechendi:
@zhouyuan , oh, yes, it is possible do codegen compile in another node, Hmm, I can't think up a nice solution to remove these folders, do you have any idea?

Issue #1535, opened 2020-07-27
[oap-native-sql] Add ColumnarWindowExec


zhztheplayer:
This is to add columnar window support to native SQL engine.

zhztheplayer:
This requires for following subtasks to be done respectively:

1. Support for regular aggregation functions;
2. Support for window functions;
3. Support for partitioning;
4. Support for ordering;
5. Support for frame clauses.

We may break this down to several sub tickets then follow each to track the progress.

PR #1536, opened 2020-07-27
[OAP-1535][oap-native-sql] Add ColumnarWindowExec

zhztheplayer:
See #1535


oap-native-sql/core/hs_err_pid1035461.log
```diff
@@ -0,0 +1,2890 @@
+#
```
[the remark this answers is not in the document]
zhztheplayer: Thanks but this work is still in progress... So they will not appear in the final patch.

Anyway I'll add some entries to .gitignore to get rid of cases like this, finally.

oap-native-sql/cpp/src/codegen/arrow_compute/ext/code_generator_base.h
```diff
@@ -36,8 +36,22 @@ class CodeGenBase {
...
+  virtual arrow::Status MakeResultIterator(
+      std::shared_ptr<arrow::Schema> schema,
+      std::shared_ptr<ResultIterator<arrow::RecordBatch>>* out) {
+    return arrow::Status::NotImplemented(
+        "CodeGenBase MakeResultIterator is an abstract interface.");
+  }
+
   virtual arrow::Status MakeResultIterator(
       std::shared_ptr<arrow::Schema> schema,
+      const ArrayList& in,
```
xuechendi, line 54: That breaks the design, can you use Process interface inside ResultIterator?

oap-native-sql/cpp/src/codegen/arrow_compute/ext/kernels_ext.cc
```diff
@@ -1344,6 +1346,387 @@ arrow::Status HashArrayKernel::Evaluate(const ArrayList& in,
   return impl_->Evaluate(in, out);
 }
 
+WindowPartitionKernel::WindowPartitionKernel(arrow::compute::FunctionContext *ctx):
```
xuechendi, line 1349: I think you can move this kernel to a separate file, since it is bit huge

oap-native-sql/cpp/src/codegen/arrow_compute/ext/window_sort_kernel.h
```diff
@@ -0,0 +1,917 @@
...
+ private:
+  using BuilderType_ = typename arrow::TypeTraits<DataType>::BuilderType;
+  using ArrayType_ = typename arrow::TypeTraits<DataType>::ArrayType;
+  std::unique_ptr<BuilderType_> builder_;
+  std::vector<std::shared_ptr<ArrayType_>> cached_arr_;
+  arrow::compute::FunctionContext* ctx_;
+};
+
+///////////////  SortArraysToIndices  ////////////////
+class WindowSortKernel::Impl {
```
xuechendi, line 120: Since this is Impl, may put in .cc instead of .h?

oap-native-sql/cpp/src/codegen/arrow_compute/ext/code_generator_base.h
zhztheplayer: removed entirely this code, now there remains implementation of `Finish`

oap-native-sql/cpp/src/codegen/arrow_compute/ext/kernels_ext.cc
zhztheplayer: done

oap-native-sql/cpp/src/codegen/arrow_compute/ext/window_sort_kernel.h
zhztheplayer: Let's put this action off... Rui is about to make a complete improvement for sorts. For window we can turn to use that one then. Thus this file is considered temporary.

zhztheplayer:
@xuechendi Could you please review the changes to actions_impl.cc?

https://github.com/Intel-bigdata/OAP/pull/1536/files#diff-c41b77736e161997501ecec40681e462

In window's code there are uses of aggregate actions. It seems that we must avoid redefinition of these methods by separating them to headers and sources.

zhztheplayer:
I am merging this. Others can still review on master code. Anything abnormal I can perform roll back then.

Issue #1548, opened 2020-07-30
[oap-native-sql] Fallback with unsupported expression


xuechendi:
Spark supported great amount of expressions which are lot more than gandiva supported, so in this Issue, I proposed to add a way to check if we support one expression and fallback to use vanilla spark projection when not support.

PR #1549, opened 2020-07-30
[OAP-1548][oap-native-sql] Use rowBased HashAggregate and Project when some exp…

xuechendi:
…ression not supportColumnar

Fixed: https://github.com/Intel-bigdata/OAP/issues/1548


xuechendi:
@rui-mo , @zhouyuan  Q27, Q36, Q86 is runnable using this fallback, while result is still incorrect, bug is in columnarHashAggr, will fix later.
In these three quries, lacks columnar shiftright + bitwise_and + castByte + castString, will add later.
when we enabled columnarShuffle, will saw a segment fault of decompress buffer in shuffle read.

PR #1555, opened 2020-07-31
[OAP-1541][oap-native-sql] TreeNode children not replaced by columnar…

### LINKED ISSUE #1541: [oap-native-sql] TreeNode children not replaced by columnar operators
`TreeNode.withNewChildren(...)` depends on children's implementation of `fastEquals`[1]. That said replacement happens only if new child is **not equal to** old child. However e.g. `ColumnarHashAggregateExec` will be equal to regular `HashAggregateExec` if both have same fields which means the replacement is not always reliable as we expect. Which means we can probably get a `HashAggregateExec` rather than `ColumnarHashAggregateExec` in physical plan even if the columnar rules has been processed on the aggregation node. 

@zhouyuan @xuechendi Have you been experiencing such a problem as I suppose the issue should happen regularly in many cases even a simple `select a, sum(b) from t group by a limit 10`.

[1] https://github.com/Intel-bigdata/spark/blob/3fdfce3120f307147244e5eaf46d61419a723d50/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/trees/TreeNode.scala#L253

zhztheplayer:
… operators

Also, add ColumnarPluginTest.

Closes #1541

### DIFF
--- a/oap-native-sql/core/src/main/scala/com/intel/oap/execution/ColumnarBatchScanExec.scala
+++ b/oap-native-sql/core/src/main/scala/com/intel/oap/execution/ColumnarBatchScanExec.scala
@@ -47,4 +47,14 @@ class ColumnarBatchScanExec(output: Seq[AttributeReference], @transient scan: Sc
       r
     }
   }
+
+  override def canEqual(other: Any): Boolean = other.isInstanceOf[ColumnarBatchScanExec]
+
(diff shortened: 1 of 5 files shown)


xuechendi:
@rui-mo , just checked with Hongze, he has submitted fastEquals PR, FYI

xuechendi:
@zhztheplayer  verified, merge here.

PR #1584, opened 2020-08-09
[WIP][OAP-1583][oap-native-sql] Optimize ColumnarExchange partition id computation

rongma1997:

- SinglePartitioning, RoundRobinPartitioning, HashPartitioning: Move partition id computation from java to native.

- RangePartitioning: Fallback to java.

## How was this patch tested?
Java/cpp unit tests.


zhouyuan:
@rongma1997  I verified this patch locally, TPCH tests is correct. I think we could fix the performance in the next patch

PR #1586, opened 2020-08-10
[OAP-1587][oap-native-sql] tpcds enabling (part1)

xuechendi:
This PR adds support for TPCDS workload(part1), major changes are:
- codegen hash aggr will faultily skip null key, now add validity vector to fix this
- add a optimization when project between two RowBased operator 
- switch to hash64 for hashagg/hashjoin
- limited decimal support
- fallback to row-based operator if there are unsupported expressions

as of this patch, TPCDS queries are all working with below configurations:

```
spark.sql.columnar.codegen.hashAggregate true
spark.sql.columnar.sort.broadcastJoin false
#spark.shuffle.manager org.apache.spark.shuffle.sort.ColumnarShuffleManager
spark.sql.adaptive.enabled false

```
remaining issues will be fixed in part2

Fixed: https://github.com/Intel-bigdata/OAP/issues/1587

### DIFF
--- a/oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarGroupbyHashAggregation.scala
+++ b/oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarGroupbyHashAggregation.scala
@@ -238,8 +238,7 @@ object ColumnarGroupbyHashAggregation extends Logging {
     val mode = if (aggregateExpressions.size > 0) {
       aggregateExpressions(0).mode
     } else {
-      throw new UnsupportedOperationException(
-        s"Unable to handle Aggregation with no aggregate expressions.")
+      null
     }
 
     val originalInputFieldList = originalInputAttributes.toList.map(attr => {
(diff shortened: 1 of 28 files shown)


xuechendi:
By above fixed, TPCDS Q90, Q96 and Q5 is runnable, and Q90 Q96 got correct result.

oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarGroupbyHashAggregation.scala
```diff
@@ -417,37 +432,84 @@ object ColumnarGroupbyHashAggregation extends Logging {
...
-          val aggBufferAttr = min.inputAggBufferAttributes
-          val attr = ConverterUtils.getAttrFromExpr(aggBufferAttr(0))
-          aggregateAttr += attr
+          case _ => {
+            aggregateAttr += aggregateAttributeList(res_index)
+            res_index += 1
+          }
+        }
+        case Count(_) => mode match {
+          case partial => {
```
rui-mo, line 475: should be "Partial" here, or else will not match.

xuechendi:
@zhouyuan , verified in my env and also checked with @rui-mo , passed all tpcds in one spark context. And also verified with TPCH w/wo AQE. Can you do a final check, so we can merge it.

PR #1611, opened 2020-08-14
[OAP-MLlib] fix repartition bug for not enough partitions

xwu99:

Fix repartition bug, repartition to executorNum if not enough partitions

## How was this patch tested?

CI


xwu99:
@JkSelf I will merge this to fix the bug.

JkSelf:
Good catch. Thanks @xwu99

PR #1615, opened 2020-08-18
[OAP-1583][oap-native-sql] Optimize ColumnarExchange partition id computation follow-up

rongma1997:

A follow-up to #1584 

## How was this patch tested?

UT


rongma1997:
@zhouyuan Could you check whether performance issues are resolved? I benchmark TPCH over sr247 cluster and performance is better than last time.

PR #1619, opened 2020-08-18
[OAP-1618][oap-native-sql] Fix columnar reuse exchange

rongma1997:
ReuseExchange is a spark rule that is used to detect same shuffle map stages by analysing the canonicalized spark plan. It will convert the repeated nodes as a single ReusedExchange leaf node. Thus the shuffle map stage will only compute once and other reduce stages can reuse it's output.
Currently, `equals` method in columnar operators determine whether references are equal but not values, which leads to SparkPlan `this.canonicalized == other.canonicalized` always return false. So ReuseExchange cannot correctly identify the same shuffle map stages.

## How was this patch tested?
Add Scala unit test and manually tested TPCH q18
before
image

after
image


rongma1997:
cc @zhztheplayer @xuechendi @zhouyuan

Issue #1647, opened 2020-08-25
[oap-native-sql] row to columnar for decimal


rui-mo:
convert decimal from row to columnar

PR #1649, opened 2020-08-25
[OAP-1647][oap-native-sql] row to columnar for decimal

rui-mo:
This PR adds support for decimal in the RowToColumnar.
 
fixed https://github.com/Intel-bigdata/OAP/issues/1647

### DIFF
--- a/oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarUnaryOperator.scala
+++ b/oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarUnaryOperator.scala
@@ -169,9 +169,8 @@ class ColumnarCast(child: Expression, datatype: DataType, timeZoneId: Option[Str
           if float.getPrecision() == FloatingPointPrecision.DOUBLE => 21
         case date: ArrowType.Date if date.getUnit == DateUnit.DAY => 10
         case decimal : ArrowType.Decimal =>
-          val precision = decimal.getPrecision()
-          val scale  = decimal.getScale()
-          precision
+          // Add two to precision for decimal point and negative sign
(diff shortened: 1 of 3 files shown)


rui-mo:
@xuechendi could you try and review this patch? I can have correct result for q83 on my env with preferColumnar = true.

xuechendi:
Looks great to me, I think we can merge this firstly. @zhouyuan

xuechendi:
@rui-mo , @zhouyuan , verified locally, Q83 is now correct.

oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarUnaryOperator.scala
```diff
@@ -169,9 +169,7 @@ class ColumnarCast(child: Expression, datatype: DataType, timeZoneId: Option[Str
           if float.getPrecision() == FloatingPointPrecision.DOUBLE => 21
         case date: ArrowType.Date if date.getUnit == DateUnit.DAY => 10
         case decimal : ArrowType.Decimal =>
-          val precision = decimal.getPrecision()
-          val scale  = decimal.getScale()
-          precision
+          (decimal.getPrecision() + 2)
```
zhouyuan, line 172: magic number 2?

rui-mo: no, the main change is this:  https://github.com/Intel-bigdata/OAP/pull/1649/files#diff-f17cfe4ae8804014360b8ea71a2041e8R195. We found the data is wrong after row to columnar.

Issue #1669, opened 2020-09-01
[oap-native-sql] TPCH Q1 results is not correct w/ hashagg codegen off


zhouyuan:
TPCH Q1 generated wrong results if hashagg codegen to false 
`spark.sql.columnar.codegen.hashAggregate false`

![image](https://user-images.githubusercontent.com/43917/91783825-e6ecfe80-ec33-11ea-85a1-e6bf44aaee91.png)

PR #1670, opened 2020-09-01
[OAP-1669][oap-native-sql] use distinct ordinal list

rui-mo:
fixed https://github.com/Intel-bigdata/OAP/issues/1669

### DIFF
--- a/oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarAggregation.scala
+++ b/oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarAggregation.scala
@@ -198,7 +198,7 @@ class ColumnarAggregation(
 
   // 4. create aggregate native Expression
 
-  var projectOrdinalList = projectOrdinalListBuffer.toList
+  var projectOrdinalList = projectOrdinalListBuffer.toList.distinct
   val aggregateFieldList = aggregateFieldListBuffer.toList
 
   // 5. create nativeAggregate evaluator


oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarAggregation.scala
```diff
@@ -198,7 +198,7 @@ class ColumnarAggregation(
 
   // 4. create aggregate native Expression
 
-  var projectOrdinalList = projectOrdinalListBuffer.toList
+  var projectOrdinalList = projectOrdinalListBuffer.toList.distinct
```
xuechendi, line 201: @rui-mo , I am not sure about this, this list is originally designed to map original input to pre-aggregate project expression, by adding distinct seems may broke the original idea. Let me check the codes, stay tuned

rui-mo: @xuechendi by removing the "distinct" from line 89 in ColumnarProjection (val inputFieldList = inputList.asScala.toList.distinct) also works. The main issue I see is aggregateProjectCols is not distinct but project schema is distinct.

xuechendi: @rui-mo , I checked the codes, I think this fix is workable! Thanks

Issue #1673, opened 2020-09-01
[oap-native-sql] double round support


rui-mo:
For now, we are using decimal round, which needs data to be casted as decimal first.

PR #1674, opened 2020-09-01
[OAP-1673][oap-native-sql] Adding native double round function

rui-mo:
This patch adds native round function support for double.

initially we made a workaround by using round support for decimal. 

depends on: https://github.com/Intel-bigdata/arrow/pull/91
fixes: https://github.com/Intel-bigdata/OAP/issues/1673


rui-mo:
on sf1000, query: select round(31.05, 1), round(-1234.56789, 4), round(ss_ext_list_price, 1), round(ss_ext_tax, 1) from store_sales
decimal round: 3872ms, wrong result
double round: 3618ms, correct result

zhouyuan:
note: this patch is used in TPC-DS queries, it will not impact TPC-H queries.

PR #1675, opened 2020-09-02
[OAP-1651][oap-native-sql] Adding fallback rules for join/shuffle

xuechendi:
This patch adds another fallback rule for joins under some special cases.
Native code doesnt support whole stage code generation yet, there's a performance gap w/ vanilla Spark. e.g, Q72.
The shuffle code will be switch back to row-based operator to get better performance, if  we already fall back to row-based execution for joins. 

Fixed: https://github.com/Intel-bigdata/OAP/issues/1651

This Commits depends on arrow commit https://github.com/Intel-bigdata/arrow/pull/92

### DIFF
--- a/oap-native-sql/core/src/main/scala/com/intel/oap/ColumnarPlugin.scala
+++ b/oap-native-sql/core/src/main/scala/com/intel/oap/ColumnarPlugin.scala
@@ -127,8 +127,9 @@ case class ColumnarPreOverrides(conf: SparkConf) extends Rule[SparkPlan] {
         plan.withNewChildren(children)
       }
     case plan: ShuffleExchangeExec =>
-      if (columnarConf.enableColumnarShuffle) {
-        val child = replaceWithColumnarPlan(plan.child)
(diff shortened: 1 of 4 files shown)


xuechendi:
@zhouyuan , @rui-mo , please take a look, this is a small commit to further "joinOptimization"

xuechendi:
Hi, @rongma1997 , I added some new murmurhash functions in gandiva so we can generate same hash codes as spark do, please take a look of this PR and the one in arrow side https://github.com/Intel-bigdata/arrow/pull/92

xuechendi:
@zhouyuan , this PR is tested with TPCH and TPCDS, results are all correct. And I removed NormalizedNaNandZero support in this PR, because codes still cause some gandiva project fail for some reason. Please take a look and have a verification, thanks

zhouyuan:
@rongma1997 there's some small changes on native shuffle codes, i verified this locally, will merge this first. please kindly fixes on this if necessary

Issue #1682, opened 2020-09-07
[oap-native-sql] fix aggregate without codegen


rui-mo:
When "spark.sql.columnar.codegen.hashAggregate" is set as false, there are some TPC-DS queries failing or having incorrect result.

PR #1683, opened 2020-09-07
[OAP-1682][oap-native-sql] fix aggregate without codegen

rui-mo:
Fixed query failing and result mismatch when setting "spark.sql.columnar.codegen.hashAggregate" as false.

fixed: https://github.com/Intel-bigdata/OAP/issues/1682


rui-mo:
verified on TPC-DS (sf100 and sf1000) and TPCH.

Issue #1700, opened 2020-09-10
[oap-native-sql] Support inside join condition project


xuechendi:
Currently, if we saw project inside Join Condition, we will fallback to use rowBased join instead.
This issue is aim to support doing project inside condition

PR #1701, opened 2020-09-10
[OAP-1700][oap-native-sql] support join-inside condition project

xuechendi:
Support project inside Join Condition
1. if we can handle this function with cpp code generation, we will directly add project codes inside conditionCheck function.
2. if we can't handle this function currently, we will create gandiva function instead.
3. if both above can't handle this project function, we will do fallback

Fixed: https://github.com/Intel-bigdata/OAP/issues/1700


xuechendi:
@zhouyuan , verified with TPCH and TPCDS

xuechendi:
@zhouyuan , rebased to master

zhouyuan (approved):

zhouyuan:
@xuechendi great feature! thanks for the hard work 👍

Issue #1704, opened 2020-09-10
[oap-native-sql] Add ColumnarUnion and ColumnarExpand


xuechendi:
as titled

PR #1705, opened 2020-09-10
[OAP-1704][oap-native-sql] Support ColumnarUnion and ColumnarExpand

xuechendi:
Fixed: https://github.com/Intel-bigdata/OAP/issues/1704


xuechendi:
@zhouyuan , Verified with TPCDS

Issue #1706, opened 2020-09-11
[oap-native-sql] Optimize columnar shuffle write


rongma1997:
To avoid write twice if no spill.

PR #1710, opened 2020-09-11
[OAP-1706][oap-native-sql] Optimize shuffle write

rongma1997:
The existing shuffle write will firstly produce number of reducer partition files, and then merge them into one data file. This results in twice inevitable file writes. 
This optimization handles two possibilities:
1. If one reducer partition buffer can hold all data, and when split finishes, the buffered data will be directly flushed to the final data file.
2. If the buffer cannot hold all data, which means data will spill to disk in the middle of split. When split finishes, it will firstly copy the spilled file to the final data file, and then flush the buffer.

issue: #1706


rongma1997:
Benchmarked TPCH 1.5T on sr247  cluster w/ AQE. Power run result shows current total shuffle write time is 100s and the optimized one is 80s.

oap-native-sql/core/src/main/scala/org/apache/spark/shuffle/ColumnarShuffleWriter.scala
```diff
@@ -58,7 +60,6 @@ class ColumnarShuffleWriter[K, V](
   private val localDirs = blockManager.diskBlockManager.localDirs.mkString(",")
   private val nativeBufferSize =
```
zhouyuan, line 61: the buffer size seems to control the number of batches in each write, maybe we should use another option for this behavior?

PR #1711, opened 2020-09-14
[OAP-1694][oap-native-sql][Scala] fix hash join w/ empty batch

### LINKED ISSUE #1694: [oap-native-sql] TPC-H q15 failed for ConditionedProbeArraysVisitorImpl MakeResultIterator does not support dependency type other than Batch
Error: Error running query: java.io.IOException: nativeFinishForIterator: finish failed with error msg Invalid: ConditionedProbeArraysVisitorImpl MakeResultIterator does not support dependency type other than Batch. (state=,code=0)

zhouyuan:

This patch fixes the hash join w/ empty batch. 
fixes #1694 

## How was this patch tested?
locally verified with TPCH Q15

### DIFF
--- a/oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarShuffledHashJoin.scala
+++ b/oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarShuffledHashJoin.scala
@@ -80,28 +80,15 @@ class ColumnarShuffledHashJoin(
       buildIter: Iterator[ColumnarBatch]): Iterator[ColumnarBatch] = {
 
     var _buildTime: Long = 0
+    var htSize :Long = 0
 
     while (buildIter.hasNext) {
       if (build_cb != null) {
         build_cb = null
       }
       build_cb = buildIter.next()
-      if (build_cb == null) {
-        val res = new Iterator[ColumnarBatch] {
-          override def hasNext: Boolean = {
-            false
-          }
-
-          override def next(): ColumnarBatch = {
-            val resultColumnVectors = ArrowWritableColumnVector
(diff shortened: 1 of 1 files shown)


xuechendi:
@zhouyuan , I think this fix may cause some logical error. I comment inline

oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarShuffledHashJoin.scala
```diff
@@ -86,7 +86,7 @@ class ColumnarShuffledHashJoin(
         build_cb = null
       }
       build_cb = buildIter.next()
-      if (build_cb == null) {
+      if (build_cb.numRows == 0) {
         val res = new Iterator[ColumnarBatch] {
           override def hasNext: Boolean = {
             false
```
xuechendi, line 92: this iterator is used to indicate that this is the last iterator, and when numRows is 0, this only means one build_cb is empty

oap-native-sql/core/src/main/scala/com/intel/oap/expression/ColumnarShuffledHashJoin.scala
```diff
@@ -128,9 +115,7 @@ class ColumnarShuffledHashJoin(
         projectedBuildKeyCols.foreach(v => v.close)
       }
     }
-    if (build_cb != null) {
-      build_cb = null
-    } else {
+    if (build_cb == null || htSize == 0) {
```
zhouyuan, line 118: @xuechendi modify the logic to work only when all batches are empty

PR #1724, opened 2020-09-17
[oap-native-sql]Change ResultIterator interface to be more flexible

xuechendi:


xuechendi:
@rui-mo , the new resultIterator interface is here.
Basically, anyone uses result_iterator can create something like ResultIterator<ArrayList> to return the real data inside, and this interface can be called by not only cpp, but also inside scala

rui-mo:
got it!

PR #1726, opened 2020-09-17
[OAP]Add prompts for users to install Java8 if there is no Java in env.

yao531441:

Add prompts for users to install Java8 if there is no Java in env.


HongW2019:
LGTM

PR #1731, opened 2020-09-18
[OAP-CACHE] [OAP-1730] [POAE-428] Add OAP cache runtime enable

jikunshang:
Add OAP cache runtime enable, will fall back to default Parquet/ORC file format if disabled.

## How was this patch tested?

Add ut and did e2e test.

### DIFF
--- a/oap-cache/oap/src/main/scala/org/apache/spark/sql/internal/oap/OapConf.scala
+++ b/oap-cache/oap/src/main/scala/org/apache/spark/sql/internal/oap/OapConf.scala
@@ -144,6 +144,13 @@ object OapConf {
       .booleanConf
       .createWithDefault(false)
 
+  val OAP_CACHE_RUNTIME_ENABLE =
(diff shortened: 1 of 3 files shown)


jikunshang:
@winningsix @offthewall123 @yma11  Please take a review.  Thanks.

oap-cache/oap/src/main/scala/org/apache/spark/sql/internal/oap/OapConf.scala
```diff
@@ -144,6 +144,13 @@ object OapConf {
       .booleanConf
       .createWithDefault(false)
 
+  val OAP_CACHE_RUNTIME_ENABLE =
+    SqlConfAdapter.buildConf("spark.sql.oap.cache.runtime.enable")
```
winningsix, line 148: minor hit: spark.sql.oap.cache.enabled

winningsix (approved):
A minor comment. You can go ahead to merge.

Issue #1734, opened 2020-09-18
[oap-native-sql] use non-codegen for sort with one key


rui-mo:
We are now using codegen for sort with one key inside several cols. If we change this case into non-codegen version, we suspect the compilation time will be reduced.

PR #1735, opened 2020-09-18
[OAP-1734][oap-native-sql]use non-codegen for sort with single key

rui-mo:
1. changes sort with one key inside several cols into non-codegen version. 
2. fixes data type issue and null value issue for InplaceSort.

Fixes: https://github.com/Intel-bigdata/OAP/issues/1734

### DIFF
--- a/oap-native-sql/cpp/src/codegen/arrow_compute/ext/sort_kernel.cc
+++ b/oap-native-sql/cpp/src/codegen/arrow_compute/ext/sort_kernel.cc
@@ -572,6 +572,7 @@ class SortInplaceKernel : public SortArraysToIndicesKernel::Impl {
     auto valid_indices_end = indices_end;
     if (nulls_total_ > 0) {
       // we use arrow sort for this scenario
+      // the indices_out of arrow sort is nulls_last
       std::shared_ptr<arrow::Array> indices_out;
       RETURN_NOT_OK(
           arrow::compute::SortToIndices(ctx_, *concatenated_array_.get(), &indices_out));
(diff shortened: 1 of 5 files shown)


rui-mo:
@zhztheplayer Here is an initial version, and it only works for asc and null first case. The gaps are:
1. In function FinishInternal, need to support different sort cases.
2. Need to support string type.
3. move Appender class to a new file.

oap-native-sql/cpp/src/codegen/arrow_compute/ext/sort_kernel.cc
```diff
@@ -630,18 +630,20 @@ class SortInplaceKernel : public SortArraysToIndicesKernel::Impl {
...
       auto length = (total_length_ - offset_) > batch_size_ ? batch_size_
                                                             : (total_length_ - offset_);
-      uint64_t count = 0;
+      uint64_t valid_count = 0;
+      uint64_t total_count = 0;
       if (offset_ >= nulls_total_) {
-        while (count < length) {
-          RETURN_NOT_OK(builder_0_->Append(result_arr_->GetView(offset_ + count++)));
+        while (total_count < length) {
+          RETURN_NOT_OK(builder_0_->Append(result_arr_->GetView(offset_ + total_count++)));
```
zhouyuan, line 637: the logic seems complicated... can you move the total_count++ to a separate line? so it should be much easier to review

oap-native-sql/cpp/src/codegen/arrow_compute/ext/sort_kernel.cc
```diff
@@ -630,19 +631,36 @@ class SortInplaceKernel : public SortArraysToIndicesKernel::Impl {
     arrow::Status Next(std::shared_ptr<arrow::RecordBatch>* out) {
       auto length = (total_length_ - offset_) > batch_size_ ? batch_size_
                                                             : (total_length_ - offset_);
+      /* Here we take value from the sorted result_arr_ and append to builder.
+      valid_count is used to count the valid value, for accessing the valid value
+      in result_arr_.
+      total_count is used to count both valid and null value, for determing if all values
+      are appended to builder in while loop.
+      */
```
zhouyuan, line 639: great notes!

rui-mo:
@zhouyuan @xuechendi verified on TPC-H. TPC-DS queries are all runnable now but having result mismatch, see https://github.com/Intel-bigdata/OAP/issues/1751.

zhouyuan:
notes: 
- sort by multiple keys is using codegen code path
- sort by single key but w/o payload is using inplace sort
- sort by single key w/ payload is using non-codegen code path

The multiple keys sorting w/ different orders, w/ nulls first/last will be fixed in another PR

zhouyuan (approved):

PR #1746, opened 2020-09-23
[OAP-1741][rpmem-shuffle]To make java side native library from jar directly

Eugene-Mark:
Make PmemBuffer to load native library from jar directly instead of fix path of system.


Jian-Zhang:
LGTM.

PR #1761, opened 2020-09-28
[oap-native-sql] quick fix sort on string by fallback to row

### LINKED ISSUE #1760: [oap-native-sql] fix columnar sorting on string
the query results on tpch q16, q20 is not identical to vanilla spark under thrift-server. The issue seems due to uncorrect sort result returned w/ string.

zhouyuan:

It seems there's an unknown bug on sorting by string. We'll fall
back to row based sort on this case as a quick fix

fixes #1760 

## How was this patch tested?
locally verified on TPCH


zhouyuan:
note: @rui-mo will provide a better fix in next PR.

Issue #1765, opened 2020-09-28
[oap-native-sql] Support WSCG in nativesql


xuechendi:
Support WSCG in nativesql

PR #1766, opened 2020-09-28
[OAP-1765][oap-native-sql] Support WSCG for nativesql and use non-codegen join for remainings

xuechendi:
This PR is aim to support WSCG for nativesql.
For now, all TPCH queries are runnable and correct.

Fixed: https://github.com/Intel-bigdata/OAP/issues/1765


xuechendi:
@zhouyuan , passed all TPCH queries, let's see if we can do some performance test on our TEST ENV firstly.

xuechendi:
@zhouyuan , added the switch to enable and disable wholestagecodegen
spark.oap.sql.columnar.wholestagecodegen false

oap-native-sql/cpp/src/codegen/arrow_compute/ext/codegen_common.cc
```diff
@@ -37,9 +40,6 @@ std::string BaseCodes() {
   return R"(
 #include <arrow/compute/context.h>
 #include <arrow/record_batch.h>
-#include <math.h>
-#include <numeric>
-#include <limits>
```
zhouyuan, line 43: it looks like we need to keep these headers, otherwise it will break codegen of hashagg/SMJ

oap-native-sql/cpp/src/codegen/arrow_compute/ext/codegen_common.cc
```diff
@@ -308,11 +513,14 @@ arrow::Status CompileCodes(std::string codes, std::string signature) {
...
-                    " -O3 -march=native -shared -fPIC -lspark_columnar_jni 2> " + logfile;
+                    " -O1 -march=native -shared -fPIC -lspark_columnar_jni 2> " + logfile;
 #ifdef DEBUG
   std::cout << cmd << std::endl;
 #endif
-  int ret = system(cmd.c_str());
+  int ret;
+  int elapse_time = 0;
+  TIME_MICRO(elapse_time, ret, system(cmd.c_str()));
+  std::cout << "CodeGeneration took " << TIME_TO_STRING(elapse_time) << std::endl;
```
zhouyuan, line 523: move this output to debug section?

oap-native-sql/cpp/src/codegen/arrow_compute/ext/codegen_common.cc
```diff
@@ -308,11 +513,14 @@ arrow::Status CompileCodes(std::string codes, std::string signature) {
   std::string cmd = env_gcc + " -std=c++14 -Wno-deprecated-declarations " + arrow_header +
                     arrow_lib + arrow_lib2 + nativesql_header + nativesql_header_2 +
                     nativesql_lib + cppfile + " -o " + libfile +
-                    " -O3 -march=native -shared -fPIC -lspark_columnar_jni 2> " + logfile;
+                    " -O1 -march=native -shared -fPIC -lspark_columnar_jni 2> " + logfile;
```
zhouyuan, line 516: let's keep using `O3`, `O1` seems consuming much more memory in local tests

zhouyuan:
note: this patch adds initial support of columnar whole stage codegen.
`spark.oap.sql.columnar.wholestagecodegen false` is still required for TPC-DS due to hash conflict. A follow-up patch with new hashmap implementation will fix the issue

PR #1779, opened 2020-10-10
[OAP-1778][oap-native-sql] optimize sort process for array w/o nulls

### LINKED ISSUE #1778: [oap-native-sql] optimize columnar sort
- [x] improve for array does not contain nulls

- [x] improve for nulls_first/nulls_last

zhouyuan:
##  What changes were proposed in this pull request?
improve sort performance if array does not contain nulls.
fixes #1778 

## How was this patch tested?
locally tested

### DIFF
--- a/oap-native-sql/cpp/src/codegen/arrow_compute/ext/sort_kernel.cc
+++ b/oap-native-sql/cpp/src/codegen/arrow_compute/ext/sort_kernel.cc
@@ -763,27 +763,35 @@ class SortOnekeyKernel<DATATYPE, CTYPE, enable_if_number<CTYPE>>
     // we should support nulls first and nulls last here
     // we should also support desc and asc here
     for (int array_id = 0; array_id < num_batches_; array_id++) {
-      for (int64_t i = 0; i < length_list_[array_id]; i++) {
-        if (nulls_first_) {
-          if (!cached_key_[array_id]->IsNull(i)) {
-            (indices_begin + nulls_total_ + indices_i)->array_id = array_id;
-            (indices_begin + nulls_total_ + indices_i)->id = i;
-            indices_i++;
+      if (cached_key_[array_id]->null_count()) {
+        for (int64_t i = 0; i < length_list_[array_id]; i++) {
+          if (nulls_first_) {
+            if (!cached_key_[array_id]->IsNull(i)) {
+              (indices_begin + nulls_total_ + indices_i)->array_id = array_id;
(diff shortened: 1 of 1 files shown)


oap-native-sql/cpp/src/codegen/arrow_compute/ext/sort_kernel.cc
```diff
@@ -763,28 +763,36 @@ class SortOnekeyKernel<DATATYPE, CTYPE, enable_if_number<CTYPE>>
...
-            (indices_begin + indices_null)->array_id = array_id;
-            (indices_begin + indices_null)->id = i;
-            indices_null++;
-          }
-        } else {
-          if (!cached_key_[array_id]->IsNull(i)) {
-            (indices_begin + indices_i)->array_id = array_id;
-            (indices_begin + indices_i)->id = i;
-            indices_i++;
+      if (cached_key_[array_id]->null_count()) {
```
zhouyuan, line 766: @rui-mo can you please take a look here? this could benefit sorting w/ big number rows

oap-native-sql/cpp/src/codegen/arrow_compute/ext/sort_kernel.cc
```diff
@@ -989,28 +997,36 @@ class SortOnekeyKernel<DATATYPE, CTYPE, enable_if_string<CTYPE>>
...
+              indices_null++;
+            }
           }
         }
+      } else {
+        for (int64_t i = 0; i < length_list_[array_id]; i++) {
+          (indices_begin + nulls_total_ + indices_i)->array_id = array_id;
+          (indices_begin + nulls_total_ + indices_i)->id = i;
+          indices_i++;
+        }
```
rui-mo, line 1029: - should we add support nulls_last_ also?

            (indices_begin + indices_i)->array_id = array_id;
            (indices_begin + indices_i)->id = i;
            indices_i++;

- For codegen sort, I can add this optimization also.

zhouyuan: the code path would be reached only when cached_key_[array_id]->null_count() == 0, so i assume null_first and null_last should be the same?

rui-mo: cached_key_[array_id]->null_count() == 0  means this array has no null value, but nulls_total_ may not be zero?

zhouyuan:
replaced by #1968

PR #1785, opened 2020-10-12
[OAP-1765][oap-native-sql] Support WSCG for nativesql(PART 2)

xuechendi:
This PR is aimed to fix hash collision issue

We will introduce a row-wised format to store hash keys, then compared multiple key in hashmap to handle hash collision
By using this method, all tpcds and tpch queries generate correct result.

Fixed: #1765


xuechendi:
@zhouyuan , new PR to fix hash collision

PR #1788, opened 2020-10-13
[OAP-1765][oap-native-sql] Fix for dropped CoalecseBatches before ColumnarBroadcastExchange

xuechendi:
Since current columnarBHJ is applied after all additional Rules applied, previous optimization to remove CoalecseBatches when child is ColumnarToRow leads to CoalecseBatches removed in ColumnarBroadcastExchange, which shows huge performance degradation.

Fixed in this PR


xuechendi:
@zhouyuan

zhouyuan (approved):

PR #1807, opened 2020-10-19
[OAP-1806][oap-native-sql] Fix Columnar Shuffle Memory Leak

rongma1997:
issue: #1806


rongma1997:
@zhztheplayer @zhouyuan I've verified with repartition and checked log locally and there are no warning for memory leak detected after this fix.

PR #1819, opened 2020-10-22
[OAP-1828][oap-shuffle]Update PmofShuffleManager.scala log

kelvin-qin:
The log content is inconsistent with the actual Class Name，it should be PmofShuffleManager.

Fixed printing the wrong class name.


kelvin-qin:
Anybody check it?Thanks.

PR #1830, opened 2020-10-23
[OAP-1829][oap-native-sql] Optimize columnar shuffle and option to use AVX-512

rongma1997:
This patch includes two parts:
1. Optimize the split in columnar shuffle. This is the default split version w/o AVX-512. TPCH 1.5T power run is 60s less than before.(974s vs 1074s)
2. Add a new split version that uses the AVX-512 gather/scatter intrinsics. Because of the limitations of the instructions themselves, the changes are applied only to 32-bit and 64-bit fixed-width data type. The way other data types are handled is not changed.

AVX-512 split can be built with setting cmake option USE_AVX512=ON.

issue: #1829


zhouyuan:
notes: to build w/ avx512, cpp/compile.sh need to be modified to use the building instruction.  

```
diff --git a/oap-native-sql/cpp/compile.sh b/oap-native-sql/cpp/compile.sh
index 8f89e491b5..38af4973c6 100755
--- a/oap-native-sql/cpp/compile.sh
+++ b/oap-native-sql/cpp/compile.sh
@@ -11,7 +11,7 @@ if [ -d build ]; then
 fi
 mkdir build
 cd build
-cmake ..
+cmake .. -DUSE_AVX512=ON
 make

 set +eu

```

PR #1841, opened 2020-10-27
[OAP-1579][OAP-cache]Fix web UI to show cache size

Aaxida:
Modify codes according to the former commit.

Tested manually, and the following is the result.

屏幕截图 2020-10-27 190344

### DIFF
--- a/oap-cache/oap/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
+++ b/oap-cache/oap/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
@@ -1116,7 +1116,11 @@ class ExternalCache(fiberType: FiberType) extends OapCache with Logging {
   override def cacheStats: CacheStats = {
     val array = new Array[Long](4)
     // TODO:total size will be incorrect due to it's an external cache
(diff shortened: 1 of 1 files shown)


oap-cache/oap/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -1116,9 +1116,8 @@ class ExternalCache(fiberType: FiberType) extends OapCache with Logging {
   override def cacheStats: CacheStats = {
     val array = new Array[Long](4)
     // TODO:total size will be incorrect due to it's an external cache
-    // it will influence performance a little.
-    // plasmaClientPool(clientRoundRobin.getAndAdd(1) % clientPoolSize).metrics(array)
-    cacheTotalSize = new AtomicLong(0)
+    plasmaClientPool(clientRoundRobin.getAndAdd(1) % clientPoolSize).metrics(array)
```
winningsix, line 1119: @jikunshang this will impact performance a little bit?

jikunshang: I think it won't. This part will be called period, I remember it's 2 second.

jikunshang:
LGTM, merged.

PR #1845, opened 2020-10-28
[OAP-1844][oap-native-sql] Fix columnar shuffle spilled file not deleted

rongma1997:


zhouyuan:
@rongma1997 thanks for the quick fix!

PR #1855, opened 2020-11-02
[OAP-1854][oap-native-sql] Fix columnar shuffle file not deleted

rongma1997:


oap-native-sql/core/src/main/scala/com/intel/oap/ColumnarPlugin.scala
```diff
@@ -548,13 +542,6 @@ case class ColumnarPostOverrides(conf: SparkConf) extends Rule[SparkPlan] {
     case plan: RowToColumnarExec =>
       val child = replaceWithColumnarPlan(plan.child)
       RowToArrowColumnarExec(child)
-    case ReusedExchangeExec(id, s: ShuffleExchangeExec)
-        if SQLConf.get.adaptiveExecutionEnabled && columnarConf.enableColumnarShuffle =>
-      val exchange = ColumnarShuffleExchangeExec.exchanges.get(s) match {
-        case Some(e) => e
-        case None => throw new IllegalStateException("Reused exchange operator not found.")
-      }
```
zhouyuan, line 556: does this mean we discard reuseexchange w/ AQE enabled?

rongma1997: I tested resue exchange w/ AQE and it still work. I think this part of code is in the wrong direction.

PR #1857, opened 2020-11-04
[OAP-1833][oap-native-sql] Fix HashAggr hasNext won't stop issue

xuechendi:


xuechendi:
@zhouyuan , I added a quick fix, can you take a look?

PR #1874, opened 2020-11-08
[OAP-1873][oap-native-sql] Columnar shuffle split variable length use UnsafeAppend

rongma1997:
1. Treat the sizes of the first input recordbatch's variable length arrays as empirical sizes for the destination buffers, and reserve data for each destination buffer when allocate.
2. Use ArrayBuilder->UnsafeAppend instead of ArrayBuilder->Append


rongma1997:
There's performance regression. I'll take further look into it.

Issue #1888, opened 2020-11-12
[oap-native-sql] Add hash collision check for all HashJoins and hashAggr


xuechendi:
previously, we did some optimization for HashMap to only compare hash 64 key instead a hashKey + original data.
Even though it shows correct result in out TPCH and TPCDS tests, it still leave some possibility of getting incorrect result for some corner cases.
Solution: Add a configurable parameter to enable and disable hash collision check.

PR #1889, opened 2020-11-12
[OAP-1888][oap-native-sql]Add hash collision check for all HashJoins and hashAggr

xuechendi:
Fixed: https://github.com/Intel-bigdata/OAP/issues/1888

updated 2020/11/12
a. Add a config spark.oap.sql.columnar.hashCompare default false
b. Saw performance drop in TPCH Q1 after merge multiple key as unsafeRow and compare


xuechendi:
@zhouyuan , spark.oap.sql.columnar.hashCompare  can used to enable / disable hashcheck

zhouyuan:
notes: the performance of hash compare seems much lower than we expected, nativeMalloc seems suspicious on this. 
Using a new malloc instead of glibc may help: 
https://github.com/microsoft/mimalloc
https://github.com/jemalloc/jemalloc

zhouyuan (approved):

Issue #1903, opened 2020-11-16
[oap-native-sql] BHJ related UT fix


xuechendi:
BHJ will fail in Local Mode, we need to fix this issue

PR #1904, opened 2020-11-16
[oap-native-sql] Fix Local Mode BHJ related UT fail issue

xuechendi:
Fixed: https://github.com/Intel-bigdata/OAP/issues/1903

Updated 2020/11/16:
    1. noticed this issue is caused by BroadcastExchange will delete original data since it has been broadcasted, while in local mode, this obj will also used by next stage. So we now add a timeoutClose here.


xuechendi:
@rui-mo , This fix should pass all BHJ related failure, can you have a test.

rui-mo:
verified on unit tests and TPC-DS queries.

PR #1911, opened 2020-11-17
[OAP-1852][oap-native-sql] Enable memory pool in columnar shuffle

rongma1997:


zhztheplayer:
fixed in https://github.com/Intel-bigdata/OAP/commit/58c05ea1ac07ce25868a1f02aaa06542d0495afc

PR #1922, opened 2020-11-19
[OAP-1921][rpmem-shuffle] For BDBA analysis to exclude unused library

Eugene-Mark:
Exclude unused library libsqlitejdbc.jnilib which is for MAC OS, the modification passed build process.


Jian-Zhang:
LGTM.

PR #1928, opened 2020-11-20
[OAP-1892][oap-native-sql] Optional cache spilled data in memory

rongma1997:
Add configuration "spark.oap.sql.columnar.shuffle.preferSpill", defaultValue = true
Whether to spill the partition buffers when buffers are full. If false, the partition buffers will be cached in memory first, and the cached buffers will be spilled when reach maximum memory.


zhouyuan:
@rongma1997 does this overwrite #1913 ?

rongma1997:
> 
> 
> @rongma1997 does this overwrite #1913 ?

May submit more commits to both this branch and #1913 . These two are in different directions now.

PR #1961, opened 2020-11-27
[OAP-1924][OAP-CACHE]Decouple hearbeat message and use conf to determine whether to report locailty information for branch branch-1.0-spark-3.x

Aaxida:
We are using external DB to store cache locality info, but the old OAP report mechanism still reports cache locality to the driver through RPC. Need to decouple this.

1:Using external DB => not report cache locality to the driver through RPC. But still, need to report cache metrics to the driver through RPC.
2:Using report mechanism =>follow current behavior.

In the code, it will check the configuration of whether to use Redis.
If used, just report metrics information.
If not, report both metrics and locality information.


jikunshang:
LGTM.  @offthewall123  Please take a review, thanks.

offthewall123:
LGTM, if you had an E2E test to verify it. @Aaxida

PR #1962, opened 2020-11-27
[OAP-1963] Fix for wrong conda channels selection to use Intel(R) distribution for Python*

dmsivkov:
Install Intel packages (numpy, MKL, oneDAL) instead of default. With up to 2x performance.

## How was this patch tested?
Docker image built successfully. 

N/A


jerrychenhf:
Hi Dmitry,
Thank you for the change.
The change looks OK for me. While we will also do a check whether the replaced libraries installed from Intel channel will break the functionality or not. Will update you and we will merge soon when our check it's done. 

Thanks,
Haifeng

PR #1966, opened 2020-11-30
[OAP-1947][oap-native-sql][C++] reduce sort kernel memory footprint

zhouyuan:

reduce memory usage for sort kernel by using more compact data structure

## How was this patch tested?

locally verified


zhouyuan:
Note: this patch only make Sort using a smaller data structure - the logic is not touched

Issue #1967, opened 2020-11-30
[oap-native-sql] NaN support and optimization in columnar sort


rui-mo:
In columanr sort:
1.  we should add switchable NaN check for sort kernels; 
2. To remove the dependency to arrow sort, we can replace arrow sort with the rewritten code. 
3. In Inplace sort, we can change multiple times of value copy into one-time memcopy.

PR #1968, opened 2020-11-30
[OAP-1967][oap-native-sql] add NaN support and optimization in columnar sort

rui-mo:
In this pr:
1. we added switchable NaN check for sort kernels;
2. to remove the dependency to arrow sort and improve the performance, we replaced arrow sort with the rewritten code;
3. to improve the performance, we changed multiple times of value copy into one-time memcopy in Inplace sort;
4. improved sort partition if array does not contain nulls.

fixes: https://github.com/Intel-bigdata/OAP/issues/1967
fixes: https://github.com/Intel-bigdata/OAP/pull/1778

### DIFF
--- a/oap-native-sql/cpp/src/codegen/arrow_compute/ext/sort_kernel.cc
+++ b/oap-native-sql/cpp/src/codegen/arrow_compute/ext/sort_kernel.cc
@@ -43,27 +43,25 @@
 #include "utils/macros.h"
 
 /**
- *               The Overall Implementation of Sort Kernel
+                 The Overall Implementation of Sort Kernel
  * In general, there are three kenels to use when sorting for different data.
    They are SortInplaceKernel, SortOnekeyKernel and SortArraysToIndicesKernel.
- * If sorting for one non-string col without payload, SortInplaceKernel is used.
-   In this kernel, if sorted data has no null value, ska_sort is used for asc 
-   direciton, and std sort is used for desc direciton. If sorted data has null
-   value, arrow sort is used.
- * If sorting for one col with payload, or one string col without payload,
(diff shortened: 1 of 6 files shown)


oap-native-sql/cpp/src/codegen/arrow_compute/ext/sort_kernel.cc
```diff
@@ -43,27 +43,25 @@
 #include "utils/macros.h"
 
 /**
- *               The Overall Implementation of Sort Kernel
+                 The Overall Implementation of Sort Kernel
```
zhouyuan, line 46: thanks for maintaining this notes!

oap-native-sql/cpp/src/codegen/arrow_compute/ext/sort_kernel.cc
```diff
@@ -805,55 +1003,137 @@ class SortInplaceKernel : public SortArraysToIndicesKernel::Impl {
...
+        const std::shared_ptr<arrow::Buffer>& buffer) {
+        ARROW_ASSIGN_OR_RAISE(auto out, AllocateBuffer(size * length_, pool_));
+        auto out_data = out->mutable_data();
+        auto data_begin = buffer->data();
+        std::memcpy(out_data, data_begin + size * offset_, size * length_);
+        return std::move(out);
+      }
+      
+      arrow::Status SliceBuffer(const std::shared_ptr<arrow::Buffer>& buffer) {
+        return SliceBufferImpl(buffer).Value(&out_data_.buffers[1]);
```
xuechendi, line 1075: I think maybe we should use two args in this function, one for input and one for output, since use member variable "out_data_" is a little strange to me.

zhouyuan (approved):

PR #1988, opened 2020-12-02
[OAP-1983][oap-native-sql] Fix Q38 and Q87 when unsafeRow contains null

xuechendi:
When hashcheck is enabled, observed that Q38 and Q87 will generate incorrect result.
Root cause is setNull function in unsafeRow has some bug, which leads to incorrect validity. So fix in this PR

Fixed: https://github.com/Intel-bigdata/OAP/issues/1989


xuechendi:
@zhouyuan , fixed incorrect result issue in Q38 and Q87, this PR is quite simple

Issue #2007, opened 2020-12-07
[oap-native-sql]utility to verify row/columnar results


zhouyuan:
We should provide a utility to verify the results. 
The floating points precision issue is a bit annoying - this will be fixed once we support decimal

PR #2009, opened 2020-12-07
[oap-native-sql] Add a new folder to store ut and post_process scripts

xuechendi:
Fixed: https://github.com/Intel-bigdata/OAP/issues/2007

### DIFF
--- a/oap-native-sql/tools/process.py
+++ b/oap-native-sql/tools/process.py
@@ -1,0 +1,176 @@
+#!/usr/bin/python
+
+import sys
+import pprint
+import re
+import traceback
+import copy
+from collections import OrderedDict
+import numpy as np
+import collections
+
+def process(input):
+    try:
+        return int(input)
+    except:
+        try:
+            return float(input)
+        except:
+            return input
+
+def loadResultAsDict(path):
+    result = OrderedDict()
+    query = ""
+    table_status = 0
+    table_title = []
+    table = []
+    try:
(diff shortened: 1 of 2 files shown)


xuechendi:
@rui-mo , @zhouyuan 
I submit this process script here.

oap-native-sql/tools/process.py
```diff
@@ -0,0 +1,176 @@
...
+def loadResultAsDict(path):
+    result = OrderedDict()
+    query = ""
+    table_status = 0
+    table_title = []
+    table = []
+    try:
+        with open(path, 'r') as f:
+            for line in f.readlines():
+                if line[0:13] == "[SPARK_QUERY]":
```
zhouyuan, line 30: on thrift-server there's one more corner case: the query is executed, however the results is empty due to some issue. here's one example i ran into: 
```
. . . . . . . . . . . . . .>         case when avg_monthly_sales > 0 then abs(sum_sales - avg_monthly_sales) / avg_monthly_sales else null end > 0.1
. . . . . . . . . . . . . .>  order by sum_sales - avg_monthly_sales, 3
. . . . . . . . . . . . . .>  limit 500
. . . . . . . . . . . . . .>
. . . . . . . . . . . . . .>
. . . . . . . . . . . . . .> 0: jdbc:hive2://sr606:10001> use arrow_tpcds_500;
```

PR #2017, opened 2020-12-08
[OAP-2014][OAP-COMMON][POAE7-545] Add clflush support for pmem write

xieqi:

Add clflush for pmem write.

## How was this patch tested?

use the built-in test case.


winningsix:
@xieqi can you check the failed travis issue? Basically this JIRA is talking about explicitly flushing data onto PMEM to avoid random access? Let's wait and see how it impacts performance in multi-thread. How about introducing no-ops work? It's different thing, correct? BTW, should we have another PRs changing the parts who use this new APi?

Issue #1995, opened 2020-12-03
[oap-nativesql] support NaN check in hashagg


zhouyuan:
do not use Inf to replace Nan in hashagg, we should allow Nan comparison

PR #2032, opened 2020-12-10
[OAP-1995][oap-native-sql] add NaN support in comparison

rui-mo:

This pr aims to fix NaN and Infinity result mismatch in TPC-DS q17, q39a, q39b, by adding NaN comparison support.
fixes: https://github.com/Intel-bigdata/OAP/issues/1995
depends on: https://github.com/Intel-bigdata/arrow/pull/101

## How was this patch tested?
locally tested


rui-mo:
spark.sql.columnar.sort.NaNCheck should be set **true** to get correct result for q17, q39a, q39b

rui-mo:
@zhouyuan verified TPC-DS/H with correct results

PR #2042, opened 2020-12-11
[OAP-2041][oap-native-sql] skip null checking if allowed

zhouyuan:

when joining, we could skip the if_is_null checking if the whole batch didn't contain `null`

## How was this patch tested?

locally verified on TPC-H/TPC-DS


zhouyuan:
notes: the CI failures are not related

PR #2059, opened 2020-12-16
[OAP-2057][oap-native-sql]package dependencies into the jar file

weiting-chen:
## What changes were proposed in this pull request?

1. Add Arrow Header Files into third_party/arrow
2. Add libarrow.so.17 into the jar file
3. Add libgandiva.so.17 into the jar file
4. Add libparquet.so.17 into the jar file
5. Set to build libprotobuf from source to true

Closes #2057 
Closes #2058 

(Please fill in changes proposed in this fix)

## How was this patch tested?

Tested on my own development server and another server
(Please explain how this patch was tested. E.g. unit tests, integration tests, manual tests)

(If this patch involves UI changes, please attach a screenshot; otherwise, remove this)

PR #2065, opened 2020-12-18
[OAP-2057][oap-native-sql]do not package pre-built binary into jar

zhouyuan:
Revert "[OAP-2057][oap-native-sql]package dependencies into the jar file (#2059)"

This reverts commit d51156a952e8848484db68b63f067d8f39828267.


zhouyuan:
@weiting-chen this patch reverts packaging with dependency

PR #2069, opened 2020-12-18
[OAP-1804][Doc]Using NUMA-patched Spark to get better performance for…

HongW2019:
… external strategy

Add details on using NUMA-patched Spark to get better performance for `external` strategy on PMem.

## How was this patch tested?

N/A

### DIFF
--- a/oap-cache/oap/docs/User-Guide.md
+++ b/oap-cache/oap/docs/User-Guide.md
@@ -254,6 +254,8 @@ Install `numactl` to bind the executor to the PMem device on the same NUMA node.
 
    ```yum install numactl -y ```
 
+We recommend you use NUMA-patched Spark to achieve better performance gain for the `external` strategy compared with Community Spark.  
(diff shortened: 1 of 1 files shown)


oap-cache/oap/docs/User-Guide.md
```diff
@@ -254,6 +254,8 @@ Install `numactl` to bind the executor to the PMem device on the same NUMA node.
 
    ```yum install numactl -y ```
 
+We recommend you use NUMA-patched Spark to achieve better performance gain for the `external` strategy compared with Community Spark.
```
HongW2019, line 257: @offthewall123 @jikunshang Do we need keep this NUMA-patched Spark suggestion for external stategy in OAP 0.8.4 and OAP 1.0.0 ?

jikunshang: Yes, we need.

PR #2077, opened 2020-12-21
[OAP-2078][oap-native-sql] Refine ColumnarPlugin rules to make it more readable and modifiable

xuechendi:
Fixed: https://github.com/Intel-bigdata/OAP/issues/2078


rui-mo:
@xuechendi verified on TPC-DS sf1000 and TPC-H sf500 with correct results. Unit tests in repo are all runnable.

Issue #1856, opened 2020-11-04
[oap-native-sql] should follow upstream design on AQE


zhouyuan:
https://issues.apache.org/jira/browse/SPARK-32332

- [ ] columnar shuffle
- [ ] columnar BHJ
- [x] columnar wscg

zhouyuan:
@xuechendi let's use this issue id for the columnar wscg refactor

zhouyuan:
https://github.com/Intel-bigdata/OAP/pull/2061

PR #2081, opened 2020-12-21
[OAP-1856][oap-native-sql]Remove ColumnarBHJ logic from AQE to ColumnarPlugin 

xuechendi:
In this PR, we moved out most modifications inside AQE, and put them to ColumnarPlugin.scala
Passed TPCDS tests with AQE enabled + V2
Passed TPCDS tests with DPP enabled + V1
Also tested with TPCH Q2 and Q16 with AQE is off.

put remaining modification diff as a patch in tools folder, so we can keep a track.

Fixed: https://github.com/Intel-bigdata/OAP/issues/1856

### DIFF
--- a/oap-native-sql/core/src/main/scala/com/intel/oap/ColumnarGuardRule.scala
+++ b/oap-native-sql/core/src/main/scala/com/intel/oap/ColumnarGuardRule.scala
@@ -91,6 +91,8 @@ case class ColumnarGuardRule(conf: SparkConf) extends Rule[SparkPlan] {
             plan.condition,
             plan.left,
             plan.right)
+        case plan: BroadcastExchangeExec =>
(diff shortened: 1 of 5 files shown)


xuechendi:
@rui-mo , this PR is to remove our modification to AQE, I have tested with TPCDS and passed UT here, can you also test all UT in your env , thanks

xuechendi:
@zhouyuan , checked with latest spark AQE codes, remaining modification is aligned with NVIDIA's change, so when it update to 3.0.1, we won't need to change AQE file

rui-mo:
@xuechendi all unit tests succeed in my env.

xuechendi:
> @xuechendi all unit tests succeed in my env.

Thanks!

PR #2091, opened 2021-02-02
[OAP-2068][POAE7-768][OAP-CACHE]PlasmaTimeOutWrapper

offthewall123:

Wrapper origin plasmaClient methods  with timeout mechanism.

## How was this patch tested?

E2E


offthewall123:
@jikunshang @winningsix @yeyuqiang @xieqi Please take a review, thanks.

oap-cache/oap/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -1052,30 +1072,72 @@ class ExternalCache(fiberType: FiberType) extends OapCache with Logging {
...
-    else false
+    val plasmaClient = plasmaClientPool(clientRoundRobin.getAndAdd(1) % clientPoolSize)
+    val paramValue = new Array[Object](1)
+    paramValue(0) = objectId
+    var res: Boolean = false
+    try {
+      res = PlasmaTimeOutWrapper.run("contains", plasmaClient, executorService, paramValue, 3)
+              .asInstanceOf[Boolean]
+    } catch {
+      case ie: InterruptedException =>
```
jikunshang, line 1101: can we refactor like this?  
` case e: InterruptedException | ExecutionException  | TimeoutException =>     `
`logWarning("contains exception, type: " + e.getClass.getName + " message " + e.getMessage) `

oap-cache/oap/src/main/java/org/apache/PlasmaTimeOutWrapper.java
```diff
@@ -0,0 +1,65 @@
...
+    this.paramValues = paramValues;
+    this.timeOut = timeOut;
+    FutureTask<Object> futureTask = (FutureTask<Object>) executorService.submit(this);
+    executorService.execute(futureTask);
+    return futureTask.get(timeOut, TimeUnit.SECONDS);
+  }
+
+  @Override
+  public Object call() throws Exception {
+    Method[] methods = plasmaClient.getClass().getDeclaredMethods();
```
jikunshang, line 54: hmmm, do we have some better way? or make this a global variable or map for better performance?

jikunshang (changes requested):

oap-cache/oap/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
```diff
@@ -994,17 +995,34 @@ class ExternalCache(fiberType: FiberType) extends OapCache with Logging {
...
   override def getEmptyFiber(fiberLength: Long, fiberId: FiberId = null): FiberCache = {
+    val length: Int = fiberLength.toInt
     val objectId = hash(fiberId.toString)
     val plasmaClient = plasmaClientPool(clientRoundRobin.getAndAdd(1) % clientPoolSize)
+    val paramValue = new Array[Object](2)
+    paramValue(0) = objectId
+    paramValue(1) = length.asInstanceOf[Object]
     try {
-      val buf: ByteBuffer = plasmaClient.create(objectId, fiberLength.toInt)
+      val buf: ByteBuffer = PlasmaTimeOutWrapper.run("create", plasmaClient, executorService, paramValue, 3)
```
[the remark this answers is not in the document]
offthewall123: Added, original thougths is to avoid additional conf.

oap-cache/oap/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
offthewall123: Seems can't case match in this way would cause "object java.util.concurrent.ExecutionException is not a value"

oap-cache/oap/src/main/java/org/apache/PlasmaTimeOutWrapper.java
offthewall123: Yes,can define a method map here, but i think it's strange to define it.We should get methods through reflect.

jikunshang: My question here is whether this `call` method will be called multi times. if so,  `getDeclaredMethods` may influence performance I guess.

offthewall123: Yes, this method will be called many times.

oap-cache/oap/src/main/scala/org/apache/spark/sql/execution/datasources/oap/filecache/OapCache.scala
jikunshang: try this:
`case e @ (_: InterruptedException  | _ : ExecutionException | _:TimeoutException ) => `

oap-cache/oap/src/main/java/org/apache/PlasmaTimeOutWrapper.java
jikunshang: Could we pass some lambda function(wrap plasma methods)?

oap-cache/oap/src/main/java/org/apache/spark/sql/execution/datasources/oap/filecache/PlasmaLambdaWrapper.java
```diff
@@ -0,0 +1,51 @@
+package org.apache.spark.sql.execution.datasources.oap.filecache;
```
jikunshang, line 1: add license

oap-cache/oap/src/main/java/org/apache/spark/sql/execution/datasources/oap/filecache/PlasmaLambdaWrapper.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.apache.arrow.plasma.PlasmaClient;
+
+interface PlasmaLambdaWrapper {
```
jikunshang, line 5: rename to `PlasmaMethodWrapper` ?

oap-cache/oap/src/main/java/org/apache/spark/sql/execution/datasources/oap/filecache/PlasmaLambdaWrapper.java
```diff
@@ -0,0 +1,51 @@
...
+    Object execute(PlasmaClient client, Object... param);
+}
+
+class PlasmaCreate implements PlasmaLambdaWrapper {
+
+  @Override
+  public Object execute(PlasmaClient client, Object... param) {
```
jikunshang, line 12: can we define an extra class to replace `object[] param ` here?

oap-cache/oap/src/main/java/org/apache/spark/sql/execution/datasources/oap/filecache/PlasmaLambdaWrapper.java
jikunshang, line 5: Do we need to wrap connect method?

offthewall123: `for ( i <- 0 until clientPoolSize) {
    try {
      plasmaClientPool(i) = new plasma.PlasmaClient(externalStoreCacheSocket, "", 0)
    } catch {
      case e: Exception =>
        logError(s"Error occurred when connecting to plasma server" + e.getMessage)
    }
  }`

Connect is just new PlasmaClient().

oap-cache/oap/src/main/java/org/apache/spark/sql/execution/datasources/oap/filecache/PlasmaLambdaWrapper.java
offthewall123: Yes, we can, but i think it's too over-design.
1: Here param length is variable

2: At most 3 params
if define a class
plasmaParam {
  objectid
  timeout
  length
  isMetadata
}

will exist some empty value in the class.

jikunshang: some member could be null or default value if it's not used. it's over-design but more readable. For someone  don't have context, param here will be confusing.

jikunshang:
LGTM.

